Issue #2, opened 2020-04-14
[iOS] Enter Safe Address Screen UI


DmitryBespalov:
- Use Address Input
- Open Safari screen
- Loading state
- Showing of errors (to be clarified)

[Zeplin link](https://zpl.io/VxZmWqw)

PR #9, opened 2020-04-15
 Added elements to the enter address screen

DmitryBespalov:
wip #2 

- Added Enter Safe Address screen
- Added the beginning of the Address Input
- Connected screens and buttons to open each other


Multisig/Presentation/Add Safe/SafariViewController.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+//  Created by Dmitry Bespalov on 15.04.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+import SafariServices
+import SwiftUI
+
+struct SafariViewController: UIViewControllerRepresentable {
```
MouazAlzahabi, line 13: Have you tried to use UIViewRepresentable instead?

DmitryBespalov: No, I thought because this is a view controller, so I used the controller representable ...

PR #13, opened 2020-04-21
add AppSettings model; add createdAt field to models;

sche:
Please pay attention to AppSettings.getOrCreate() method. Here I have some open questions:

1) Should we use model entity extension to provide methods that create/return NSManagedObject (like AppSettings in our case) at all?

2) Or would you prefer to have these methods in a separate class like SafesController? 

If yes, should we save context immediately, or should it be done by the caller at some point?
I was implementing the "save_and_select_safe" method but later figured out that it might be confusing to select not persisted yet safe. 
So I think that it makes sense to give to a caller (logic layer) responsibility to create/persist/link/revert concrete entities. (option 3)

3) Or would it be ok not to have these methods at all and let logic layer to create/link/save/revert entities (NSManagedObject) with a granular control?

### DIFF
--- a/Multisig.xcodeproj/project.pbxproj
+++ b/Multisig.xcodeproj/project.pbxproj
@@ -34,6 +34,10 @@
 		0AF64FA524475BC30018D070 /* SafariViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AF64FA424475BC30018D070 /* SafariViewController.swift */; };
 		0AF6500A24485E450018D070 /* Identicon.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AF6500924485E450018D070 /* Identicon.swift */; };
 		0AF6501B24488D280018D070 /* FullSize.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AF6501A24488D280018D070 /* FullSize.swift */; };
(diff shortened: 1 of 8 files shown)


DmitryBespalov:
> Please pay attention to AppSettings.getOrCreate() method. Here I have some open questions:
> 
>     1. Should we use model entity extension to provide methods that create/return NSManagedObject (like AppSettings in our case) at all?

This was our final decision, as far as I remember.

> 
>     2. Or would you prefer to have these methods in a separate class like SafesController?
> 
> 
> If yes, should we save context immediately, or should it be done by the caller at some point?
> I was implementing the "save_and_select_safe" method but later figured out that it might be confusing to select not persisted yet safe.

This was my initial proposal, as this puts the responsibility of manipulating safes and implementing the validation or additional logic to one class.

> So I think that it makes sense to give to a caller (logic layer) responsibility to create/persist/link/revert concrete entities. (option 3)
> 
>     1. Or would it be ok not to have these methods at all and let logic layer to create/link/save/revert entities (NSManagedObject) with a granular control?

The logic layer is responsible to when to save or reset, yes, and to change the model property (selected safe). To make such change, it needs to get the entity. This code of accessing the CoreData should be somewhere, we decided to put it to the extensions on the data objects.

PR #37, opened 2020-04-29
Gh 14/safe switcher styling

### LINKED ISSUE #14: [iOS] Safe Switcher
- List of safes with switching the safe selection

- There is a list of Safes (Can be empty)

### LINKED ISSUE #30: [iOS] Create database relationship to get selected safe from the settings

sche:
- Changing relationship: closes #30 
- Changing styles: closes #14


Multisig.xcodeproj/project.pbxproj
```diff
@@ -928,7 +928,7 @@
 					"@executable_path/Frameworks",
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = io.gnosis.multisig.prerelease;
-				PRODUCT_NAME = "Safe Multisig";
+				PRODUCT_NAME = Multisig;
```
DmitryBespalov, line 931: @sche what's that? why not Safe Multisig ? 😄

Multisig/UI/UI Library/SafeCell.swift
```diff
@@ -22,6 +22,7 @@ struct SafeCell: View {
             VStack (alignment: .leading){
                 BodyText(label: safe?.name ?? "")
                 AddressText(safe?.address ?? "", style: .short)
+                    .padding(EdgeInsets(top: 4, leading: 0, bottom: 0, trailing: 0))
```
DmitryBespalov, line 25: there's more succinct way:

```
.padding(.top, 4)
```

Multisig.xcodeproj/project.pbxproj
sche: then in test target, I need to import Safe_Multisig. So I returned it to how it was (as I introduced Safe Multisig in 0.1.0 release) so we could decide later together on the name.

Multisig/UI/UI Library/SafeCell.swift
sche: nice

PR #108, opened 2020-05-13
gh-79  changed bg color of the lists

### LINKED ISSUE #79: Safe Settings: background color of the screen
Background color behind the list should be like section header color

DmitryBespalov:
fixes #79


Multisig/UI/Safe Management/Switch Safe/SwitchSafeView.swift
```diff
@@ -28,11 +28,11 @@ struct SwitchSafeView: View {
                     }
                 }
             }
-            .navigationBarTitle("", displayMode: .inline)
-            .navigationBarItems(leading: closeButton)
             .onAppear {
-                UITableView.appearance().separatorInset = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 0)
+                UITableView.appearance().backgroundColor = nil
```
MouazAlzahabi, line 32: I would prefer to move all UITableView.appearance() to UIApplicationDelegate, it's not necessary to be called each time the view appear, also creating something like ThemeManager to be responsable of all general styles like this one.

DmitryBespalov: The problem with the appearnace() is that it applies to all instances of the tables. In this specific case of SwitchSafeView we don't want to have that background color, so I make it nil. 

Good idea about ThemeManager, I'll refactor the current code.

PR #112, opened 2020-05-13
Gh 64 enter ens name UI

### LINKED ISSUE #64: ENS name entry visual differences with the UI design
- [x] The ENS name is auto-capitalized, this should not be the case. Also, the grammar correction should be disabled.
- [x] Add "." to the keyboard by default (I think there should be a separate keyboard for that)
- [x] The text field border and icon are not what the UI specifies
- [x] The check mark icon is in the wrong place (should be near the "Address found" text)
- [x] The address text is not bold as it should be in the design.

DmitryBespalov:
closes #64 

Extracted system images to a one central location (Icons.swift)


Multisig/UI/UI Library/Styleguide/Icons.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+//  Created by Dmitry Bespalov on 13.05.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import SwiftUI
+
+extension Image {
+
+    // using static let ... = ... is crashing the app when the views are
```
sche, line 13: "Fun fact" series

PR #119, opened 2020-05-14
Gh 88/balances

sche:
WIP
Added:
- Networking
- Basic layout
- Kingsfisher

TODO:
- Formatting numbers
- Final design


Multisig/Cross-layer/App.swift
```diff
@@ -23,11 +23,12 @@ class App {
...
+        url: URL(string: "https://safe-relay.rinkeby.gnosis.io")!, logger: LogService.shared)
     
-    var safeTransactionService = SafeTransactionService(url: URL(string: "https://safe-transaction.rinkeby.gnosis.io")!,
-    logger: LogService.shared)
+    var safeTransactionService = SafeTransactionService(
+        url: URL(string: "https://safe-transaction.rinkeby.gnosis.io")!, logger: LogService.shared)
 
-    var nodeService = EthereumNodeService(url: URL(string: "https://rinkeby.infura.io/v3/438e11915f8b4834a05e7810b88db4b3")!)
+    var nodeService = EthereumNodeService(
+        url: URL(string: "https://rinkeby.infura.io/v3/438e11915f8b4834a05e7810b88db4b3")!)
```
DmitryBespalov, line 33: you know what, I've added this when adding ENS, and now our API key is published here :(

Multisig/Data/Services/Networking Utils/HTTPClient.swift
```diff
@@ -81,6 +81,8 @@ public class HTTPClient {
     }
 
     private func send(_ request: URLRequest) -> URLDataTaskResult {
```
DmitryBespalov, line 83: this should break the `SafeRelayServiceIntegrationTests` - please update from master and run locally

Multisig/Cross-layer/App.swift
sche: damn, we need to invalidate it

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -20,6 +20,10 @@ class SafeTransactionService {
     func safeInfo(at address: String) throws -> SafeStatusRequest.Response {
         return try httpClient.execute(request: SafeStatusRequest(address: address))
     }
+
+    func safeBalances(at address: String) throws -> [SafeBalancesRequest.Response] {
```
DmitryBespalov, line 24: please add an integration test for that endpoint

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -45,3 +49,29 @@ struct SafeStatusRequest: JSONRequest {
         let version: String
     }
 }
+
+struct SafeBalancesRequest: JSONRequest {
+    let address: String
```
DmitryBespalov, line 54: I would keep this as `var`, because the whole struct can be made immutable with `let`

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -45,3 +49,29 @@ struct SafeStatusRequest: JSONRequest {
...
+
+    init(address: String) {
+        self.address = address
+    }
+
+    struct Response: Decodable {
+        let tokenAddress: String? // nil == Ether
+        let token: Token? // nil == Ether
+        let balance: String
+        let balanceUsd: String
```
DmitryBespalov, line 69: these too, I would keep them as `var`s (same for Token)

Multisig/UI/App/ContentView.swift
```diff
@@ -21,33 +23,43 @@ struct ContentView: View {
...
+                        }
                     }
-                }
-                .tag(2)
+                    .tag(2)
+            }
+        }
+    }
+
+    var assetsOrAddSafeIntroView: some View {
```
DmitryBespalov, line 57: please extract it into a separate `AssetsContainerView`

DmitryBespalov: together with the fetch request

Multisig/Data/Services/SafeTransactionService.swift
sche: What is an additional value in it? If you are not going to change smth, why not make it 'let'. If the case like this appears, I would rather make it 'var' later.

Multisig/UI/App/ContentView.swift
DmitryBespalov: it is good to encapsulate this and separate from the content view.

Multisig/UI/Assets/AssetsView.swift
```diff
@@ -0,0 +1,57 @@
...
+
+import SwiftUI
+
+struct AssetsView: View {
+    @ObservedObject
+    var model = AssetsViewModel()
+
+    var body: some View {
+        ZStack(alignment: .center) {
+            if model.isLoading {
```
DmitryBespalov, line 17: please create a new issue with a `technical debt` to refactor this piece into a separate more generic view that has loading, error, and content.

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+//  Created by Andrey Scherbovich on 13.05.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+struct TokenBalance: Identifiable {
+    var id: String {
+        return address
```
DmitryBespalov, line 14: `return` can and should be removed to use implicit returns - more straightforward.

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    var imageURL: URL {
+        return URL(string: "https://gnosis-safe-token-logos.s3.amazonaws.com/\(address).png")!
```
DmitryBespalov, line 17: here too, `return` can be removed

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    let address: String
+    let balance: String
+    let balanceUsd: String
```
DmitryBespalov, line 21: again, why we need to constrain structs with `let`?

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+
+class AssetsViewModel: ObservableObject {
+    @Published var balances = [TokenBalance]()
+    @Published var isLoading: Bool = true
+    @Published var errorMessage: String? = nil
+
+    private var subscribers = Set<AnyCancellable>()
+
+    init() {
+        let request = Safe.fetchRequest().selected()
```
DmitryBespalov, line 32: the selected safe should be passed as an argument to the initializer from the view itself, because we already fetch the selected safe in the `ContentView`,  I think this would work out nicely here.

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+        let selectedSafe = try! CoreDataStack.shared.viewContext.fetch(request).first!
+        isLoading = true
+        Just(selectedSafe.address!)
```
DmitryBespalov, line 35: this is not good, I'd rather `Just(address).compactMap { $0 }` - we don't have a requirement that the address is going to be known at all times, and this model will be more useful.

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+            .setFailureType(to: Error.self)
```
DmitryBespalov, line 36: good find! Can you find other places in code where I used `tryMap` or something - to also use this failure type instead?

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+            .flatMap { address in
+                Future<[TokenBalance], Error> { promise in
+                    DispatchQueue.global().async {
+                        do {
+                            let balancesResponse = try App.shared.safeTransactionService.safeBalances(at: address)
+                            let tokenBalances = balancesResponse.map {
+                                TokenBalance(address: $0.tokenAddress ?? "0x0",
+                                             balance: $0.balance,
+                                             balanceUsd: $0.balanceUsd)
```
DmitryBespalov, line 45: What do you think about extracting these two methods into a business-logic method on the Safe, because this will be more complicated in the nearest future?

Podfile
```diff
@@ -20,6 +20,8 @@ target 'Multisig' do
   pod 'idn2Swift', :git => 'https://github.com/gnosis/pod-idn2.git', :branch => 'master', :testspecs => ['Tests']
   pod 'BigInt', '4.0.0'
   
+  pod 'Kingfisher/SwiftUI'
```
DmitryBespalov, line 23: please add a specific version requirement to this line, as an example - see line 21 above

Multisig/UI/UI Library/TokenImage.swift
```diff
@@ -0,0 +1,27 @@
...
+import struct Kingfisher.KFImage
+
+struct TokenImage: View {
+    var imageURL: URL
+    var size: CGFloat = 28
+
+    var body: some View {
+        KFImage(imageURL)
+            .placeholder {
+                Image(systemName: "arrow.2.circlepath.circle")
```
DmitryBespalov, line 19: It doesn't look pretty 😄 I think it is better to use the icon from the previous project.

Multisig/UI/Assets/AssetsViewModel.swift
DmitryBespalov: the icon for ETH is not working.

Multisig/UI/Assets/AssetsView.swift
MouazAlzahabi: As we still have time, why to make technical debt from now

Multisig/Data/Services/SafeTransactionService.swift
DmitryBespalov: the meaning of the `let` is that it declares a constant value that is immutable. In Swift, structs themselves are immutable. 

When I read `let` inside of the struct, I read it like "here is the constant address". And I wonder why the address cannot change? 

The `var` is the default choice, I would keep it as it is, having the `let` here constrains this struct from other usage and disallows me to change just one value of the struct - I need to re-initialize this struct all together with all the values the same except that one.

I think this does not bring much value, and it confuses me when I read it.

DmitryBespalov: surely I should be able to construct a request with one address, and if I want, change that address property to another value. But I won't be able to do that, because the address is a constant. 

And then I become frustrated.

DmitryBespalov: I don't want to come back to this code later, this makes this code not reusable, as it would be reusable by default

DmitryBespalov: in general, one of the design objective is to anticipate the future changes

Multisig/UI/Assets/AssetsView.swift
DmitryBespalov: I think this should be done in another PR, to reduce growth of this PR

Multisig/Data/Services/Networking Utils/HTTPClient.swift
sche: fixed integration tests

Multisig/Data/Services/SafeTransactionService.swift
sche: done

Multisig/UI/App/ContentView.swift
sche: done

Multisig/UI/Assets/AssetsView.swift
sche: done: https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/128

Multisig/UI/Assets/AssetsViewModel.swift
sche: done

Multisig/UI/Assets/AssetsViewModel.swift
sche: this will be replaced soon, just a tmp solution

Multisig/Data/Services/SafeTransactionService.swift
sche: discussed. Decision: not to change in this PR.

Multisig/UI/UI Library/TokenImage.swift
sche: The final design will come with the next pr. This is a placeholder.

Podfile
sche: done

Multisig/UI/Assets/AssetsViewModel.swift
sche: done

Multisig/UI/Assets/AssetsViewModel.swift
sche: As the safe is passed in the initializer, the address will always be there.

Multisig/UI/Assets/AssetsViewModel.swift
sche: I created a debt issue: https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/132
Sorry, now I want to merge this PR asap.

Multisig/UI/Assets/AssetsViewModel.swift
sche: I will make final adjustments with the final PR. Maybe yes, maybe no, let's see.

PR #120, opened 2020-05-14
Gh 46 top tab view

DmitryBespalov:
adds the top tab view.

Right now every time the tab is switched, it is refreshed (request is made). I think this is not what we want. I'll fix that in another PR.

Also, SwiftUI generics is fun 😄


Multisig/Assets.xcassets/ico-app-settings.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
```
MouazAlzahabi, line 2: Why not use single scale image?

Multisig/Assets.xcassets/ico-app-settings.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
...
+    {
```
MouazAlzahabi, line 3: why not using single scale image?

Multisig/Assets.xcassets/ico-app-settings.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
...
+      "filename" : "ico-app-settings.pdf",
+      "idiom" : "universal",
+      "scale" : "1x"
```
MouazAlzahabi, line 6: same here

Multisig/Assets.xcassets/ico-safe-settings.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
```
MouazAlzahabi, line 2: same here

Multisig/UI/Settings/SettingsTopTabView.swift
```diff
@@ -10,33 +10,28 @@ import SwiftUI
 
 struct SettingsTopTabView: View {
```
MouazAlzahabi, line 11: I would remove the word "Top" wherever it occurs

Multisig/Assets.xcassets/ico-app-settings.imageset/Contents.json
DmitryBespalov: will do, forgot to add it :)

Multisig/UI/Settings/SettingsTopTabView.swift
DmitryBespalov: I will rename it to `SettingsContainerView`

Multisig/UI/UI Library/TopTabView/BottomBorder.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 14.05.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import SwiftUI
+
+// Draws a configurable-height bottom border
+struct BottomBorder: Shape {
```
MouazAlzahabi, line 12: BottomBorder is not clear here, usually they call it "selector" or so

Multisig/UI/UI Library/TopTabView/TopTabChildView.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Created by Dmitry Bespalov on 14.05.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import SwiftUI
+
+// Wraps together label and content view to be used in the TopTabView
+struct TopTabChildView<Label: View, Content: View>: View {
+
+    var label: Label
```
MouazAlzahabi, line 14: Do we need to have label?

Multisig/UI/UI Library/TopTabView/BottomBorder.swift
DmitryBespalov: `Underline`?

Multisig/UI/UI Library/TopTabView/TopTabChildView.swift
DmitryBespalov: this is a content of the tab bar (icon + text)

Multisig/UI/Settings/SettingsTopTabView.swift
MouazAlzahabi: Good one, also please remove "Top" from the other areas

Multisig/UI/UI Library/TopTabView/BottomBorder.swift
MouazAlzahabi: I still prefer selector, as it's indicate the selection state

DmitryBespalov: but it is a shape, I would name it selector if it would be a view, then it would have a clear purpose, but the shape is very generic, it can be selector in one place, and a decoration in another place (let's say underline of a text field)

DmitryBespalov: and bottom border is saying what the shape is, and it is clear what will be drawn

Multisig/UI/UI Library/TopTabView/TopTabChildView.swift
DmitryBespalov: I can rename it to the `tabItemContent`

DmitryBespalov: or `tabButtonLabel`

DmitryBespalov: I took the name from the SwiftUI's tabItem:

```
    public func tabItem<V>(@ViewBuilder _ label: () -> V) -> some View where V : View
```

it declares the content as the `label`, so that's why I used `label`. It seems to be the preferred name for such things.

MouazAlzahabi: keep it label please

Multisig/UI/UI Library/TopTabView/BottomBorder.swift
MouazAlzahabi: Both ways are fine I think

Issue #70, opened 2020-05-11
Advanced Safe Settings content


DmitryBespalov:
- [x] We need not to show the master copy address, because we already show it in the main screen together with the contract version
- [x] The fallback handler should have a human-readable text. If the handler is our default address (right now it is 0xd5D82B6aDDc9027B22dCA772Aa68D5d74cdBdF44), then we display "DefaultFallbackHandler". If it is a zero address, we should display text "Not set" and do not display the address. In other cases, we should display "Unknown".
- [x] The screen should show the addresses truncated as in the designs

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/e8015487-7644-4fef-ac48-be3ef66cfe55" width=600px/>

sche:
I would add here the "share icon" should be not bold.

DmitryBespalov:
> I would add here the "share icon" should be not bold.

This is mentioned in a different ticket, I think

PR #133, opened 2020-05-19
gh-70 apply changes

MouazAlzahabi:


Multisig/Cross-layer/App.swift
```diff
@@ -33,6 +33,8 @@ class App {
     let licensesURL = URL(string:"https://gnosis-safe.io/licenses/")!
     let appVersion = "\(Bundle.main.infoDictionary?["CFBundleShortVersionString"] as! String) (\(Bundle.main.infoDictionary?["CFBundleVersion"] as! String))"
 
+    let defaultFallbackHandler = "0xd5D82B6aDDc9027B22dCA772Aa68D5d74cdBdF44"
```
DmitryBespalov, line 36: please make this of an Address type, so that it won't be lowercase/uppercase dependent.

Multisig/UI/UI Library/Safe Settings/AdvancedSafeSettingsView.swift
```diff
@@ -42,12 +38,14 @@ struct AdvancedSafeSettingsView: View {
     }
     
     var fallbackHandlerView : some View {
-        Group {
-            if (safe.fallbackHandler ?? "").isEmpty  {
-                 EmptyView()
+        let fallbackHandler = safe.fallbackHandler ?? ""
```
DmitryBespalov, line 41: I would add the `isDefaultFallbackHandler` to the Safe model, and then just use that property in the condition below.

```
let title = safe.isDefaultFallbackHandler ? "..." : "Unknown"
```

Multisig/UI/UI Library/Address/AddressCell.swift
```diff
@@ -29,6 +42,16 @@ struct AddressCell: View {
...
         .padding(EdgeInsets(top: 2, leading: 0, bottom: 6, trailing: 0))
     }
+
+    var addressText: some View {
+        var addressStyle: AddressText.Style = .long
+        if style == .shortAddress {
+            addressStyle = .short
+        }
+
+        return AddressText(address, style: addressStyle)
```
DmitryBespalov, line 52: can't we just use the ternary operator here?

```
style: style == .shortAddress ? .short : .long
```

Multisig/UI/UI Library/Address/AddressCell.swift
```diff
@@ -9,8 +9,16 @@
 import SwiftUI
 
 struct AddressCell: View {
+
+    enum Style {
```
DmitryBespalov, line 13: instead of this, we already have the `AddressStyle` enum, maybe let's use that one?

Multisig/UI/UI Library/Safe Settings/AdvancedSafeSettingsView.swift
MouazAlzahabi: I don't think so, this feature is related to address not safe, maybe we need to check if given address is default fallback handler or not

DmitryBespalov: per discussion: this is related to the Safe contract (default values) and similar to the decision about the supported master copies. So, creating a separate SafeContract class would be the way to go.

Multisig/Logic/Models/Safe.swift
```diff
@@ -38,6 +38,10 @@ extension Safe: Identifiable {
         URL(string: "https://etherscan.io/address/\(address)")!
     }
 
+    func isDefaultFallbackHandler() -> Bool {
+        return fallbackHandler == App.shared.defaultFallbackHandler?.hex(eip55: true)
```
DmitryBespalov, line 42: ok, I see this is done here, good, but now we are again dependent on the server-returned formatting. It would be better to compare addresses together, not strings:

```
Address(fallbackHandler) == App.shared.defaultFallbackHandler
```

PR #135, opened 2020-05-19
Gh 88/design

### LINKED ISSUE #88: [iOS] View ETH and ERC20 token balance
- User can see their ETH balance, ETH is always on top
- User can see the balance of all ERC20 tokens with balance > 0, tokens are ordered by their names, ascending
- User sees fiat value in USD per asset
- On appear or coming to foreground, the screen should refresh if it is shown to the user (whenever the screen is destroyed by the OS).
- Pull to refresh will reload the balances manually

### API Services

- To get the balances, use the transaction service:
  - https://safe-transaction.gnosis.io/
  - GET /safes/{address}/balances/usd/
  - [x] The balances are not paginated
  - Returns both the token balance and the balance in USD
  - the "null" token is "ETH" token.
  - The endpoint returns the token information (name, decimals). 
  - The logo is to be added, but for now you can take it by using the URL as a template from the relay service.
    - https://safe-relay.gnosis.io/
    - endpoint GET /v1/tokens/

## Depends on 
https://github.com/gnosis/safe-transaction-service/issues/91
https://github.com/gnosis/safe-transaction-service/issues/86

Screenshot 2020-05-11 at 19 09 31

sche:
closes #88 

Screenshot 2020-05-19 at 14 12 05


Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -8,17 +8,37 @@
...
+        self.symbol = response.token?.symbol ?? "ETH"
+
+        let tokenFormatter = TokenFormatter()
+        let decimals =  response.token?.decimals ?? 18
+        self.balance = tokenFormatter.string(from: BigDecimal(BigInt(response.balance) ?? 0, decimals))
+
+        let currencyFormatter = NumberFormatter()
+        let number = currencyFormatter.number(from: response.balanceUsd) ?? 0
+        currencyFormatter.numberStyle = .currency
+        currencyFormatter.currencyCode = "USD"
```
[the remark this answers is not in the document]
DmitryBespalov: discussion result:  to clarify with product this system behavior.

Multisig/Assets.xcassets/ico-ether.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
```
MouazAlzahabi, line 2: Just use single scale instead of individual

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -8,17 +8,37 @@
...
-        URL(string: "https://gnosis-safe-token-logos.s3.amazonaws.com/\(address).png")!
+    var imageURL: URL? {
+        // will be replaced when https://github.com/gnosis/safe-transaction-service/issues/86 is ready
+        guard let address = address else { return nil }
+        return URL(string: "https://gnosis-safe-token-logos.s3.amazonaws.com/\(String(describing: address)).png")!
     }
-    let address: String
+    let address: String?
+    let symbol: String
     let balance: String
```
MouazAlzahabi, line 25: As these values are usually parsed from backend and in term of defensive programing, I don't expect that server return values, so I used to make them all optional to avoid cashing on parsing.

Multisig/UI/Assets/AssetsViewModel.swift
```diff
@@ -8,17 +8,37 @@
...
     let balanceUsd: String
+
+    init(_ response: SafeBalancesRequest.Response) {
+        self.address = response.tokenAddress
+        self.symbol = response.token?.symbol ?? "ETH"
+
+        let tokenFormatter = TokenFormatter()
```
MouazAlzahabi, line 32: I think both tokenFormatter and currencyFormatter can be defined as shared constants on app level, no need to have an instance foreach token

sche: In different parts of the app we might need to use different formatting options. I think it is ok to have it like this.

Multisig/UI/Assets/AssetsViewModel.swift
sche: All these fields are required by the specification. Optional means that the server may return nil value for this field.
I would not use optional everywhere.

Multisig/Assets.xcassets/ico-ether.imageset/Contents.json
sche: ether icon images were provided as 3 images

Multisig/UI/Assets/AssetsViewModel.swift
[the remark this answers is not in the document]
sche: decision: display depending on the user locale

PR #142, opened 2020-05-20
Gh 77 error handling safe address

### LINKED ISSUE #77: Load Safe: bad error message when not a safe address provided
Bad error message when entering not a safe address.

Screenshot 2020-05-11 at 17 38 21

sche:
closes #77


Multisig/Data/Services/Networking Utils/HTTPClient.swift
```diff
@@ -104,11 +119,16 @@ public class HTTPClient {
...
-                throw Error.networkRequestFailed(request, result.response, result.data)
+        guard let httpResponse = result.response as? HTTPURLResponse else {
+            throw Error.networkRequestFailed(request, result.response, result.data)
         }
-        return data
+        if (200...299).contains(httpResponse.statusCode) {
+            return result.data ?? Data()
+        } else if httpResponse.statusCode == 404 {
+            throw Error.entityNotFound(request, result.response, result.data)
+        } else if httpResponse.statusCode == 422 {
```
DmitryBespalov, line 129: great! can we have integration tests for each of these cases?

Multisig/UI/Safe Management/Load Safe/EnterSafeAddress/EnterSafeAddressViewModel.swift
```diff
@@ -80,7 +80,11 @@ class EnterSafeAddressViewModel: ObservableObject {
             }
             .sink(receiveCompletion: { completion in
                 if case .failure(let error) = completion {
-                    self.setError(error.localizedDescription)
+                    if case HTTPClient.Error.entityNotFound(_, _, _) = error {
+                        self.setError("We could not find a safe with this address.")
+                    } else {
```
DmitryBespalov, line 85: As far as I tried with the [invalid safe address "123"](https://safe-transaction.staging.gnosisdev.com/api/v1/safes/123/), the 422 would mean here that the user address is not valid, so I would add it here with a human-readable message instead/or in addition to the "Unprocessable entity"

sche: I think we have client validation of valid addresses; can you tell me how you managed to trigger 422 error?

Multisig/Data/Services/Networking Utils/HTTPClient.swift
sche: added integration tests

MultisigIntegrationTests/SafeTransactionServiceIntegrationTests.swift
```diff
@@ -22,6 +22,34 @@ class SafeTransactionServiceIntegrationTests: XCTestCase {
         semaphore.wait()
     }
 
+    func test_safeInfo_whenSendingNotASafe_returns404Error() {
+        let semaphore = DispatchSemaphore(value: 0)
+        DispatchQueue.global().async {
+            try? XCTAssertThrowsError(try self.service.safeInfo(at: "0xc778417E063141139Fce010982780140Aa0cD5Ab")) {
```
sche, line 28: Fun fact (seems like Xcode bug):
It should work without 'try?', but the compiler produces an error "Invalid conversion from throwing function of type '() throws -> Void' to non-throwing function type '@convention(block) () -> Void'"
With try? compiler complains "No calls to throwing functions occur within 'try' expression"

Multisig/UI/Safe Management/Load Safe/EnterSafeAddress/EnterSafeAddressViewModel.swift
DmitryBespalov: yes, that's right. I triggered by following the URL in the comment above.

MultisigIntegrationTests/SafeTransactionServiceIntegrationTests.swift
DmitryBespalov: I think it is because the async {} method expects non-throwing closure. Might be that the XCTAssertThrows actually rethrows?

MultisigIntegrationTests/SafeTransactionServiceIntegrationTests.swift
```diff
@@ -22,6 +22,34 @@ class SafeTransactionServiceIntegrationTests: XCTestCase {
...
+                }
+            }
+            semaphore.signal()
+        }
+        semaphore.wait()
+    }
+
+    func test_safeInfo_whenSendingNotNormalizedAddress_returns422Error() {
+        let semaphore = DispatchSemaphore(value: 0)
+        DispatchQueue.global().async {
```
DmitryBespalov, line 41: I want to prototype re-implementing the networking using Combine, we might get rid of chunks of the code.

sche: sounds good

PR #168, opened 2020-05-22
Gh 155/qr code scanner does not support ethereum scheme urls

MouazAlzahabi:


Pods/Web3/Web3/Classes/Core/Transaction/EthereumAddress.swift
```diff
@@ -163,6 +166,19 @@ public struct EthereumAddress {
         return hex
     }
 
+    // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md
```
DmitryBespalov, line 169: this is a wrong place to modify the pod files. The pods will be re-generated by the next `pod install` or `pod update`, and these changes will be lost.

Also, adding this behavior to *all* decoding addresses from string, is plain wrong to me. Instead, it needs to be done in an extension to the Address in the Logic/Models/Address.swift

Pods/Web3/Web3/Classes/Core/Transaction/EthereumAddress.swift
```diff
@@ -163,6 +163,7 @@ public struct EthereumAddress {
         return hex
     }
 
+
```
DmitryBespalov, line 166: still one edit left, please revert 😄

Multisig/Logic/Models/Address.swift
```diff
@@ -43,3 +43,28 @@ extension Address {
         try? self.init(hex: value, eip55: false)
     }
 }
+
+extension Address {
+
+    init(_ value: String, _ isERC681: Bool) throws {
```
DmitryBespalov, line 49: please make the 2nd argument named, because it is not clear what "true" means without the isERC681 label

Multisig/Logic/Models/Address.swift
```diff
@@ -43,3 +43,28 @@ extension Address {
...
+        var addressString = value
+        if isERC681 {
```
DmitryBespalov, line 51: could you please add some unit tests for this added code?

PR #169, opened 2020-05-22
gh-100  added keyboard adaptive behavior to safe name screen

### LINKED ISSUE #100: Safe name input hidden on small devices
I'm using an iPhone SE with the app version 0.2.0.

- Add new Safe
- On the screen where I get prompted to enter a name for the Safe, the keyboard overlays the input field (see screenshot below)

Expected result: I'm able to see the input field somehow, perhaps scrolling could be enabled on this screen?

IMG_9131

DmitryBespalov:
fixes #100

### DIFF
--- a/Multisig/UI/Safe Management/Load Safe/EnterSafeName/EnterSafeNameView.swift
+++ b/Multisig/UI/Safe Management/Load Safe/EnterSafeName/EnterSafeNameView.swift
@@ -42,6 +42,7 @@ struct EnterSafeNameView: View {
         }
         .padding(.top, 24)
         .padding(.horizontal)
+        .keyboardAdaptive()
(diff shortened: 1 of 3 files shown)


Multisig/UI/Safe Management/Load Safe/EnterSafeName/EnterSafeNameView.swift
```diff
@@ -42,6 +42,7 @@ struct EnterSafeNameView: View {
         }
         .padding(.top, 24)
         .padding(.horizontal)
+        .keyboardAdaptive()
```
MouazAlzahabi, line 45: It's useful if we add the same for EditSafeName

PR #194, opened 2020-05-27
gh-189 Create TitleText, EmptyListPlaceholder, TransactionsView

MouazAlzahabi:


Multisig/UI/Transactions/TransactionsView.swift
```diff
@@ -0,0 +1,36 @@
...
+    var body: some View {
+        let safe = selectedSafe.first
+        return ZStack {
+            Rectangle()
+               .edgesIgnoringSafeArea(.all)
+               .foregroundColor(Color.gnoWhite)
+            if safe == nil {
+                AddSafeIntroView()
+            }
+            else {
```
DmitryBespalov, line 25: braces should be on one line

Multisig/UI/Transactions/TransactionsView.swift
```diff
@@ -0,0 +1,36 @@
...
+    @FetchRequest(fetchRequest: Safe.fetchRequest().selected())
+    var selectedSafe: FetchedResults<Safe>
+
+    var body: some View {
+        let safe = selectedSafe.first
+        return ZStack {
+            Rectangle()
+               .edgesIgnoringSafeArea(.all)
+               .foregroundColor(Color.gnoWhite)
+            if safe == nil {
```
DmitryBespalov, line 22: I'd just use `selectedSafe.first` instead of `safe` here and the whole thing will not need the explicit return

Multisig/UI/Transactions/TransactionsView.swift
```diff
@@ -0,0 +1,36 @@
...
+
+struct TransactionsView: View {
+
+    @FetchRequest(fetchRequest: Safe.fetchRequest().selected())
+    var selectedSafe: FetchedResults<Safe>
+
+    var body: some View {
+        let safe = selectedSafe.first
+        return ZStack {
+            Rectangle()
```
DmitryBespalov, line 19: I have a feeling that this zstack and background must be part of the `EmptyListPlaceholder,` not part of this view

Multisig/UI/Transactions/TransactionsView.swift
MouazAlzahabi: Then when we need to define the content view we pass selectedSafe.first again ?

DmitryBespalov: yes, or define it as a computed property on the `TransactionView` struct

Multisig/UI/Transactions/TransactionsView.swift
MouazAlzahabi: This way, when we define the content view we have to add the background again to it,
the background actually is for the list itself not for empty state only

DmitryBespalov: I see

Multisig/UI/Transactions/TransactionsView.swift
```diff
@@ -0,0 +1,33 @@
...
+    @FetchRequest(fetchRequest: Safe.fetchRequest().selected())
+    var selectedSafe: FetchedResults<Safe>
+
+    var body: some View {
+        ZStack {
+            Rectangle()
+               .edgesIgnoringSafeArea(.all)
+               .foregroundColor(Color.gnoWhite)
+            if selectedSafe.first == nil { AddSafeIntroView() }
+            else {
```
DmitryBespalov, line 22: Mouaz, it should be this:
```
if <condition> {
  something
} else {
  something else
}

PR #198, opened 2020-05-29
Gh 81 ci

### LINKED ISSUE #81: Set up the CI build for unit tests
Needs to be researched as there might be some interesting options available compared to our own machine.

- For now, only run unit tests for the "Multisig" target.
- The tests must update the code coverage and log it.
- Depending on the CI provider, we will need to use the `fastlane` to run the builds, but if there is a simple way to live without it, it would be preferable to do so.

DmitryBespalov:
# Overview
I have played and tried to configure the CI with fastlane, but it is slow (because of gems and many dependencies), and the Xcode's automatic signing becomes tricky to manage manually with 5 different bundle ids.

Instead, I tried to use the Xcode's command line tools, and they worked fine. Using them, we're able to run unit tests, and upload to TestFlight.

The only thing is code coverage reports - the newest Xcode's Test Plan format is not supported by any code coverage tool yet, but I included the overall code coverage results in the build log.

Current setup:
- Every branch started with "gh-" will be built and unit tested
- Every push to master will be archived and distributed to "Development Rinkeby" and "Development Mainnet" apps in TestFlight.

The build number reflects the build number for the current branch (we will need to bump it for the master branch, because I already submitted the build number 30 or so). 

closes #81


bin/test.sh
```diff
@@ -0,0 +1,29 @@
...
+| tee "$OUTPUT_DIR/xcodebuild-test.log" | xcpretty -c -r junit
+
+# print the total code  coverage
+xcrun xccov view --report --only-targets "$TEST_BUNDLE_PATH"
+
+# archive the test results
+tar -czf "$TEST_BUNDLE_PATH".tgz "$TEST_BUNDLE_PATH"
+
+# codecov does not support new Xcode coverage format (Test Plan) for now.
+# curl -s https://codecov.io/bash | bash -s -- -D "$TEST_BUNDLE_PATH" -t "${CODECOV_TOKEN}"
```
sche, line 29: So we will not display code coverage in our README, right?

DmitryBespalov: that's right

PR #199, opened 2020-06-03
Gh 96 terms animation ui

DmitryBespalov:
- refactored VStacks and padding to remove not needed empty rectangles used for layout
- Changed the animation of the terms view from the abrupt 'jumping' to the slide from the bottom
- Changed animation of switching to main view to be 'fade' like
- Removed 'fixedFrame' from texts as it does not have an effect (checked on all simulators).

### DIFF
--- a/Multisig/UI/UI Library/Overlays/BottomOverlayView.swift
+++ b/Multisig/UI/UI Library/Overlays/BottomOverlayView.swift
@@ -22,20 +22,18 @@ struct BottomOverlayView<Content>: View where Content: View {
 
     var body: some View {
         ZStack(alignment: .bottom) {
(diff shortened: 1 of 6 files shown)


Multisig/UI/UI Library/Overlays/BottomOverlayView.swift
```diff
@@ -13,6 +13,7 @@ struct BottomOverlayView<Content>: View where Content: View {
     private var content: Content
 
     private let cardBackgroundColor = Color.white
+    private let contentHeight: CGFloat = 400
```
sche, line 16: It seems like fixed value does not make this overlay view generic.

Issue #86, opened 2020-05-11
Create multiple configurations for the different versions of the app.


DmitryBespalov:
- The configurations:
 - ethereum chain ID
 - relay URL
 - transaction service URL
 - ethereum RPC URL
 - links 
   - to the Gnosis Safe web

The configuration should be in a separate file, and the file must be encrypted if it is ever to be stored in the repository.

PR #209, opened 2020-06-08
gh-86  using configuration from config files

DmitryBespalov:
- Using the configurations from the Info.plist and Config.xcconfig
- Added the "AppConfiguration" to model the configuration parameters


Multisig/UI/Onboarding/TermsView.swift
```diff
@@ -29,8 +29,8 @@ struct TermsView: View {
                 BulletText("We collect anonymized app usage data and crash reports to ensure the quality of our app.")
 
                 HStack {
-                    LinkButton("Privacy Policy", url: App.shared.privacyPolicyURL)
-                    LinkButton("Terms of Use", url: App.shared.termOfUseURL)
+                    LinkButton("Privacy Policy", url: App.configuration.legal.privacyURL)
```
MouazAlzahabi, line 32: I would create private let legal = App.configuration.legal as you did in BasicAppSettingsView.swift

Issue #177, opened 2020-05-26
[iOS] Transaction Details Type and Status


MouazAlzahabi:
### Type and Status UI Map

![TransactionTypeAndStatusUIMap](https://user-images.githubusercontent.com/284056/84517218-80081b00-acbe-11ea-8338-6b7b89642b46.png)

DmitryBespalov:
- [ ] Extract the "confirmations view" into a new ticket
- [ ] Extract the header views into a new ticket
- [ ] Extract all other fields into a new ticket
- [ ] Keep this ticket only for the whole transaction details screen

liliya-soroka:
Verified 0.5.0 (62)

PR #231, opened 2020-06-12
gh-177 Create transaction confirmation view

MouazAlzahabi:


Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfiramtionStatusView.swift
```diff
@@ -0,0 +1,101 @@
...
+            return "Executed"
+        case .confirmed:
+            return "Confirmed"
+        case .rejected:
+            return "Rejected"
+        case .canceled:
+            return "Canceled"
+        case .failed:
+            return "Failed"
+        case .waitingConfirmations(let remining):
```
DmitryBespalov, line 35: typo: "remining" -> "remaining"

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfiramtionStatusView.swift
```diff
@@ -0,0 +1,101 @@
...
+            return .gnoDarkBlue
+        case .failed:
+            return .gnoTomato
+        case .waitingConfirmations(_):
+            return .gnoMediumGrey
+        case .confirm:
+            return .gnoHold
+        case .execute:
+            return .gnoHold
+        }
```
DmitryBespalov, line 87: instead of 3 method with 3 switches, I'd have a lookup table

```
let viewData = [String: (title: Stiring, imageName: String, color: Color)]
// ...
color = viewData[style].color
// etc.
```

The only thing is the `waitingConfirmations(Int)` which prevents putting everything into an easy lookup table/dict

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationCell.swift
```diff
@@ -0,0 +1,37 @@
...
+struct TransactionConfirmationCell: View {
+    let address: String
+    let style: TransactionConfiramtionStatusViewStyle
+
+    var body: some View {
+        VStack (alignment: .leading, spacing: 0) {
+            TransactionConfiramtionStatusView(style: style)
+
+            HStack {
+                if [TransactionConfiramtionStatusViewStyle.canceled, TransactionConfiramtionStatusViewStyle.executed].contains(style) {
```
DmitryBespalov, line 20: please extract this to a named computed variable on the `TransactionConfiramtionStatusViewStyle` itself that would return true/false

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
```diff
@@ -0,0 +1,71 @@
...
+
+import SwiftUI
+
+struct TransactionConfirmationsView: View {
+    let transaction: BaseTransactionViewModel
+    let safe: Safe
+
+    var body: some View {
+        return VStack(alignment: .leading, spacing: 1) {
+            TransactionConfirmationCell(address: transaction.confirmations!.first!.address, style: .created)
```
DmitryBespalov, line 17: please add an assert before that somewhere in the initializer or in the body, something like:

```
assert(tranaction.confirmations?.count > 1)
```

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
```diff
@@ -0,0 +1,71 @@
...
+
+            ForEach(transaction.confirmations ?? [], id: \.address) { confirmation in
```
DmitryBespalov, line 19: I think better way would be to make the confirmations ad this view's variable and work with that because then, during initializaiton the confirmations can be set as an empty array and no need for this logic or implicit unwrap of the optionals multiple times.

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationCell.swift
```diff
@@ -0,0 +1,37 @@
...
+                    VerticalBarView().hidden()
+                }
+                else {
+                    VerticalBarView()
```
DmitryBespalov, line 24: instead of if-else, maybe use opacity? 

```
VerticalBarView().opacity(status.isTrue ? 1 : 0)
```

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsView.swift
```diff
@@ -0,0 +1,29 @@
...
+struct TransactionDetailsView: View {
+    @FetchRequest(fetchRequest: Safe.fetchRequest().selected())
+    var selectedSafe: FetchedResults<Safe>
+
+    let transaction: BaseTransactionViewModel
+    var body: some View {
+        List {
+            TransactionHeaderView(transaction: transaction)
+            if !(transaction.confirmations?.isEmpty ?? true) {
+                TransactionConfirmationsView(transaction: transaction, safe: selectedSafe.first!)
```
DmitryBespalov, line 20: if the selection suddenly changes to nil while this screen is opened, the app will crash

Multisig/UI/Transactions/TransactionsViewModel/TransactionsViewModel.swift
```diff
@@ -297,6 +324,16 @@ class TransactionsViewModel: ObservableObject {
         return result
     }
 
+    func confirmation(from confirmation: TransactionConfirmation) -> TransactionConfirmationViewModel {
+        let confirmationModel = TransactionConfirmationViewModel()
+        confirmationModel.address = confirmation.owner
+        confirmationModel.data = confirmation.data
+
+        confirmationModel.date = confirmation.submissionDate
+
```
DmitryBespalov, line 333: blank lines around this seem to be extra

Multisig/UI/UI Library/VerticalBarView.swift
```diff
@@ -0,0 +1,27 @@
...
+//  VerticalBarView.swift
+//  Multisig
+//
+//  Created by Moaaz on 6/11/20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import SwiftUI
+
+struct VerticalBarView: View {
```
DmitryBespalov, line 11: nice thing!

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
```diff
@@ -0,0 +1,71 @@
...
+                TransactionConfirmationCell(address: confirmation.address, style: .confirmed)
+            }
+
+            if transaction.status == .success {
+                TransactionConfirmationCell(address: transaction.confirmations!.last!.address, style: .executed)
```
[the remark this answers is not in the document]
MouazAlzahabi: yes, this was clearly ,mentioned 
https://gnosisinc.slack.com/archives/C63LQTGNB/p1591951807423900?thread_ts=1591882536.399400&cid=C63LQTGNB

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfiramtionStatusView.swift
MouazAlzahabi: Well, I think it still more clear to have the style as defined enum rather than lookup table, isn't it ?

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsView.swift
```diff
@@ -0,0 +1,29 @@
...
+
+struct TransactionDetailsView: View {
+    @FetchRequest(fetchRequest: Safe.fetchRequest().selected())
+    var selectedSafe: FetchedResults<Safe>
+
+    let transaction: BaseTransactionViewModel
+    var body: some View {
+        List {
+            TransactionHeaderView(transaction: transaction)
+            if !(transaction.confirmations?.isEmpty ?? true) {
```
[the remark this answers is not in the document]
MouazAlzahabi: You mean TransactionViewModel right

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsView.swift
MouazAlzahabi: Then there will be no transactions to be displayed right?

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
```diff
@@ -0,0 +1,71 @@
...
+        }
+        else if status == .waitingExecution {
+            return .execute
+        }
+
+        return nil
+    }
+
+    var isConfirmed: Bool {
+        return transaction.confirmations!.first { (confirmation) -> Bool in
```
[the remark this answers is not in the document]
MouazAlzahabi: I need to check if the safe address is one of confirmers so this tx already confirmed by the current safe or not, I will rename the method

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationCell.swift
MouazAlzahabi: don't you think this will be specific thing , I mean it used only one time

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
MouazAlzahabi: the problem here that in some other parts we need the transaction information to create the other parts

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfiramtionStatusView.swift
DmitryBespalov: I thought that the lookup table will be more compact in terms of lines of code. In this specific case, the switches are very simple, and clear, indeed, but for some of them there is code duplication. 


Even then, this can be improved:
```
switch style {
case .cancelled:
  return .gnoDarkBlue
case .waitingForConfirmations:
  return .gnoMediumGrey
case .rejected, .failed:
  return .gnoTomato
default:
  return .gnoHold
}

File: Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfiramtionStatusView.swift [hunk 13]
```diff
@@ -0,0 +1,101 @@
...
+        case .canceled:
+            return "ico-confirm"
+        case .failed:
+            return "ico-reject"
+        case .waitingConfirmations(_):
+            return "ico-empty-circle"
+        case .confirm:
+            return "ico-empty-circle"
+        case .execute:
+            return "ico-empty-circle"
```
DmitryBespalov, line 63: same here, multiple values are repeated and should be brought into a single case with multiple patterns:

```
case .created, .executed, .confirmed:
  ...
case .confirm, .execute:
  ...
```

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationCell.swift
DmitryBespalov: this is for readability, if there's no logical name for it, I'd create then a variable on this view with the name like `hasVerticalBar`

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
DmitryBespalov: Now I saw it!

Multisig/UI/Transactions/TransactionDetails/TransactionConfirmationsView/TransactionConfirmationsView.swift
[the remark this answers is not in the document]
DmitryBespalov: em.. but confirmations are not by the safe itself but by the safe's owners, now I am even more confused :smile:

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsView.swift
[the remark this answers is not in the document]
DmitryBespalov: yes, into the model

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsView.swift
DmitryBespalov: I was thinking about a situation when this screen is already open, and then because of push notification the app switches to a new safe, and then let's say there's a bug... now I see that it becomes very unlikely, but still. At your judgement. 😄

PR #236, opened 2020-06-16
Gh 95 tracking

### LINKED ISSUE #95: [iOS] Tracking
[+] Handling firebase configuration files (use encrypted files)
[+] Handling firebase configurations for different envs
[+] Add firebase to the project
[+] Add pod file
[+] Copy Tracker files from the old project (console/firebase tracking), or use some 3rd party lib 
[+] Setup user properties: - num_safes
[+] Disable personalized tracking
[+] Check on automatic/manual screen tracking (experiment with firebase debug view)
[+] Align with Android & product on tracking events names for screens (Should we be aligned with Android?)

# What to test
The tracking events are listed here: https://docs.google.com/spreadsheets/d/1KVaeOZ21aaNgR2jc5JM69gj0NWaOdOhlyhYXUMzh1l4/edit#gid=783259250

Only screen events are implemented. 

When the screen is shown the first time, or after it was covered by another screen, the event should be fired.

sche:
closes #95


Multisig/Cross-layer/Logger/LogService.swift
```diff
@@ -111,30 +111,30 @@ public final class LogService: Logger {
...
         }
         let normalizedEnabledLoggers = enabledLoggers.split(separator: ",").map {
             $0.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
         }
-        if normalizedEnabledLoggers.contains(LogServiceConsoleLoggerIdentifier) {
+        if normalizedEnabledLoggers.contains(ConsoleLoggerIdentifier) {
             add(ConsoleLogger())
         }
-//        if normalizedEnabledLoggers.contains(LogServiceCrashlyticsLoggerIdentifier) {
+//        if normalizedEnabledLoggers.contains(CrashlyticsLoggerIdentifier) {
```
DmitryBespalov, line 137: Maybe adding `#warning("TODO: re-enable with crashlytics implementation")`?

Multisig/Logic/Models/Safe.swift
```diff
@@ -23,6 +24,14 @@ extension Safe: Identifiable {
 
     var displayENSName: String { "" }
 
+    static var count: Int {
+        let context = App.shared.coreDataStack.viewContext
+        if let count = try? context.count(for: Safe.fetchRequest().all()) {
```
DmitryBespalov, line 29: one liner:

```swift
return (try? context.count(for: Safe.fetchRequest().all())) ?? 0
```

sche:
>I see the classes are added to the "Tracker", I created the "Analytics" folder already, now I think they are serving the same needs. Maybe merge/move the Tracker to the Analytics? Otherwise, if rename/delete the Analytics folder, the configuration and scripts need to be updated because of the Firebase config directory 😃

Okay, I moved files to the Analytics folder.

Multisig/Cross-layer/Logger/LogService.swift
sche: ok

Multisig/Logic/Models/Safe.swift
sche: ok

Issue #233, opened 2020-06-15
Restrict adding safe to the master copy >= 1.0.0


DmitryBespalov:
# How to test
- Try adding a safe that has master copy other than in the list:
  - 0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F
  - 0xaE32496491b53841efb51829d6f886387708F99B
  - 0xb6029EA3B2c51D09a50B53CA8012FeEB05bDa35A

For example, safes with the following master copies should not be addable:
 - 0x8942595A2dC5181Df0465AF0D7be08c8f23C93af

Example of a safe that has unknown master copy:
 - 0xb19BDaFf408bB502Ae348aF731C3812670667224

For such safes, when the safe address is entered, the app will show error that the safe master copy is not supported.

liliya-soroka:
Verified.
app version 0.5.0 (65)

PR #238, opened 2020-06-17
Gh 229/refactor

DmitryBespalov:
The PR is big, sorry about that.

Refactored and rewrote the JSON -> View Models, moving it into separate files and classes.

More unit tests to be added in another PR.

New type aliases introduced:
- `BigInt` -> `Int256`
- `BigUInt` -> `UInt256`

Whenever you need to use address, use the `Address`. If you need it formatted, please use its `checksummed` variable.

New types that handle our backend's String-based types:
- `AddressString` - converts the JSON String to Address
- `UInt256String` - converts the JSON String or Number to UInt256
- `DataString` - converts the JSON hex String to Data

I'm a bit confused now with the placement of such types as `Transaction`, `TokenInfo` and others. We use them for both decoding and as models, so they will change for different reasons, therefore I am for separating them into model's Transaction and backend's Transaction types. On the other hand, this would introduce the boilerplate as we need to copy the values from the json Transaction model to our model. I decided to wait for now and don't change things there.


Multisig/Logic/Ethereum/GnosisSafe.swift
```diff
@@ -17,10 +17,10 @@ class GnosisSafe {
     }
 
     var versions: [(masterCopy: Address, version: String)] = [
-        (Address("0xAC6072986E985aaBE7804695EC2d8970Cf7541A2")!, "0.0.2"),
-        (Address("0x8942595A2dC5181Df0465AF0D7be08c8f23C93af")!, "0.1.0"),
-        (Address("0xb6029EA3B2c51D09a50B53CA8012FeEB05bDa35A")!, "1.0.0"),
-        (Address("0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F")!, "1.1.1")
+        ("0xAC6072986E985aaBE7804695EC2d8970Cf7541A2", "0.0.2"),
```
sche, line 20: we decided not to support versions <1.0.0

Multisig/Logic/Models/Transaction/MethodRegistry+ERC721.swift
```diff
@@ -0,0 +1,41 @@
...
+                    let to = data.parameters[1].addressValue,
+                    let tokenId = data.parameters[2].uint256Value else {
+                        return nil
+                }
+                (self.from, self.to, self.tokenId) = (from, to, tokenId)
+            }
+        }
+
+        static func isValid(_ tx: Transaction) -> Bool {
+            #warning("TODO: tx.contractInfo.type == ContractInfoType.ERC721")
```
sche, line 33: is it a server dependency?

Multisig/Logic/Models/Transaction/MethodRegistry.swift
```diff
@@ -0,0 +1,33 @@
...
+
+        init(_ name: String, _ params: String...) {
+            self.name = name
+            self.parameterTypes = params
+        }
+    }
+
+}
+
+func == (lhs: TransactionData, signature: MethodRegistry.MethodSignature) -> Bool {
```
sche, line 26: why "lhs & signature" but not "lhs & rhs" or "transaction & signature"?

Multisig/Logic/Ethereum/GnosisSafe.swift
DmitryBespalov: not in this PR/ticket, there's a separate issue #233

Multisig/Logic/Models/Transaction/MethodRegistry+ERC721.swift
DmitryBespalov: yes, still waiting. This will affect token fetching

Multisig/Logic/Models/Transaction/MethodRegistry.swift
DmitryBespalov: No specific reason, just a preference, I thought they are not the same types, so I'd name them different 😃

Multisig/UI/Assets/TokenBalance.swift
```diff
@@ -0,0 +1,51 @@
...
+
+import Foundation
+import SwiftCryptoTokenFormatter
+
+struct TokenBalance: Identifiable, Hashable {
+    var id: String {
+        address
+    }
+    var imageURL: URL? {
+        // will be replaced when https://github.com/gnosis/safe-transaction-service/issues/86 is ready
```
sche, line 17: could you also add a warning here?

MultisigTests/Logic/Models/GnosisSafeTests.swift
```diff
@@ -11,7 +11,7 @@ import XCTest
 
 class GnosisSafeTests: XCTestCase {
 
-    let someAddress = Address("0x4995D78E5a672CC035929822E97EAcEB4464f97A")!
+    let someAddress: Address = "0x4995D78E5a672CC035929822E97EAcEB4464f97A"
```
sche, line 14: nice :)

MultisigTests/Logic/Models/TransactionTests.swift
```diff
@@ -0,0 +1,123 @@
...
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import XCTest
+@testable import Multisig
+
+class TransactionTests: XCTestCase {
+
+    func testAddOwner() {
+        let txJson = """
```
sche, line 15: I would extract JSONs into a separate file. Otherwise, it is hard to read.

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -88,39 +92,41 @@ struct SafeBalancesRequest: JSONRequest {
...
+            let decimals: UInt256String
         }
     }
 }
 
 struct TransactionsRequest: JSONRequest {
     let address: String
-    let limit: Int = 100
-    let offset: Int = 0
+    let limit: Int
```
MouazAlzahabi, line 115: We used to have PagedResponse, can we use it here ?

MultisigTests/Logic/Models/TransactionTests.swift
DmitryBespalov: Please ignore this, this test is work in progress, I'll do that and other tests in the following PR 🙏

DmitryBespalov:
@sche thank you, this won't happen anymore - I'll find a way to break up large PR into smaller batches.

Multisig/UI/Assets/TokenBalance.swift
DmitryBespalov: will do in the follow-up PR

PR #241, opened 2020-06-17
Gh 229/unit test erc20 - batch 1

DmitryBespalov:
Unit tests, batch number 1.

Remember the "var" vs. "let" discussion? Now I hit this wall: if you use "let", the struct doesn't let you (pun intended) to have generated initializer with just the properties you need, ignoring all nil properties - you have to provide all properties that are declared as "let". So I changed to "var".

### DIFF
--- a/MultisigTests/Logic/Models/MethodRegistryTestCase.swift
+++ b/MultisigTests/Logic/Models/MethodRegistryTestCase.swift
@@ -1,0 +1,22 @@
+//
+//  MethodRegistryTestCase.swift
+//  MultisigTests
+//
+//  Created by Dmitry Bespalov on 17.06.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import XCTest
+@testable import Multisig
+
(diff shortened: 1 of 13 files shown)


MultisigTests/Logic/Models/MethodRegistryTestCase.swift
```diff
@@ -0,0 +1,22 @@
...
+@testable import Multisig
+
+class MethodRegistryTestCase: XCTestCase {
+
+    typealias RawTxData = (label: String, method: String, params: [(name: String, type: String, value: String?)])
+
+    func txData(_ d: RawTxData) -> TransactionData {
+        TransactionData(
+            method: d.method,
+            parameters: d.params.map(TransactionDataParameter.init(name:type:value:))  )
```
sche, line 19: cool. So will it take $0, $1, $2 for name/type/value?

DmitryBespalov: yes 😄

Issue #202, opened 2020-06-03
Fix ENS name property in Safe


sche:
displayENSName always returns an empty string

DmitryBespalov:
double check reverse resolution

liliya-soroka:
Verified
app version 0.5.0 (65)

PR #246, opened 2020-06-17
gh-202 fix reverse resolution

sche:
this can be tested on tttk.eth


Multisig/Logic/Ethereum/ENS.swift
```diff
@@ -63,7 +63,7 @@ final class ENS {
 
     func name(for address: Address) throws -> String? {
         // construct a reverse node
-        let addressString = address.hexadecimal
+        let addressString = String(address.hexadecimal.dropFirst(2)) // skip 0x
```
DmitryBespalov, line 66: I think, it is better to work with data directly than manipulate strings here. I'd do the following:

```swift
extension Address {
  var data: Data {
      Data(rawAddress)
  }
}

// and on line 66 here:
let addressString = address.data.toHexString()
```

sche: ok

Issue #251, opened 2020-06-18
[iOS] Display transaction confirmation view


MouazAlzahabi:
Confirmation view should display: 
- Created 
- Confirmations (if exist otherwise display Confirm)
- Executer (if exists otherwise display the state) 
# Updates
- Incoming transactions should not have confirmations shown

liliya-soroka:
verified 0.5.0 (62)

PR #254, opened 2020-06-18
gh-251 Display executed, confirmations

MouazAlzahabi:


Multisig/UI/Transactions/TransactionConfirmationsView/TransactionConfirmationsView.swift
```diff
@@ -55,17 +56,15 @@ struct TransactionConfirmationsView: View {
...
 
         return nil
     }
 
-    var isConfirmedBySelectedSafe: Bool {
-        transaction.confirmations!.first { (confirmation) -> Bool in
-            confirmation.address == safe.address
-        } != nil
+    var isConfirmed: Bool {
+        !transaction.confirmations!.isEmpty
```
DmitryBespalov, line 68: we have the "hasConfirmations" that can be used here, right?

Issue #192, opened 2020-05-27
Padding issue


tschubotz:
**Describe the bug**
There is not enough space left and right of the screen (see screenshots)

**To Reproduce**
Open the app, visible on every screen.

**Expected behavior**
Nice UI.

**Screenshots**
<img src="https://user-images.githubusercontent.com/1380529/83003393-8afc5380-a00e-11ea-88d1-6e6fbc3c257e.PNG" width="200" /><img src="https://user-images.githubusercontent.com/1380529/83003412-92bbf800-a00e-11ea-85ce-be16c68d8ef0.PNG" width="200" />

**Smartphone (please complete the following information):**
 - Device: iPhone SE
 - iOS Version: iOS 13.5
 - App Version: 0.3.0 (1)

PR #260, opened 2020-06-18
gh-192  that fixed it

DmitryBespalov:
Don't mess with the NavigationView, they said 😆


Multisig/UI/UI Library/Frame/FrameForNavigationBar.swift
```diff
@@ -13,7 +13,7 @@ struct FrameForNavigationBar: ViewModifier {
     func body(content: Content) -> some View {
         // fixes unbounded growth of this view when inside the bar
         content
-            .frame(width: 270, height: 44, alignment: .leading)
+            .frame(width: 250, height: 44, alignment: .leading)
```
sche, line 16: omg :)

Issue #294, opened 2020-06-22
Auto-loader component for the network-based screens


DmitryBespalov:
- Refactor/create a component with the common logic for the **initial** data loading. 
  - Automatically loads the data on the app start
  - Automatically reloads the data when the app comes from background to foreground
    - Question: isn't that annoying?
  - Create full-screen "data can't be loaded" error message (https://zpl.io/bLlPkeQ)
  - Show that error screen in case the data loading fails for whatever reason.

liliya-soroka:
Reopened:
**Actual result**
"Data can't be loaded" is displayed after calling the app from the background after the internet connection is turned on. Note: Loading spinner is displayed before "data can't be loaded" message displaying

**Steps to reproduce**
1. Run the app ( balance, transactions list or settings)
2. Turn the internet connection off
3. send the app to the backgroung and call it back
Result: The page is updated with " data can't be loaded - The correct result

4. Send app to the background and turn internet connection on
5. Call the app from background in a few min
Result: "data can't be loaded" still displayed after loading spinner

6. I have tried a few times to call the app from background and also waited for 5 min between calling =>Data can't be loaded is displayed.

The page is updated without any issue if the other tab is opened or the app is restarted


App version: 0.5.0(67)
Device: iPhone X /13.15.1

liliya-soroka:
The new issue is created : https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/333

liliya-soroka:
Verified.
The task is closed.
The new issue is created and will be processed in the next sprints ( requires discussion)

app version 0.6.0 (73)

PR #314, opened 2020-06-24
gh-294 generalise loadable views

sche:
I created LoadableView and used it in every place where we show a loading screen.
No new logic is added in this PR, refactoring only.


DmitryBespalov (approved):
👏

Multisig/UI/UI Library/LoadableView.swift
```diff
@@ -0,0 +1,43 @@
...
+    var isLoading: Bool { get set }
+    var errorMessage: String? { get set }
+}
+
+protocol LoadableView: View {
+    associatedtype LoadableModel: LoadableViewModel
+    var model: LoadableModel { get }
+}
+
+struct Loadable<V: LoadableView>: View {
```
MouazAlzahabi: I think we should swap between Loadable& LoadableView names, Loadable is more like protocol name rather than view's

MouazAlzahabi (approved):

Multisig/UI/UI Library/LoadableView.swift
sche: I thought about it too. But `Loadable(SomeView())` looks organic for me as. I think it is better than `LoadableView(SomeView())`
But I don't have a strong opinion here. @DmitryBespalov wdyt?

DmitryBespalov: I'd go with Mouaz's suggestion here, because also will be more consistent with other types we have

sche: So what would be your suggestion here?
`LoadableViewModel `
`Loadable` (former LoadableView)
`LoadableView` (former Loadable)

I think it will be more confusing. Or did I get it wrong @MouazAlzahabi ?

DmitryBespalov: yes, like that

DmitryBespalov: and rename the "V" to the "Content" - also "view" variable can be named "content"

sche: all done

PR #317, opened 2020-06-24
Gh 294 reload data and error view

sche:
- reload data on entering foreground
- "Data cannot be loaded" view on errors (design approved by Sasha)

IMG_E9DEF5BAF5D1-1
IMG_FE832E054BCF-1


Multisig/UI/UI Library/LoadableView.swift
```diff
@@ -34,10 +35,25 @@ struct Loadable<V: LoadableView>: View {
...
             } else {
                 view
             }
         }
+        .onReceive(NotificationCenter.default.publisher(for: UIApplication.willEnterForegroundNotification)) { _ in
+            self.model.reloadData()
+        }
+    }
+
+    var noDataView: some View {
```
DmitryBespalov, line 48: would be good to extract it in a separate piece, but it's ok for now.

DmitryBespalov (approved):

Multisig/UI/UI Library/LoadableView.swift
```diff
@@ -34,10 +35,25 @@ struct Loadable<V: LoadableView>: View {
...
+        VStack {
+            HStack {
+                Image("ico-server-error")
+                Text("Data cannot be loaded").font(Font.gnoTitle3).foregroundColor(.gnoMediumGrey)
```
MouazAlzahabi: Can you please use "TitleText" view instead?
Just update it to accept the color as init parameter like what we have in BodyText.swift

MouazAlzahabi (approved):

Multisig/UI/UI Library/LoadableView.swift
sche: done

Issue #290, opened 2020-06-22
Fetch next transaction page from the backend


DmitryBespalov:
- Backend response has "next" and "previous" links. Use the link as it is. This means we need a new request or adapt existing one. 
- The offsets/limits or next/previous might have edge cases when the new transaction comes into response, and so the next page can have duplicate results. The last item of one page can then be the first item of the next page, as the whole response will shift by the number of transactions added to the list.

liliya-soroka:
Verified
app version 0.6.0(73)

PR #319, opened 2020-06-25
gh-290 Load transactions by url

MouazAlzahabi:


Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -111,22 +111,29 @@ struct SafeBalancesRequest: JSONRequest {
...
-    let address: String
-    let limit: Int
-    let offset: Int
+    let address: String?
+    let limit: Int?
+    let offset: Int?
     var httpMethod: String { "GET" }
-    var urlPath: String { "/api/v1/safes/\(address)/all-transactions/" }
+    var urlPath: String {
+        guard let address = address else { return "" }
```
sche: how about taking it from 'url'? There there will be no need to introduce 'url' in many other places.

Multisig/Data/Services/Utils/JSONHTTPClient.swift
```diff
@@ -14,13 +14,15 @@ public protocol JSONRequest: Encodable {
     /// Query parameters
     var query: String? { get }
 
+    var url: String? { get }
```
DmitryBespalov: I think, spacing is missing after this line :) . Also please add a documentation comment that explains how this request will be treated differently if the url present vs when it is not present

Multisig/Data/Services/Utils/HTTPClient.swift
```diff
@@ -78,6 +79,10 @@ public class HTTPClient {
     }
 
     private func urlRequest<T: HTTPRequest>(from request: T) throws -> URLRequest {
+        if let stringURL = request.url, let url = URL(string: stringURL)  {
+            return URLRequest(url: url)
```
DmitryBespalov: can this be a "POST" request? then it is not handled here

Multisig/Data/Services/SafeTransactionService.swift
DmitryBespalov: I like it. ~But that means we need to parse the URL itself, which is not desired~ I actually prefer that

DmitryBespalov: the way would be to URL -> URLComponents -> use in the "urlPath" and in "query"

MouazAlzahabi: But this means we need to know how to split the url, which I think we should avoid to count on

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -111,22 +115,29 @@ struct SafeBalancesRequest: JSONRequest {
...
     }
+
+    var url: String?
     
     typealias Response = PagedResponse<Transaction>
     typealias ResponseType = Response
 
-    init(address: Address, limit: Int, offset: Int) {
-        self.address = address.checksummed
+    init(address: Address?, limit: Int?, offset: Int?, url: String? = nil) {
```
DmitryBespalov: is the "url" used anywhere? don't see it is used now

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -111,22 +115,29 @@ struct SafeBalancesRequest: JSONRequest {
 }
 
 struct TransactionsRequest: JSONRequest {
-    let address: String
-    let limit: Int
-    let offset: Int
+    let address: String?
```
DmitryBespalov: Do we need these optionals now or it can be reverted?

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -55,7 +55,11 @@ class SafeTransactionService {
         return try httpClient.execute(request: SafeBalancesRequest(address: address))
     }
 
-    func transactions(address: Address, offset: Int = 0, limit: Int = 100) throws -> TransactionsRequest.Response {
+    func transactions(address: Address?, offset: Int? = 0, limit: Int? = 20, url: String? = nil) throws -> TransactionsRequest.Response {
```
DmitryBespalov: I'd create 2 different methods instead - one with all parameters, one with just the URL, because they're mutually exclusive

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -55,8 +55,16 @@ class SafeTransactionService {
...
-        return try httpClient.execute(request: TransactionsRequest(address: address, limit: limit, offset: offset))
+    func transactions(address: Address?, offset: Int = 0, limit: Int = 20) throws -> TransactionsRequest.Response {
+        return try httpClient.execute(request: TransactionsRequest(address: address!, limit: limit, offset: offset))
+    }
+
+    func loadTransactionsPage(url: String) throws -> TransactionsRequest.Response {
+        let request = PagedRequest<Transaction>(url)
+
+        assert(request != nil)
+
```
DmitryBespalov: imo the blank lines are not needed - does it reads better for you with them? @sche @MouazAlzahabi

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -115,11 +123,15 @@ struct TransactionsRequest: JSONRequest {
...
+    var urlPath: String {
+        "/api/v1/safes/\(address)/all-transactions/"
+    }
 
     var query: String? {
-        return "limit=\(limit)&offset=\(offset)&queued=true"
+        "limit=\(limit)&offset=\(offset)&queued=true"
     }
+
+    var url: String?
```
DmitryBespalov: is this variable used anywhere?

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -55,8 +55,16 @@ class SafeTransactionService {
         return try httpClient.execute(request: SafeBalancesRequest(address: address))
     }
 
-    func transactions(address: Address, offset: Int = 0, limit: Int = 100) throws -> TransactionsRequest.Response {
-        return try httpClient.execute(request: TransactionsRequest(address: address, limit: limit, offset: offset))
+    func transactions(address: Address?, offset: Int = 0, limit: Int = 20) throws -> TransactionsRequest.Response {
+        return try httpClient.execute(request: TransactionsRequest(address: address!, limit: limit, offset: offset))
+    }
+
+    func loadTransactionsPage(url: String) throws -> TransactionsRequest.Response {
```
DmitryBespalov: Just curious, is there a reason for using the url instead of the `nextPageRequest` from the previous response?

Multisig/Data/Services/SafeTransactionService.swift
MouazAlzahabi: I like to split "paragraphs" with lines.

Multisig/Data/Services/SafeTransactionService.swift
MouazAlzahabi: Yes, I didn't want to have reference for the request (which is part of logic layer) in view model, I preferred to have reference for the url as strings,

Multisig/Data/Services/SafeTransactionService.swift
DmitryBespalov: what I was thinking is these are just 3 lines, not paragraphs, so they go together

Multisig/Data/Services/SafeTransactionService.swift
DmitryBespalov: ... then probably those "nextRequest" and "previousRequest" - are they needed at all?

MouazAlzahabi: No not used, but I thought to keep them for now in case we changed something later

Multisig/Data/Services/SafeTransactionService.swift
```diff
@@ -55,8 +55,16 @@ class SafeTransactionService {
...
+        let request = PagedRequest<Transaction>(url)
+
+        assert(request != nil)
+
+        return try httpClient.execute(request: request!)
```
DmitryBespalov: I'd prefer not to crash the app in production, if the url returned by backend is invalid. I think it is an expected case and must be handled appropriately

DmitryBespalov (changes requested):

Multisig/Data/Services/SafeTransactionService.swift
sche: I would also remove blank lines here

Multisig/Data/Services/SafeTransactionService.swift
sche: I would use `let request = PagedRequest<Transaction>(url)!`
If it crashes here, I would treat it as a developer error

Multisig/Data/Services/Utils/JSONHTTPClient.swift
```diff
@@ -86,6 +86,7 @@ public class JSONHTTPClient {
         return try response(from: data)
     }
 
+
```
sche: Why did you put an extra space here?

Multisig/UI/Transactions/TransactionsList/TransactionsViewModel.swift
```diff
@@ -13,6 +13,13 @@ class TransactionsViewModel: LoadableViewModel {
     @Published var transactionsList = TransactionsListViewModel()
     @Published var isLoading: Bool = true
     @Published var errorMessage: String? = nil
+    @Published var isLoadingNextPage: Bool = true
+
+    var viewState = App.shared.viewState
+
+    
+    private var prevURL: String?
```
DmitryBespalov: please use the full name for this variable that has the class scope: `previousURL`

DmitryBespalov: is the previous URL used anywhere? do we need it at all?

Multisig/UI/Transactions/TransactionsList/TransactionsViewModel.swift
```diff
@@ -58,4 +69,44 @@ class TransactionsViewModel: LoadableViewModel {
...
+    func loadNextPage() {
+        subscribers.forEach { $0.cancel() }
+        isLoadingNextPage = true
+        Just(nextURL)
+            .setFailureType(to: Error.self)
+            .flatMap { url in
+                Future<[TransactionViewModel], Error> { promise in
+                    DispatchQueue.global().async {
+                        do {
+                            if let transactions = try App.shared.safeTransactionService.loadTransactionsPage(url: url!) {
```
DmitryBespalov: because of `url!`, I'd add the ".compactMap { $0 }" after the "Just()"

DmitryBespalov (changes requested):

DmitryBespalov:
on the safe `0xEefFcdEAB4AC6005E90566B08EAda3994A573C1E` the list is always displaying the loading indicator (if scrolled after the app launch)

same for '0xCf0eC5CDc7905b6aBD72B7Ac67688dd056Ede5E1'

Multisig/UI/Transactions/TransactionsList/TransactionsViewModel.swift
MouazAlzahabi: It's returned from the backend but it's not used

DmitryBespalov (approved):

DmitryBespalov (changes requested):
crashing on the safe 0xaE3c91c89153DEaC332Ab7BBd167164978638c30

DmitryBespalov (approved):

Issue #310, opened 2020-06-24
Implement shownig and auto-hiding of the snackbar messages


DmitryBespalov:
The message  automatically hides after 4 seconds.

How to check: 

Assets -> Collectibles -> Show button

liliya-soroka:
verified
app version 0.6.0 (73)

PR #321, opened 2020-06-26
gh-310  added message hiding

DmitryBespalov:
Simply hides the message after 4 seconds.

The multiple notification will be handled (if required) in another task.

### DIFF
--- a/Multisig/UI/App/ViewState.swift
+++ b/Multisig/UI/App/ViewState.swift
@@ -7,8 +7,11 @@
 //
 
 import SwiftUI
+import Combine
 
 class ViewState: ObservableObject {
+
+    // Active tab
 
     @Published
     private(set) var state: ViewStateMode? = .balances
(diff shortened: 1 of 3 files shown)


Multisig/UI/App/ViewState.swift
```diff
@@ -7,19 +7,26 @@
 //
 
 import SwiftUI
+import Combine
 
 class ViewState: ObservableObject {
 
+    // Active tab
```
sche, line 14: MARK ?

Multisig/UI/App/ViewState.swift
```diff
@@ -29,14 +36,24 @@ class ViewState: ObservableObject {
...
-            snackbarMessge = nil
-            showsSnackbar = false
-        }
+    private let messageHiding = PassthroughSubject<Void, Never>()
+    private var subscribers = Set<AnyCancellable>()
+    private let messageDuration: TimeInterval = 4
+
+    init() {
+        messageHiding
+            .debounce(for: .seconds(messageDuration), scheduler: RunLoop.main)
```
sche, line 45: nice :)

sche (approved):

PR #344, opened 2020-07-08
Gh 179 crashlytics

sche:
Please pay attention to "project.pbxproj"
As I understand, it should upload DSYM file, but we can test it only when the new build is ready in TestFlight.


Multisig/Cross-layer/Logger/LogService.swift
```diff
@@ -118,26 +118,21 @@ public final class LogService: Logger {
...
+        addLoggers()
     }
 
-    private func addLoggers(from bundle: BundleProtocol) {
-        guard let enabledLoggers = bundle.object(forInfoDictionaryKey: LoggersKey) as? String else {
-            return
-        }
-        let normalizedEnabledLoggers = enabledLoggers.split(separator: ",").map {
+    private func addLoggers() {
+        let normalizedEnabledLoggers = App.configuration.app.loggers.split(separator: ",").map {
```
MouazAlzahabi, line 127: App.configuration.app is used twice, it's good if we define it as private field

MouazAlzahabi (approved):

Multisig/Cross-layer/Logger/LogService.swift
sche: I think in this case it is okay not to have it as a private field. This would just add an extra line of code :)
In our prev project, we had a very flexible approach to moving such things to a private field. In the majority of the cases, we did not do it. But we can discuss it during our learning group meeting.

DmitryBespalov:
~Note: this needs to be tested for the case when new git clone doesn't have the Firebase configuration files. The project should still build and run.~

All clear.

Multisig.xcodeproj/project.pbxproj
```diff
@@ -2104,7 +2103,7 @@
 			buildSettings = {
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 3;
```
DmitryBespalov: no needed to commit

DmitryBespalov (approved):

PR #345, opened 2020-07-08
Gh 328/transactions pagination after loading scrolls to the first item

MouazAlzahabi:


Multisig/UI/Transactions/TransactionsList/TransactionsListViewModel.swift
```diff
@@ -19,37 +19,42 @@ struct TransactionsListViewModel {
...
-                queued.append(model)
-            } else {
-                history.append(model)
-            }
-        }
+       for model in models {
+           if model.status.isInQueue {
+               allSections[0].transactions.append(model)
+           } else {
+               allSections[1].transactions.append(model)
```
DmitryBespalov, line 55: didn't it cause crash before?

DmitryBespalov: `assert(allSections.count > 1)`

MouazAlzahabi: Please see the init function, previously we used to filter the empty sections and that's why it cased crash, now I'm filtering them only in the getter "sections"

DmitryBespalov (approved):

PR #359, opened 2020-07-09
gh-357  refactored tx service into separate request per file

DmitryBespalov:
I want to make the SafeTransactionService extendable and modularize the requests.

I extracted each request of transaction service into a separate one


Multisig/Data/Services/Utils/DateDecodingStrategy+GnosisSafeServices.swift
```diff
@@ -0,0 +1,47 @@
...
+                return date
+            } else if let date = Self.backendFormatter2.date(from: str) {
+                return date
+            } else {
+                let context = DecodingError.Context(
+                    codingPath: decoder.codingPath,
+                    debugDescription: "Date! \(str)")
+                throw DecodingError.dataCorrupted(context)
+            }
+        }
```
MouazAlzahabi, line 45: Seems we need to fix the alignment here

MouazAlzahabi (approved):

Multisig/Data/Services/Utils/DateDecodingStrategy+GnosisSafeServices.swift
DmitryBespalov: ok

PR #372, opened 2020-07-15
gh-353  updated texts

DmitryBespalov:
Removed our simple text wrapper views and instead added "modifier"-like methods to the standard `Text`. Also resolved inconsistencies in how different font sizes and colors were used.


sche (approved):

Multisig/UI/Onboarding/TermsView.swift
```diff
@@ -64,7 +65,7 @@ struct TermsView: View {
                     .foregroundColor(.gnoHold)
                     .padding(.top, bulletTopPadding)
                 Text(text)
-                    .font(Font.gnoHeadline.weight(.medium))
+                    .font(.gnoBody)
```
MouazAlzahabi, line 68: Why don't we use .body() modifier here ?

Multisig/UI/Safe Management/Switch Safe/AddSafeRow.swift
```diff
@@ -20,7 +20,7 @@ struct AddSafeRow: View {
             HStack(spacing: 6) {
                 Image.plusCircle
 
-                Text("Add Safe").font(Font.gnoBody.weight(.medium))
+                Text("Add Safe").font(.gnoBody)
```
MouazAlzahabi, line 23: .body() we can define color input parameter

MouazAlzahabi: I think you already did it right?

Multisig/UI/Settings/SettingsView.swift
```diff
@@ -20,7 +20,7 @@ struct SettingsView: View {
                     HStack {
                         Image("ico-safe-settings").frame(width: 24, height: 24)
                         Text("SAFE SETTINGS")
-							.font(Font.gnoCaption1)
```
MouazAlzahabi, line 23: I wish we can define caption text also

Multisig/UI/UI Library/Styleguide/Fonts.swift
```diff
@@ -9,13 +9,48 @@
...
+    static let gnoHeadline =    system(size: 17, weight: .semibold)
+    static let gnoHeadline2 =   system(size: 17, weight: .bold)
+
+    static let gnoTitle3 =      system(size: 20)
+
+    static let gnoNormal =      custom("Averta Regular", size: 26)
+}
+
+// common text stylings
+extension Text {
```
MouazAlzahabi, line 33: I still think it's better to define it as modifier, because in that case we can compose them, doesn't it ?

MouazAlzahabi (approved):

Multisig/UI/Onboarding/TermsView.swift
DmitryBespalov: because the default color is not needed

Multisig/UI/Safe Management/Switch Safe/AddSafeRow.swift
DmitryBespalov: not using `body()` because otherwise the `foregroundColor` is not working when applied to the whole hstack

Multisig/UI/UI Library/Styleguide/Fonts.swift
DmitryBespalov: the problem with applying modifiers to the text is that it creates new type, which is not Text, and so I can't compose some Text-related modifiers with a custom modifier 😞

Issue #335, opened 2020-07-03
[iOS] Transactions list: No error handling during the scrolling to the previous page


liliya-soroka:
**Bug description**
There is no any indication that the previous page can not be loaded in case of the problems with data loading ( internet connection, back-end errors)
**Steps To Reproduce**
1. Open transactions list
2. Go to assets
3. Turn internet connection off/ send error from the back-end
4. come back to the transaction 
5. scroll down to load previous page 
Current result: Nothing on UI. The user will try to load but nothing is happened
**Expected Result**
The error message should be shown

**Device & App version**
 - Device/OS: iPhone X /13.15.1
 - App version: 0.6.0 (73)

**Additional info**
(Optional)

liliya-soroka:
Verified
app version 0.7.0 (93)

PR #375, opened 2020-07-16
gh-335 small fix

sche:
If pagination fails, it was not be triggered again because there was no "nextURL" anymore.
I changed it. Not use can re-trigger pagination by scrolling up and down again.

### DIFF
--- a/Multisig/UI/Transactions/TransactionsList/TransactionsViewModel.swift
+++ b/Multisig/UI/Transactions/TransactionsList/TransactionsViewModel.swift
@@ -89,7 +89,6 @@ class TransactionsViewModel: BasicLoadableViewModel {
             .receive(on: RunLoop.main)
             .sink(receiveCompletion: { completion in
                 if case .failure(let error) = completion {
-                    self.nextURL = nil
                     App.shared.snackbar.show(message: error.localizedDescription)
                 }
                 self.isLoadingNextPage = false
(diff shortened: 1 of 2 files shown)


DmitryBespalov (approved):

Multisig/UI/Transactions/TransactionsList/TransactionsViewModel.swift
```diff
@@ -89,7 +89,6 @@ class TransactionsViewModel: BasicLoadableViewModel {
             .receive(on: RunLoop.main)
             .sink(receiveCompletion: { completion in
                 if case .failure(let error) = completion {
-                    self.nextURL = nil
```
MouazAlzahabi, line 92: But this way if you scroll again to bottom it will trigger the pagination again, is it correct to do so ?

sche: I would assume this behavior. Otherwise, you can't re-trigger pagination at all, apart from refreshing the whole view.

DmitryBespalov: if it could decide based on the response's nextURL, and not on failure, then it would be ideal

sche: what do you mean?

DmitryBespalov: nevermind, it is already good (on the line 79)

Issue #368, opened 2020-07-14
[iOS]  Unknown fall back handler is displayed when it's not set up


liliya-soroka:
**Bug description**
Unknown fall back handler is displayed when it's not set up with the address 0x0000000
**Steps To Reproduce**
1. Add save alice.eth
2. Open safe settings->advanced

**Expected Result**
https://app.zeplin.io/project/5e8f27b9963e55bb79b13634/screen/5ebbb969cea0c078fd9ce43a
**Screenshots**

**Device & App version**
 - Device/OS: iPhone X/ 13.15.1
 - App version: 0.6.0(82)

**Additional info**
(Optional)

liliya-soroka:
Verified
not set is displayed
Prod Rinkeby 
app version 0.7.0 (100)

PR #400, opened 2020-07-22
gh-368 Fix issue

MouazAlzahabi:


Multisig/Logic/Models/Address.swift
```diff
@@ -68,6 +68,10 @@ struct Address: Hashable, ExpressibleByStringInterpolation, CustomStringConverti
         self == .zero
     }
 
+    var isEmpty: Bool {
```
DmitryBespalov: In my opinion, this should not be part of the Address type, but it should be a function in the GnosisSafe class, because this is a business logic that belongs there.

Please move it to the GnosisSafe class

Multisig/Logic/Models/Address.swift
```diff
@@ -68,6 +68,10 @@ struct Address: Hashable, ExpressibleByStringInterpolation, CustomStringConverti
...
+        isZero || checksummed.isEmpty || checksummed == "0"
```
DmitryBespalov: is it possible that the checksummed is empty or "0"?

Multisig/UI/Settings/Safe Settings/AdvancedSafeSettingsView.swift
```diff
@@ -40,15 +40,12 @@ struct AdvancedSafeSettingsView: View {
     }
     
     var fallbackHandlerView : some View {
-        let fallbackHandler = safe.fallbackHandler?.checksummed ?? ""
         let title = App.shared.gnosisSafe.fallbackHandlerLabel(fallbackHandler: safe.fallbackHandler)
-        return Group {
-            if fallbackHandler.isEmpty || fallbackHandler == "0" {
+        return Section(header: SectionHeader("FALLBACK HANDLER")) {
+            if safe.fallbackHandler?.isEmpty ?? true {
```
DmitryBespalov: the function to check that could be something like

```swift
.. GnosisSafe {

  ...

  func hasFallbackHandler(at address: Address?) -> Bool {
     address?.isZero == false 
  }

}
```

and then in this view file:

```swift
   if App.shared.gnosisSafe.hasFallbackHandler(at: safe.fallbackHandler) {
        AddressCell(...)
   else {
      Text("Not set")
  }
```

DmitryBespalov (changes requested):

Multisig/Logic/Models/Address.swift
MouazAlzahabi: @DmitryBespalov I think it should be part of address itself actually, I'm thinking about it as attribute of the address that can be used by other business logic parts, same as isZero, the current case is that it's used in the UI , what do you think?

Multisig/Logic/Models/Address.swift
MouazAlzahabi: Actually this condition I moved from AdvancedSafeSettingsView.swift as it is, so I assumed that it's possible, doesn't it?

DmitryBespalov: the `checksummed` is a derived property from the binary data, so whenever the Data is 0, the checksummed will always be the `0x000000......00`, so imo this is not possible - but we can check it with unit test 😉 

In that case, this isEmpty is not needed at all

MouazAlzahabi: Still we have the case that checksummed is empty text right ?

DmitryBespalov: no, it should never be an empty text, please write unit test and pass different values to check

Multisig/Logic/Ethereum/GnosisSafe.swift
```diff
@@ -28,6 +28,10 @@ class GnosisSafe {
 
     var fallbackHandlers:[(fallbackHandler: Address, label: String)] = [("0xd5D82B6aDDc9027B22dCA772Aa68D5d74cdBdF44", "DefaultFallbackHandler")]
 
+    func hasFallbackHandler(safe: Safe) -> Bool {
+        !(safe.fallbackHandler?.isZero ?? true)
```
DmitryBespalov: This is not easy to understand... Can you please simplify a bit to remove the negation of the ??

`safe.fallbackHandler.map { !$0.isZero } ?? false`

or

`safe.fallbackHandler?.isZero == false`

or 

`safe.fallbackHandler != nil  && safe.fallbackHandler != .zero`

DmitryBespalov (approved):

sche (approved):

Issue #402, opened 2020-07-22
[iOS]Prod Rinkeby:  ERC721 are displayed incorrectly in the transaction list


liliya-soroka:
**Bug description**
The tx with ERC721 have  0 as amount for all kind of ERC721 and id as the amount for CK
**Steps To Reproduce**
1. Load safes  liltestsafe.eth
2. Open transactions list
**Expected Result**
The correct amount should be displayed for ERC721
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/919b3569-a10a-4228-9796-8e691226bab5)
**Device & App version**
 - Device/OS: iPhone X / 13.15.1
 - App version: Prod Rinkeby - 0.7.0 (93)

**Additional info**
(Optional)

liliya-soroka:
Reopened:
Prod rinkeby
app version 0.7.0(99)
Incoming ERC721:
amount = 0 symbol
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/135b3857-c12d-4b69-9fa0-da81744c3aa7)

Outgoing ERC721 ( CK)
amount = ID ck instead of 1 
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/85f0016b-6461-4a00-bea2-4187cf64f2a6)

liliya-soroka:
**Verified:**
The amount for all ERC721 tokens are displayed correctly in the transactions list except ENS tokens
**Actual result for ENS token:**
incoming and outgoing amount = 1NFT
**Environment** 
Prod Rinkeby
App version 0.7.0 (100)

**Next steps:**
@tschubotz, I offer to close this issue and create a new one for ENS token only as not critical but major

tschubotz:
Sounds good to me  👌

liliya-soroka:
Verified.
There is no issue with ENS in the app version 0.7.0(104)

PR #414, opened 2020-07-24
gh-402  fixing erc721

DmitryBespalov:
This fixes the issue

### DIFF
--- a/Multisig/UI/Transactions/TransactionViewModel/TransferTransactionViewModel.swift
+++ b/Multisig/UI/Transactions/TransactionViewModel/TransferTransactionViewModel.swift
@@ -92,8 +92,7 @@ class TransferTransactionViewModel: TransactionViewModel {
     }
 
     fileprivate class func token(from transfer: TransactionTransfer) -> Token {
-        let token: Token
-
+        var token: Token
 
         switch transfer.type {
         case .ether, .unknown:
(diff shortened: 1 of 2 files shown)


sche (approved):

Multisig/UI/Transactions/TransactionViewModel/TransferTransactionViewModel.swift
```diff
@@ -104,6 +103,13 @@ class TransferTransactionViewModel: TransactionViewModel {
             } else if let address = transfer.tokenAddress?.address {
                 if let known = App.shared.tokenRegistry.token(address: address) {
                     token = known
+                    // Token type may be invalid because TokenRegistry
+                    // assumes "erc20" when it doesn't know the actual
+                    // type.
+                    // Now, with the information from the transfer, we
+                    // can enhance the cached token
+                    token.type = transfer.type == .erc20 ? .erc20 : .erc721
```
MouazAlzahabi, line 111: Not sure if it's valid idea but can we define same enum for transfer type& token type?

DmitryBespalov: they have different values, we could, but in a different task

DmitryBespalov: #415

MouazAlzahabi (approved):

Issue #288, opened 2020-06-22
Transaction details data loading


DmitryBespalov:
Make the transaction details data loadable from the backend.

This is about loading the data and networking requests.

Use the endpoint /transaction/hash

# How to test?
Nothing to test yet; unit tests added.

sche:
TODO: check the differences when getting the list, and with a param. @rmeissner

rmeissner:
As far as I can see the only information not returned is `transfer` and `type` all the other information are present.

DmitryBespalov:
> Uxio will check that non-safe transactions' data is downloadable

PR #421, opened 2020-07-28
gh-288  tx data loading

DmitryBespalov:
I could reuse the same transaction object


Multisig.xcodeproj/project.pbxproj
```diff
@@ -113,6 +113,7 @@
 		0A29ECAA24C1EBFD00CE28FD /* Address+LocalizedStringKey.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0A29ECA924C1EBFD00CE28FD /* Address+LocalizedStringKey.swift */; };
 		0A339C5D24C1D16E00DAD84B /* Image+Blockies.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0A339C5C24C1D16E00DAD84B /* Image+Blockies.swift */; };
 		0A3F34DB2477E4E900E872AF /* KeyboardAdaptive.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0A3F34D82477E4E900E872AF /* KeyboardAdaptive.swift */; };
+		0A42C25424D0696D00ACA770 /* GetTransactionRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0A42C25324D0696D00ACA770 /* GetTransactionRequest.swift */; };
```
MouazAlzahabi: Totally good, but I prefer to call it TransactionDetailsRequest instead

MultisigIntegrationTests/TokenRegistryIntegrationTests.swift
```diff
@@ -18,40 +18,44 @@ class TokenRegistryIntegrationTests: XCTestCase {
...
-            tokens.append(self.registry[AddressRegistry.ether])
-            // pull from backend (AQER)
-            tokens.append(self.registry[cryptoKitties])
-            // same token, cached
-            tokens.append(self.registry[cryptoKitties])
-            // pull from blockchain (LINK)
-            tokens.append(self.registry[chainLinkNotInBackend])
-            // not a token -- should be recognized as ERC20 (this is a fallback logic)
-            tokens.append(self.registry[notAToken])
+            tokens.append(self.registry[.ether])
```
MouazAlzahabi, line 46: As you explicitly capture self, do we still need to use self. ?

MouazAlzahabi (approved):

MultisigIntegrationTests/TokenRegistryIntegrationTests.swift
DmitryBespalov: yes, it's still not swift 5.2 :)

Multisig.xcodeproj/project.pbxproj
DmitryBespalov: changed

Issue #398, opened 2020-07-22
[iOS] Fatal errors are not tracked in the Crashlytics


liliya-soroka:
**Bug description**
Fatal errors are not tracked in the Crashlytics.
There is no event in the Crashlytics if the unexpected error is simulated in the mobile app using mitmproxy 

**Steps To Reproduce**
1. Use mitmproxy to simulate back-end errors 
2. check Crashlytics for events in case of unexpected errors in the app
**Expected Result**
All fatal errors should be handled by Crashlytics

**Device & App version**
 - Device/OS: iPhone X /13.15.1
 - App version: 0.7.0 (93)

**Additional info**
(Optional)

sche:
We use `crashlytics.record(error: NSError(domain: nsError.domain, code: nsError.code, userInfo: userInfo))`
It seems like this call does not track properly. I will check it.

DmitryBespalov:
I have changed the error logging so that they publish errors that are non-fatals in crashlytics. I have stepped through the code with debugger and verified that the code is actually sending errors. However, the crashlytics dashboard does not display them.

DmitryBespalov:
Postpone for now

DmitryBespalov:
should be testable in 2.0.0 (6)

liliya-soroka:
Verified
app version 2.1.0 (2)

PR #426, opened 2020-07-30
gh-398  added error arguments

DmitryBespalov:
The crashlytics is not logging to the server unless we specify the "error" parameter for the logger. I've added that.

### DIFF
--- a/Multisig/Data/Services/Utils/HTTPClient.swift
+++ b/Multisig/Data/Services/Utils/HTTPClient.swift
@@ -35,13 +35,6 @@ class HTTPClient {
         self.logger = logger
     }
 
-    fileprivate enum UnexpectedError: LoggableError {
-        case unrecognizedErrorCode(Int)
-        case missingDataInUnprocessableEntity
-        case errorDetailsDecodingFailed(String)
-        case unknownHTTPError(String)
-    }
-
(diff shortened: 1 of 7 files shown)


Multisig/Data/Services/Utils/HTTPClient.swift
```diff
@@ -35,6 +35,13 @@ class HTTPClient {
         self.logger = logger
     }
 
+    fileprivate enum UnexpectedError: LoggableError {
```
MouazAlzahabi: All good, but I prefer to find a way to define all those LoggableError enums in one place instead of define each couple values in separate file

MouazAlzahabi (approved):

Multisig/Data/Services/Utils/HTTPClient.swift
DmitryBespalov: I see. I think, good middle ground would be to put all such errors in the "Cross-layer" directory

MouazAlzahabi: something like so, because I think later we might add more, and keeping them dispersed make it hard to maintain

DmitryBespalov: ok, done

PR #440, opened 2020-08-06
Gh 434/display notification

DmitryBespalov:
- Prepared the app to receive the notifications. - this is work in progress, but I would like to merge it first
- Updated pods.


Multisig/UI/App/AppDelegate.swift
```diff
@@ -40,4 +42,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // If any sessions were discarded while the application was not running, this will be called shortly after application:didFinishLaunchingWithOptions.
         // Use this method to release any resources that were specific to the discarded scenes, as they will not return.
     }
+
+    func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
```
MouazAlzahabi, line 46: I would prefer to move this func to AppDelegate+Messaging.swift so we have everything related to notifications there

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -0,0 +1,164 @@
...
+extension UNAuthorizationStatus: CustomStringConvertible {
+
+    public var description: String {
+        switch self {
+        case .authorized: return "Authorized (granted)"
+        case .denied: return "Denied"
+        case .notDetermined: return "Not Determined"
+        case .provisional: return "Provisional (granted)"
+        @unknown default:
+            return "Unknown: \(rawValue)"
```
MouazAlzahabi: Why the default value is on two lines ?

DmitryBespalov: no specific reason

Multisig/UI/App/AppDelegate.swift
DmitryBespalov: This is not possible to move completely, because it is implementation of the UIApplicationDelegate protocol, it has to be here

MouazAlzahabi: true, my mistake

MouazAlzahabi (approved):

PR #447, opened 2020-08-07
Gh 433/notification service and triggering the register and unregister of notifications

MouazAlzahabi:


Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -0,0 +1,50 @@
...
+        let version: String
+        let deviceType: String = "IOS"
+        let buildNumber: String
+    }
+}
+
+extension SafeTransactionService {
+
+    @discardableResult
+    func register(deviceID: UUID? = nil, safes: [Address], token: String, bundle: String, version: String, buildNumber: String) throws -> RegisterNotificationTokenRequest.Response {
```
DmitryBespalov, line 47: can you please add the default values for the bundle, version, and buildNumber to be taken from the App configuration?

Multisig/Data/Services/Notification Service/UnregisterNotificationTokenRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Multisig
+//
+//  Created by Moaaz on 8/5/20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct UnregisterNotificationTokenRequest: JSONRequest {
+    let deviceID: UUID
```
DmitryBespalov, line 12: this rather should be string, or we need to test that the UUID formatting is the same as the backend expects it

DmitryBespalov: ah, no, let's see, no change needed

Multisig/Data/Services/Notification Service/UnregisterNotificationTokenRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+    var urlPath: String { return "/api/v1/notifications/devices/\(deviceID)/safes/\(address)/" }
+
+    typealias ResponseType = Response
+
+    init(deviceID: UUID, address: Address) {
+        self.address = address.checksummed
+        self.deviceID = deviceID
+    }
+
+    struct Response: Decodable {
```
DmitryBespalov: this can be renamed to "EmptyResponse", and in the body added comment "// empty"

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
DmitryBespalov: ok, no changes needed, saw your other file

MouazAlzahabi: Is it right to do so in "Request" object?, I mean I that this should be done in upper level

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -0,0 +1,50 @@
...
+        self.buildNumber = buildNumber
+    }
+
+    struct Response: Decodable {
+        let uuid: UUID
+        let safes: [String]
+        let cloudMessagingToken: String
+        let bundle: String
+        let version: String
+        let deviceType: String = "IOS"
```
DmitryBespalov: the response should not have the default value

DmitryBespalov (approved):

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -165,6 +178,11 @@ class RemoteNotificationHandler {
         DispatchQueue.main.async {
             log("Saving authorization status")
             self.authorizationStatus = status
+
+            // At the time when permission granted the token will be already set so we need to register all stored safes
+            if self.authorizationStatus != nil {
```
DmitryBespalov: I  think this is not needed here, because the registerAll will be called upon token refresh anyway

MouazAlzahabi: The thing is I tried it, when permission granted the token won't be refreshed,

DmitryBespalov: having this in  the "setStatus" is the wrong place  to have it. It is better  to move it up above the line 172 -  when the authorization granted, then trigger registering  all.

Anyway, let's merge and I'll test and fix it

DmitryBespalov (approved):

PR #448, opened 2020-08-07
Gh 445/open tx

DmitryBespalov:
- Added opening of the transaction list screen
- Added opening of the details screen when we have transaction hash
- Created placeholders for registering/unregistering

### DIFF
--- a/Multisig/Logic/Models/Safe.swift
+++ b/Multisig/Logic/Models/Safe.swift
@@ -78,14 +78,18 @@ extension Safe: Identifiable {
     static func edit(address: String, name: String) {
         dispatchPrecondition(condition: .onQueue(.main))
         let context = App.shared.coreDataStack.viewContext
-
         let fr = Safe.fetchRequest().by(address: address)
-        
         guard let safe = try? context.fetch(fr).first else { return }
-        
         safe.name = name
-
         App.shared.coreDataStack.saveContext()
+    }
+
+    static func select(address: String) {
+        dispatchPrecondition(condition: .onQueue(.main))
+        let context = App.shared.coreDataStack.viewContext
+        let fr = Safe.fetchRequest().by(address: address)
+        guard let safe = try? context.fetch(fr).first else { return }
+        safe.select()
     }
     
     static func remove(safe: Safe) {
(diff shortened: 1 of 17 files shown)


Multisig/Logic/Models/Safe.swift
```diff
@@ -78,15 +78,19 @@ extension Safe: Identifiable {
...
         let fr = Safe.fetchRequest().by(address: address)
-        
         guard let safe = try? context.fetch(fr).first else { return }
-        
         safe.name = name
-
         App.shared.coreDataStack.saveContext()
     }
+
+    static func select(address: String) {
```
MouazAlzahabi, line 87: Do we need this?, we used to have Selection.select right?

DmitryBespalov: no, there's no Selection.select(address), so I've added this static method when I don't have the Safe object

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -156,6 +180,27 @@ class RemoteNotificationHandler {
         }
     }
 
+    // MARK: - Registering in the service
```
MouazAlzahabi, line 178: In my PR I add the implementation for all those functions, after the merge you can remove them

Multisig/UI/App/ViewState.swift
```diff
@@ -20,6 +20,9 @@ class ViewState: ObservableObject {
         state = to
     }
 
+    @Published
+    var presentedSafeTxHash: String?
```
MouazAlzahabi, line 24: All good, but I think we need to wrap it in some structure

Multisig/UI/App/RemoteNotificationHandler.swift
DmitryBespalov: ok

MouazAlzahabi (approved):

Multisig/UI/App/ViewState.swift
DmitryBespalov: we'll do it in another pass over the whole notifications code next week

Issue #434, opened 2020-08-03
Displaying different notification types and opening the details in the app


DmitryBespalov:
- [x] User should be asked to enable push notifications after successfully adding their first Safe
- [x] Create new 'remote notification' app extension that will process the push payload and display the localized notification message.
- [x] Parse the notification payload
- [ ] ~Implement checking safe name and checking that safe exists before displaying the notification~
  - Not feasible at the moment: all notifications will be shown, but when clicked, if safe was removed then it might just have no effect (nothing happens)
- [x] Implement formatting the ETH amount for displaying the notification
- [x] Implement switching to the safe and then  opening the transaction details or transaction list

![NotificationsDiagram.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/3d644af5-28b6-49e3-8879-11826f57024a)

# Known issues
- The safe name will not be shown, instead, the abbreviated safe address will be shown
- It is possible that device will receive notifications even if the safe is removed, albeit it should be a rare case

liliya-soroka:
Verified
app version 2.0.0(12)

PR #454, opened 2020-08-11
gh-434  updated

DmitryBespalov:
- Introduced a queue to make sure we don't have racing registration/unregistration requests
- Renamed user defaults key
- Reordered file layout

### DIFF
--- a/Multisig/UI/App/RemoteNotificationHandler.swift
+++ b/Multisig/UI/App/RemoteNotificationHandler.swift
@@ -19,7 +19,7 @@ class RemoteNotificationHandler {
     @EnumDefault(key: "io.gnosis.multisig.authorizationStatus")
     private var authorizationStatus: UNAuthorizationStatus?
 
-    @UserDefault(key: "io.gnosis.multiSig.pushToken")
+    @UserDefault(key: "io.gnosis.multisig.pushToken")
     private var token: String?
 
     private var queue = DispatchQueue(label: "RemoteNotificationHandlerQueue")
--- a/Multisig/UI/App/RemoteNotificationHandler.swift
+++ b/Multisig/UI/App/RemoteNotificationHandler.swift
@@ -11,44 +11,31 @@ import UIKit
 import UserNotifications
 import Firebase
 
-extension UNAuthorizationStatus {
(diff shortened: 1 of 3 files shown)


sche (approved):

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -11,44 +11,31 @@ import UIKit
...
-        case .authorized: return "Authorized (granted)"
-        case .denied: return "Denied"
-        case .notDetermined: return "Not Determined"
-        case .provisional: return "Provisional (granted)"
-        @unknown default: return "Unknown: \(rawValue)"
-        }
-    }
-}
+    @EnumDefault(key: "io.gnosis.multisig.authorizationStatus")
+    private var authorizationStatus: UNAuthorizationStatus?
```
MouazAlzahabi, line 20: Why do we need to store authorizationStatus in user defaults?

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -11,44 +11,31 @@ import UIKit
...
+
+    @UserDefault(key: "io.gnosis.multiSig.pushToken")
```
MouazAlzahabi: Same here, I think storing the token in memory is enough right?

Multisig/UI/App/RemoteNotificationHandler.swift
DmitryBespalov: this is to monitor if the status have changed, and if yes and it was granted, to register the safes again

Multisig/UI/App/RemoteNotificationHandler.swift
DmitryBespalov: I would store in the storage because user can close the app and reopen, or something might happen, I thought it would be better to store it.

MouazAlzahabi: In that case we will receive it again on app start right ?

DmitryBespalov: I can't count on Firebase 😄

Issue #453, opened 2020-08-11
Check out the client gateway


DmitryBespalov:
See how to integrate it and what is the  impact of that change.

If there are mismatches, provide documentation or recommendations on how to change the response

# How to test
No visible testable  behavior, only integration tests implemented (in Xcode)

PR #465, opened 2020-08-13
gh-453  added gateway and integration tests

DmitryBespalov:
- Added new service
- Added models for that service
- Added integration tests

I'd like to merge  it but  keep it not used yet


sche (approved):

sche:
🥇

Multisig/Data/Services/Utils/JSONHTTPClient.swift
```diff
@@ -13,6 +13,8 @@ public protocol JSONRequest: Encodable {
     var urlPath: String { get }
     /// Query parameters
     var query: String? { get }
+    /// URL that overrides the urlPath and query
+    var url: URL? { get }
```
MouazAlzahabi, line 17: I also added the url attribute in a past sprint before removing it again, I think in this case we can use this approch instead of PagedRequest.

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -0,0 +1,380 @@
...
+
+protocol SCGDetailedExecutionInfo: Decodable {}
+
+enum SCGDetailedExecutionInfoType: String, Decodable {
+    case MODULE
+    case MULTISIG
+}
+
+enum SCGDetailedExecutionInfoWrapper: Decodable {
+    case Module(SCGModuleExecutionDetails)
```
MouazAlzahabi: Why upper case ?

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -0,0 +1,380 @@
...
+
+import Foundation
+
+struct SCGPage<T: Decodable>: Decodable {
+    let next: String?
+    let previous: String?
+    let results: [T]
+}
+
+struct SCGTransactionSummary: Decodable {
```
MouazAlzahabi: As ClientGateway will be our default backend, I would remove the prefix ClientGateway& SCG

MouazAlzahabi (approved):

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
DmitryBespalov: This is how it is in the safe client gateway, I went on with 1-to-1 case names wherever they are. Make it Swifty?

Multisig/Data/Services/Utils/JSONHTTPClient.swift
DmitryBespalov: yes, that would work

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
DmitryBespalov: I will keep it for  the types  that  collide with existing  types

PR #490, opened 2020-08-24
gh-398  added domain and code to non fatal errors

DmitryBespalov:
I think that the non-fatals were not logged because the NSError was missing proper error domain and error code.

Documentation: https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=ios#log-excepts


MouazAlzahabi (approved):

sche (approved):
👍

DmitryBespalov:
will be merged as part of other pr

PR #528, opened 2020-08-31
Gh 489 prototyping with tabbar

sche:


Multisig/UI/App/MainTabView.swift
```diff
@@ -9,58 +9,55 @@
...
-            SettingsView()
-                .gnoTabItem(id: ViewStateMode.settings) {
+            MainContentView(SettingsView())
+                .tabItem {
                     VStack {
                         Image("tab-icon-settings")
                         Text("Settings")
                     }
                 }
+                .tag(2)
```
MouazAlzahabi: Can we avoid using constants?we can use ViewStateMode underlying value for instance

Multisig/UI/UI Library/HideSystemNavigationBarModifier.swift
```diff
@@ -8,6 +8,7 @@
 
 import SwiftUI
 
+#warning("TODO: do we need this modifier?")
```
MouazAlzahabi: I also don't think we will use it

MouazAlzahabi (approved):

Multisig/UI/App/MainTabView.swift
sche: ok

Multisig/UI/UI Library/HideSystemNavigationBarModifier.swift
sche: removed

PR #537, opened 2020-09-09
Gh 507 import seed screen

sche:


Multisig/Logic/Models/PrivateKey/web3swift/Base58.swift
```diff
@@ -115,7 +115,6 @@ struct Base58 {
 }
```
MouazAlzahabi, line 117: Does it show two lines here?

sche: there are actually two lines. Will remove in the next pr

MouazAlzahabi (approved):

PR #558, opened 2020-09-16
Gh 510 signing transaction

sche:
Display "Confirm" button when needed.

### DIFF
--- a/Multisig/Logic/Models/Address.swift
+++ b/Multisig/Logic/Models/Address.swift
@@ -109,6 +109,19 @@ struct Address: Hashable, ExpressibleByStringInterpolation, CustomStringConverti
     }
 }
 
+extension Address: Codable {
+    init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let string = try container.decode(String.self)
+        try self.init(from: string)
+    }
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(_store.hex(eip55: true))
+    }
+}
+
 extension EthereumAddress.Error: LocalizedError {
     public var errorDescription: String? {
         switch self {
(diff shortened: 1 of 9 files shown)


Multisig/Logic/Models/Address.swift
```diff
@@ -109,6 +109,19 @@ struct Address: Hashable, ExpressibleByStringInterpolation, CustomStringConverti
     }
 }
 
+extension Address: Codable {
```
DmitryBespalov, line 112: we already have this implemented in the AddressString, could you use that one?

Multisig/Logic/Models/Safe+CoreDataProperties.swift
```diff
@@ -18,6 +18,7 @@ extension Safe {
     @NSManaged public var additionDate: Date?
     @NSManaged public var address: String?
     @NSManaged public var name: String?
+    @NSManaged public var ownersData: Data?
```
DmitryBespalov, line 21: is there need to do something extra for the lightweight migration 🤔? versioning the xcdatamodel, or something?

Multisig/UI/Settings/Safe Settings/SafeSettingsViewModel.swift
```diff
@@ -46,16 +46,21 @@ class SafeSettingsViewModel: BasicLoadableViewModel {
 }
 
 extension Safe {
-
     func update(from safeInfo: SafeStatusRequest.Response) {
+        // check if we need to update owners
+        // always updating owners will cause an infinite loop
+        let newOwners = safeInfo.owners.map { $0.address }
+        if owners != newOwners {
```
DmitryBespalov, line 53: Should this new block of code be moved to the end of this method, so that on the update of the object the other values are set?

sche:
I will re-work this PR.

Issue #570, opened 2020-09-28
Disable signing for the 2.2.0 release


DmitryBespalov:
Disable the signing so it is not visible to the user

liliya-soroka:
Verified
app version 2.2.0 -152

PR #571, opened 2020-09-28
gh-570  Disabled signing

DmitryBespalov:
Disabled both confirm button and settings cell

### DIFF
--- a/Multisig/UI/Settings/App Settings/BasicAppSettingsView.swift
+++ b/Multisig/UI/Settings/App Settings/BasicAppSettingsView.swift
@@ -26,28 +26,8 @@ struct BasicAppSettingsView: View {
 
     var body: some View {
         List {
-            if signingKeyAddress != nil {
-                ForEach(0..<1) { _ in
-                    VStack(alignment: .leading, spacing: 4) {
-                        Text("Signing key").bold()
-                        AddressView(self.signingKeyAddress!)
-                    }
-                    .padding()
-                }
-                .onDelete { _ in
-                    do {
-                        try App.shared.keychainService.removeData(forKey: KeychainKey.ownerPrivateKey.rawValue)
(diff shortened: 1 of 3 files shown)


Multisig/UI/Settings/App Settings/BasicAppSettingsView.swift
```diff
@@ -82,6 +62,33 @@ struct BasicAppSettingsView: View {
             self.trackEvent(.settingsApp)
         }
     }
+
+    @ViewBuilder
+    var signingWalletView: some View {
+        if signingKeyAddress != nil {
+            ForEach(0..<1) { _ in
```
DmitryBespalov, line 69: I'm not going to refactor  it now, but was curious why there's ForEach to produce just 1 element?

Multisig/UI/Settings/App Settings/BasicAppSettingsView.swift
```diff
@@ -82,6 +62,33 @@ struct BasicAppSettingsView: View {
...
+                VStack(alignment: .leading, spacing: 4) {
+                    Text("Signing key").bold()
+                    AddressView(self.signingKeyAddress!)
+                }
+                .padding()
+            }
+            .onDelete { _ in
+                do {
```
DmitryBespalov, line 77: This should be in the view model / model object of some sort

Multisig/UI/Settings/App Settings/BasicAppSettingsView.swift
sche: to enable swipe to delete

DmitryBespalov: wow!

sche (approved):

PR #596, opened 2020-09-30
Gh 579/fix loading

DmitryBespalov:
- Added weak selfs everywhere in closures that capture them using Combine
- Changed loadable to load only once when the view is shown

That changed the app behavior (communicated to Liliya already):
- Switching tabs does not reload the content
- Going in and out of transaction details keeps the transaction list not reloaded and scrolled to the cell that was selected.
- Somehow transaction list cell selection is not removed after coming back from the tx details
- One artifact is that transaction list is reloading whenever the Switch Safe opened (even if not switching the safe), but I'm ok with that

The content is reloaded on the app re-opening, as it was before.

Please checkout the branch and ran it on your device to check that crashes are gone (I could not reproduce them after my fixes).

### DIFF
--- a/Multisig/UI/ENS/EnterENSNameViewModel.swift
+++ b/Multisig/UI/ENS/EnterENSNameViewModel.swift
@@ -32,14 +32,16 @@ class EnterENSNameViewModel: ObservableObject {
         subscribers.forEach { $0.cancel() }
 
         $text
-            .map { v -> String in
+            .map { [weak self] v -> String in
+                guard let `self` = self else { return "" }
                 self.reset()
                 return v.trimmingCharacters(in: .whitespacesAndNewlines)
             }
             .filter { !$0.isEmpty }
             .debounce(for: 0.5, scheduler: RunLoop.main)
             .removeDuplicates()
-            .tryMap { v -> String in
+            .tryMap { [weak self] v -> String in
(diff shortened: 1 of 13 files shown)


sche (approved):
👍

Multisig/UI/ENS/EnterENSNameViewModel.swift
```diff
@@ -32,14 +32,16 @@ class EnterENSNameViewModel: ObservableObject {
...
+                guard let `self` = self else { return "" }
                 self.reset()
                 return v.trimmingCharacters(in: .whitespacesAndNewlines)
             }
             .filter { !$0.isEmpty }
             .debounce(for: 0.5, scheduler: RunLoop.main)
             .removeDuplicates()
-            .tryMap { v -> String in
+            .tryMap { [weak self] v -> String in
+                guard let `self` = self else { return "" }
```
MouazAlzahabi, line 44: You can replace 'self' with "_" as it's not used right?

MouazAlzahabi (approved):

Multisig/UI/ENS/EnterENSNameViewModel.swift
DmitryBespalov: this will be changed in the other pr

Issue #566, opened 2020-09-23
[iOS] Enter ENS Name does not resolve names to addresses


DmitryBespalov:
**Bug description**
In the latest app (release version), when adding  a safe using ENS name `alice.eth`, I get error "No resolver set for the record"

**Steps To Reproduce**
- Add safe
- Enter address using ENS name "alice.eth"

**Expected Result**
Safe with address  `0xcd2E72aEBe2A203b84f46DEEC948E6465dB51c75` is added

**Device & App version**
 - Device/OS: iPhone 6s, iOS 14
 - App version: 2.1.0 (3)
 
**Environment**
Production Mainnet

**Additional info**
(Optional)

DmitryBespalov:
Fixed in 2.2.0 (3)

liliya-soroka:
The issue still exists:
2.2.0 (5) 
prod mainnet
os 13.7
There are some specific steps ( searching for them)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/782d626c-7f6e-4b4a-8961-262e6161716a)

@DmitryBespalov and @tschubotz , there is some situation when the issue still exists, but the priority can be changed to lower as it's not easy to get the error. I am searching for the steps

liliya-soroka:
The issue was not reproduced in 2.3.0 -161

liliya-soroka:
Reopened: 
The issue was reproduced:
Steps:
1. turn internet connectin off
2. call add safe
3. select enter ENS
4. enter ens name 
5. turn the internet connection on
6. reenter the last letter to activate ens name

Current result:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/bdd6c1e0-5f80-4fd3-87e9-4ddd6b4800cf)
Prod, mainnet
app version: 2.5.0(186)

sche:
It should work

liliya-soroka:
Verified
app version 2.8.0 (266)

PR #597, opened 2020-09-30
gh-566  adjusted publishers

DmitryBespalov:
The issue was that we had pipeline that was triggered by `$text` and by the `resolve(...)` function. I rewrote this pipeline to be only single pipeline and just use changes from the `$text` for loading ENS

### DIFF
--- a/Multisig/UI/ENS/EnterENSNameViewModel.swift
+++ b/Multisig/UI/ENS/EnterENSNameViewModel.swift
@@ -28,34 +28,44 @@ class EnterENSNameViewModel: ObservableObject {
 
     private var subscribers = Set<AnyCancellable>()
 
-    func resolve(name: String) {
-        subscribers.forEach { $0.cancel() }
-
+    init() {
         $text
-            .map { v -> String in
-                self.reset()
+            .map { [weak self] v -> String in
+                self?.reset()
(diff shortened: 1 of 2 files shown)


sche (approved):

Multisig/UI/ENS/EnterENSNameViewModel.swift
```diff
@@ -28,34 +28,44 @@ class EnterENSNameViewModel: ObservableObject {
...
 
-    func resolve(name: String) {
-        subscribers.forEach { $0.cancel() }
-
+    init() {
         $text
-            .map { v -> String in
-                self.reset()
+            .map { [weak self] v -> String in
+                self?.reset()
```
MouazAlzahabi: In the other PR you used "guard let `self` = self else { return }" validation, I think it's safer if we have same validation here

MouazAlzahabi (approved):

Multisig/UI/ENS/EnterENSNameViewModel.swift
DmitryBespalov: generally, if there would be more than 1 statement that uses `self`, that would be a valid concern, but for a single statement it doesn't matter. 

This is not future-proof though, as adding more calls to self will have to use the optional chaining. I'll update it.

PR #600, opened 2020-09-30
Gh 592 transaction refactoring

sche:
transaction view model refactoring and signing


Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
```diff
@@ -0,0 +1,80 @@
...
+        refundReceiver = transaction.refundReceiver.description
+        nonce = transaction.nonce.description
+        contractTransactionHash = transaction.safeTxHash.description
+
+        let hashToSign = Data(ethHex: contractTransactionHash)
+
+        guard let pkData = try? App.shared.keychainService.data(forKey: KeychainKey.ownerPrivateKey.rawValue),
+              let pk = try? EthereumPrivateKey(pkData.bytes),
+              let sig = try? pk.sign(hash: hashToSign.bytes) else {
+            sender = ""
```
DmitryBespalov, line 57: I see that the errors are not handled here, why is that? I.e. if key is invalid or missing, we'll send the request that will be failing but we probably won't know the actual reason

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
```diff
@@ -0,0 +1,80 @@
...
+        baseGas = transaction.baseGas.description
+        gasPrice = transaction.gasPrice.description
+        gasToken = transaction.gasToken?.description
+        refundReceiver = transaction.refundReceiver.description
+        nonce = transaction.nonce.description
+        contractTransactionHash = transaction.safeTxHash.description
+
+        let hashToSign = Data(ethHex: contractTransactionHash)
+
+        guard let pkData = try? App.shared.keychainService.data(forKey: KeychainKey.ownerPrivateKey.rawValue),
```
DmitryBespalov, line 54: please use the `privateKey` in full name, and `signature`, too

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
```diff
@@ -0,0 +1,80 @@
...
+              let pk = try? EthereumPrivateKey(pkData.bytes),
+              let sig = try? pk.sign(hash: hashToSign.bytes) else {
+            sender = ""
+            signature = ""
+            return
+        }
+        sender = pk.address.hex(eip55: true)
+
+        let v = String(format:"%02X", sig.v + 27)
```
DmitryBespalov: Not sure if understand it from the first reading. Is this a conversion from int to hex? Is there a way to convert the sig obj directly to data and then to hex?

Multisig/Logic/Models/Transaction.swift
```diff
@@ -7,174 +7,38 @@
...
+    let data: DataString?
+    let operation: Operation
+    let safeTxGas: UInt256String
+    let baseGas: UInt256String
+    let gasPrice: UInt256String
+    let gasToken: AddressString?
+    let refundReceiver: AddressString // zero address if no refund receiver is set
+    let nonce: UInt256String
+    // computed based on other properties
+    let safeTxHash: HashString // can be computed from the
```
DmitryBespalov: I'd prefer to have comments on a separate line instead of end of line for better editing and readability.

Also, this comment is abruptly ending

Multisig/Logic/Models/Transaction.swift
```diff
@@ -7,174 +7,38 @@
...
+}
+
+#warning("TODO: make a validatione of safeTxHash here")
```
DmitryBespalov: typo: validatione -> validation

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsViewModel.swift
```diff
@@ -86,4 +86,26 @@ class TransactionDetailsViewModel: BasicLoadableViewModel {
             return try App.shared.clientGatewayService.transactionDetails(id: id!)
         }
     }
+
+    func sign(safeAddress: Address) {
+        guard let transferTx = transactionDetails as? TransferTransactionViewModel,
+              let transaction = transferTx.transaction else { return }
```
DmitryBespalov, line 92: would it be programmer error if  `sign(..)` is called with wrong transactionDetails or the transaction is nil?

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsViewModel.swift
```diff
@@ -86,4 +86,26 @@ class TransactionDetailsViewModel: BasicLoadableViewModel {
...
+        Just(safeAddress)
+            .receive(on: DispatchQueue.global())
+            .tryMap { _ in
```
DmitryBespalov: I think it's better here to use `address` and use it in the call below, so that we don't need to capture anything from the outer method (dk why I say this but it feels better to me)

MultisigIntegrationTests/SafeTransactionServiceIntegrationTests.swift
```diff
@@ -154,8 +154,8 @@ class SafeTransactionServiceIntegrationTests: XCTestCase {
         }
     }
 
-    func fetchTransaction(hash: String) -> Result<Transaction, Error> {
-        var result: Result<Transaction, Error>?
+    func fetchTransaction(hash: String) -> Result<CGSTransaction, Error> {
+        var result: Result<CGSTransaction, Error>?
```
DmitryBespalov: Is it a typo? `CGSTransaction`? (it was `SCGTransaction`)

DmitryBespalov (changes requested):
wow! 👏 let's do some polishing

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: I think this should be a separate task to properly manage missing private key (like when the app is recovered from the iCloud)
wdyt?

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: ok

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
DmitryBespalov: I  think this is not needed at the moment, especially that there was decision not to do iCloud syncing. This code block just skips errors. While  this is ok during  prototyping, this is not ready for production, in my opinion. Throwing and  showing human-readable message to the user would be better.

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: signature will be in conflict with struct var, so I will leave sig here

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: ok, I will rewrite it

Multisig/Logic/Models/Transaction.swift
sche: ok

Multisig/Logic/Models/Transaction.swift
sche: ok

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsViewModel.swift
sche: It would. Would you prefer to crash in guard statement?

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsViewModel.swift
sche: ok

MultisigIntegrationTests/SafeTransactionServiceIntegrationTests.swift
sche: yes, thank you

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
DmitryBespalov: To resolve the conflict, the code could be `self.signature`

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsViewModel.swift
DmitryBespalov: I would crash only in development builds, i.e. I would add `assertionFailure(...)` - not `preconditionFailure(...)`

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: I mean when the user changes the phone. We could have an address stored, but not the PK, that could lead to a confusion.
I see signing might fail only in case of this misconfiguration

Multisig/UI/Transactions/TransactionDetails/TransactionDetailsViewModel.swift
sche: ok. I will place here `assertionFailure("We have a technical problem. You need to restart the app.")`

sche: in the next pr

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: ok. will change in the next pr

DmitryBespalov (approved):

PR #613, opened 2020-10-09
Gh590/confirming transaction in tx details (ui)

sche:
This PR also contains release 2.2.0 changes

There is an issue with our backend: https://github.com/gnosis/safe-client-gateway/issues/161 
We could add a small delay as a short-term solution.


Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
```diff
@@ -54,18 +49,14 @@ struct SignTransactionRequest: JSONRequest {
...
         }
-
-        guard let pkData = try? App.shared.keychainService.data(forKey: KeychainKey.ownerPrivateKey.rawValue),
-              let privateKey = try? EthereumPrivateKey(pkData.bytes),
-              let signature = try? privateKey.sign(hash: hashToSign.bytes) else {
-            sender = ""
-            self.signature = ""
-            return
+        guard let pkData = try App.shared.keychainService.data(forKey: KeychainKey.ownerPrivateKey.rawValue) else {
+            throw "Private key now found"
```
DmitryBespalov: typo: "now" -> not

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
```diff
@@ -10,6 +10,7 @@ import Foundation
 import Combine
 
 class LoadingTransactionDetailsViewModel: ObservableObject {
+    let safe = Selection.current().safe!
```
DmitryBespalov: this should not be used as the initialized property because the details view model might be reused without destroying, but the selected safe will be changed. Getting selected safe should be happening during the sign() function execution if it is needed there

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
```diff
@@ -37,7 +38,6 @@ class LoadingTransactionDetailsViewModel: ObservableObject {
     }
 
     func reload(transaction: TransactionViewModel) {
-        guard status != .loading else { return }
```
DmitryBespalov: In the end, we can get this back, because this condition will be holding

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
```diff
@@ -73,4 +73,31 @@ class LoadingTransactionDetailsViewModel: ObservableObject {
 
         }
     }
+
+    func sign() {
+        guard let transferTx = transactionDetails as? TransferTransactionViewModel,
+              let transaction = transferTx.transaction else {
+            preconditionFailure("We have a technical problem. You need to restart the app.")
```
DmitryBespalov: This will crash the app hard. The message should be "Failed to sign: either transaction is not a transfer or the internal transaction does not exist."

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
```diff
@@ -73,4 +73,31 @@ class LoadingTransactionDetailsViewModel: ObservableObject {
...
         }
     }
+
+    func sign() {
+        guard let transferTx = transactionDetails as? TransferTransactionViewModel,
+              let transaction = transferTx.transaction else {
+            preconditionFailure("We have a technical problem. You need to restart the app.")            
+        }
+        status = .loading
+        Just(safe.address!)
```
DmitryBespalov, line 85: Do we need a safe address? Can we take the safe address not from the CoreData but from the transaction itself? That would make this signing functionality more flexible and still not so complicated.

DmitryBespalov (changes requested):
Good! Let's get a few more improvements and it'll be ready to merge

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
```diff
@@ -60,51 +60,43 @@ class TransactionDetailsViewModel: BasicLoadableViewModel {
                 .sink(receiveCompletion: { [weak self] completion in
```
DmitryBespalov, line 60: per Richard's and Jose's answers, let's add the `delay(3 seconds)` before receiving the message on RunLoop.main so that we'll request the next message after the delay

Multisig/Data/Services/Safe Transaction Service/SignTransactionRequest.swift
sche: ok

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
sche: ok

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
sche: ok

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
sche: ok

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
sche: I moved getting safe here from class property to a variable. TransactionViewModel does not know about safe

Multisig/UI/Transactions/TransactionDetails/LoadingTransactionDetailsViewModel.swift
sche: ok

DmitryBespalov (approved):

PR #638, opened 2020-10-14
Gh 587/import owner key

MouazAlzahabi:


Multisig/Logic/Models/PrivateKey/web3swift/BIP39.swift
```diff
@@ -147,11 +147,16 @@ public class BIP39 {
 
     static public func seedFromMmemonics(_ mnemonics: String, password: String = "",
                                          language: BIP39Language = BIP39Language.english) -> Data? {
-        let valid = BIP39.mnemonicsToEntropy(mnemonics, language: language) != nil
+        let formattedMnemonics = mnemonics
+            .lowercased()
+            .replacingOccurrences(of: ",", with: " ")
```
DmitryBespalov: I'd rather use CharacterSet - there should be punctuation set, and whitespace and new lines set

DmitryBespalov: also, this is a library function code, if possible, I would not modify it at all

DmitryBespalov (approved):

sche (approved):

PR #685, opened 2020-11-03
Gh 665/collectibles

sche:
Screenshot 2020-11-03 at 12 45 47

Screenshot 2020-11-03 at 12 48 16


sche:
I also changed the endpoint for collectibles in the SwiftUI app

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    @IBOutlet weak var cellWrapperView: UIView!
+    @IBOutlet weak var cellImageView: UIImageView!
+    @IBOutlet weak var cellNameLabel: UILabel!
+    @IBOutlet weak var cellDescriptionLabel: UILabel!
+
+    let cornerRadius: CGFloat = 8
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        backgroundColor = .gnoSystemWhite
```
DmitryBespalov: The styling code that can be done in XIB, should be done in XIB to reduce the code size

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+
+        cellWrapperView.backgroundColor = .gnoSystemWhite
```
DmitryBespalov: Potential bad UI graphical performance because of use of layers. Instead of this, I think, it is better to use Image view with the 'shadow' background - we have such image in the assets folder.

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Multisig
+//
+//  Created by Andrey Scherbovich on 02.11.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class CollectibleTableViewCell: UITableViewCell {
+    @IBOutlet weak var cellWrapperView: UIView!
```
DmitryBespalov: I would rename this variable to cellShadowView

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+        cellWrapperView.backgroundColor = .gnoSystemWhite
+        cellWrapperView.layer.cornerRadius = cornerRadius
+        cellWrapperView.layer.shadowColor = UIColor.gnoShadow.cgColor
+        cellWrapperView.layer.shadowOpacity = 0.6
+        cellWrapperView.layer.shadowOffset = .init(width: 1, height: 2)
+        cellWrapperView.layer.shadowRadius = 6
+
+        cellNameLabel.setStyle(.headline)
+        cellDescriptionLabel.setStyle(.body)
+        cellDescriptionLabel.numberOfLines = 3
```
DmitryBespalov: I would set this parameter in XIB directly

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    func configure(collectible: CollectibleViewModel) {
```
DmitryBespalov: The problem here is that view depends on this model. I would like to have 3 different methods that provide API to this class's users and that therefore will make this view potentially reusable.

```
setName(_ value: String)
setDescription(_ value: String)
setImage(with URL: URL?, placeholder: UIImage)
```

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesHeaderView.swift
```diff
@@ -0,0 +1,31 @@
...
+import UIKit
+
+class CollectiblesHeaderView: UITableViewHeaderFooterView {
+    @IBOutlet weak var headerImageView: UIImageView!
+    @IBOutlet weak var headerNameLabel: UILabel!
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        backgroundView = UIView(frame: self.bounds)
+        backgroundView!.backgroundColor = .gnoSystemWhite
```
DmitryBespalov: I believe this could be done in the XIB? This would reduce this boilerplate code

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesHeaderView.swift
```diff
@@ -0,0 +1,31 @@
...
+        headerNameLabel.setStyle(.headline)
+    }
+
+    func configure(collectibleSection: CollectibleListSection) {
```
DmitryBespalov: Here, too, the view depends on the model, and hence, is less reusable than it could.

I'd propose to have 2 methods:

```
setName(_ value: String)
setImage(with URL: URL, placeholder: UIImage)
```

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Multisig
+//
+//  Created by Andrey Scherbovich on 02.11.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class CollectiblesViewController: LoadableViewController, UITableViewDelegate, UITableViewDataSource {
+    let clientGatewayService = App.shared.clientGatewayService
```
DmitryBespalov: I think, this should be a variable and not a constant so that this dependency is potentially changea-ble

Multisig/Data/Services/Safe Client Gateway Service/CollectiblesRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct CollectiblesRequest: JSONRequest {
+    let address: String
+
+    var httpMethod: String { return "GET" }
+    var urlPath: String { return "/v1/safes/\(address)/collectibles/" }
```
DmitryBespalov: This code does not use modern Swift. Let's remove the "return" statements?

Multisig/Data/Services/Safe Client Gateway Service/CollectiblesRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    typealias Response = [Collectible]
```
DmitryBespalov: In my opinion, this typealias is not needed. The `ResponseType` should be used directly. That would make this code clearer.

Multisig/Data/Services/Safe Client Gateway Service/CollectiblesRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+    typealias ResponseType = Response
+
+    init(address: Address) {
```
DmitryBespalov, line 21: When structs have initializers in their struct definitions, Swift does not generate a default initializer. I would like to have that still. Can we move this initializer to the extension on the `CollectiblesRequest`?

Multisig/UI/App/UIKit/UI Library/LoadableViewController/LoadableViewController.swift
```diff
@@ -28,6 +28,7 @@ class LoadableViewController: UIViewController {
     @IBOutlet var allViews: [UIView]!
     var scrollViews: [UIScrollView] = []
     private var needsReload: Bool = false
+    let notificationCenter = NotificationCenter.default
```
DmitryBespalov: In my opinion, this should be a variable, not a constant, because then this dependency is changeable.

Multisig/UI/App/UIKit/UI Library/LoadableViewController/LoadableViewController.swift
```diff
@@ -37,12 +38,25 @@ class LoadableViewController: UIViewController {
             view.refreshControl = createRefreshControl()
         }
         setNeedsReload()
+
+        notificationCenter.addObserver(
```
DmitryBespalov: Although this behavior of reacting to safe changes is common to most of the loadable view controllers in the app, it is not all - Transaction details will not be dependent on the currently selected safe. 

Therefore, it is good to have this behavior in this parent class, however, I'd introduce a configuration variable that subclasses can override or set that would enable or disable reaction to the safe selection.

I would name this variable `reactsToSelectedSafeChanges`

Multisig/UI/App/UIKit/UI Library/UIView+UINib.swift
```diff
@@ -8,7 +8,7 @@
 
 import UIKit
 
-extension UITableViewCell {
+extension UIView {
```
DmitryBespalov, line 11: I think the UIView is the wrong place for the `reuseID`. I think, this variable should be present only in reusable classes - UITableViewCell, UITableViewHeaderFooterView, and UICollectionViewCell, (and possibly collection view headers). 

Therefore, it seems that the better way to handle reuse of this code is to declare a protocol with default implementation and make those classes adopt this protocol.

Multisig/UI/App/UIKit/UI Library/UIView+UINib.swift
```diff
@@ -32,6 +32,16 @@ extension UITableView {
         register(aClass.nib(), forCellReuseIdentifier: reuseID ?? aClass.reuseID)
     }
 
+    /// Registers a HeaderFooterView using a nib named the same as the view's class and
```
DmitryBespalov: This UITableView extension has become big enough to put it in a separate file. Let's extract?

Multisig/UI/Assets/Collectibles/CollectibleBalancesModel.swift
```diff
@@ -63,15 +63,3 @@ extension CollectibleListSection {
         }.sorted { $0.name < $1.name }
     }
 }
-
-extension Safe {
-    static func selectedSafeAddress() -> Address? {
```
DmitryBespalov, line 68: removing unused code is good! 👏

DmitryBespalov (changes requested):
Before merging I think we need to resolve some detailed software design issues

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesHeaderView.swift
DmitryBespalov: Also, I get this warning in the console: `Changing the background color of UITableViewHeaderFooterView is not supported. Use the background view configuration instead`

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueCell(CollectibleTableViewCell.self, for: indexPath)
+        let collectible = sections[indexPath.section].collectibles[indexPath.row]
+        cell.configure(collectible: collectible)
+        return cell
+    }
+
+    // MARK: - Table view delegate
+
```
DmitryBespalov, line 102: Currently, if I select a cell, it stays selected. That looks wrong. Can we deselect the cell if we tap it?

DmitryBespalov: Or even disable selection

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
sche: ok

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
sche: ok

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesHeaderView.swift
sche: done

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesHeaderView.swift
sche: done

Multisig/Data/Services/Safe Client Gateway Service/CollectiblesRequest.swift
sche: ok

Multisig/Data/Services/Safe Client Gateway Service/CollectiblesRequest.swift
sche: ok

Multisig/UI/App/UIKit/CollectiblesViewController/CollectibleTableViewCell.swift
sche: done

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesViewController.swift
sche: ok

Multisig/UI/App/UIKit/UI Library/LoadableViewController/LoadableViewController.swift
sche: ok

Multisig/UI/App/UIKit/UI Library/UIView+UINib.swift
sche: ok

Multisig/UI/App/UIKit/CollectiblesViewController/CollectiblesViewController.swift
sche: I will do in the next PR

DmitryBespalov (approved):

PR #694, opened 2020-11-05
Gh 692 safe settings to uikit part1

sche:


Multisig/UI/App/SceneDelegate.swift
```diff
@@ -56,11 +65,27 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+            SegmentBarItem(image: #imageLiteral(resourceName: "ico-app-settings"), title: "App Settings")
+        ]
+        segmentVC.viewControllers = [
+            SafeSettingsViewController(),
+            SafeSettingsViewController()
+        ]
+        segmentVC.selectedIndex = 0
+
+        let noSafesVC = NoSafesViewController()
+        noSafesVC.hasSafeViewController = segmentVC
```
DmitryBespalov, line 84: noSafesVC should be inside the segmentVC, because otherwise the app settings would not show when safe is not loaded

DmitryBespalov (approved):
Good for first iteration!

PR #710, opened 2020-11-11
Gh 700 app settings

sche:


Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+//
+
+import UIKit
+import SwiftUI
+
+fileprivate protocol SectionItem {}
+
+class AppSettingsViewController: UITableViewController {
+    private let app = App.configuration.app
+    private let legal = App.configuration.legal
```
DmitryBespalov, line 16: I think both `app` and `legal` should be `var`s because they are configurable dependencies on the outside of this class

Multisig/UI/App/UIKit/UI Library/UIFont+Styles.swift
```diff
@@ -54,9 +54,13 @@ extension UIFont {
 }
 
 extension UILabel {
-    func setStyle(_ style: GNOTextStyle) {
+    func setStyle(_ style: GNOTextStyle, textColor: UIColor? = nil) {
```
DmitryBespalov, line 57: there's `color(..)` method on the style, no need to modify this method, just build new style

`GNOTextStyle.body.color(.gnoSnowwhite)`

Multisig/UI/App/UIKit/AppSettingsViewController/ImportedKeyCell.swift
```diff
@@ -0,0 +1,26 @@
...
+class ImportedKeyCell: UITableViewCell {
+    @IBOutlet private weak var addressWithTitleView: AddressWithTitleView!
+    var onRemove: (() -> Void)?
+
+    @IBAction func removeKey() {
+        onRemove?()
+    }
+
+    func setAddress(_ value: Address) {
+        addressWithTitleView.setAddress(value)
```
DmitryBespalov, line 20: We need to remember to make the addresses copyable

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+                AppSettings.setSigningKeyAddress(nil)
+//                App.shared.snackbar.show(message: "Owner key removed from this app")
+                Tracker.shared.setUserProperty("0", for: TrackingUserProperty.numKeysImported)
+                self.reload()
+            } catch {
+                LogService.shared.error(error.localizedDescription)
+                App.shared.snackbar.show(message: error.localizedDescription)
+            }
+        }
+        let cancel = UIAlertAction(title: "Cancel", style: .cancel, handler: nil)
```
DmitryBespalov, line 168: Now "Cancel" text is green color, but it needs to be system blue

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+
+    #warning("TODO: show snackbar message")
+    private func removeImportedOwnerKey() {
+        let alertController = UIAlertController(
+            title: nil,
+            message: "Removing the owner key only removes it from this app. It doesn't delete any Safes from this app or from blockchain. For Safes controlled by this owner key, you will no longer be able to sign transactions in this app",
+            preferredStyle: .actionSheet)
+        let remove = UIAlertAction(title: "Remove", style: .destructive) { [unowned self] _ in
+            do {
+                try App.shared.keychainService.removeData(
```
DmitryBespalov, line 157: I think from the design standpoint this code should be moved to the business logic layer - to something called `SigningController` or some other name

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+            let hostingController = UIHostingController(rootView: advancedVC)
+            show(hostingController, sender: self)
+        default:
+            break
+        }
+    }
+
+    override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        let view = tableView.dequeueHeaderFooterView(BasicHeaderView.self)
+        view.setName("")
```
DmitryBespalov, line 204: Is there a default text that needs to be overridden? Maybe we can set the empty string in the header itself?

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+        case Section.General.licenses(let name):
+            return basicCell(name: name, indexPath: indexPath)
+        case Section.General.getInTouch(let name):
+            return basicCell(name: name, indexPath: indexPath)
+        case Section.General.appVersion(let name, let version):
+            return infoCell(name: name, info: version, indexPath: indexPath)
+        case Section.General.network(let name, let network):
+            return infoCell(name: name, info: network, indexPath: indexPath)
+        case Section.Advanced.advanced(let name):
+            return basicCell(name: name, indexPath: indexPath)
```
DmitryBespalov, line 118: this is really unreadable blob of case statements. I think that if there's a blank line before each case statement, it will be better readable

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -0,0 +1,226 @@
...
+        tableView.backgroundColor = tableBackgroundColor
+        tableView.separatorStyle = .none
+        tableView.registerCell(BasicCell.self)
+        tableView.registerCell(ImportedKeyCell.self)
+        tableView.registerCell(InfoCell.self)
+        tableView.registerHeaderFooterView(BasicHeaderView.self)
+
+        buildSections()
+
+        notificationCenter.addObserver(
```
DmitryBespalov, line 56: I think for closing the controller having a closure in the `EnterSeedPhraseView` instead of the notification would be enough for this screen

DmitryBespalov (changes requested):

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
DmitryBespalov: I'll remove in another PR

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
DmitryBespalov: may be minor, but we'll need to communicate that

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
sche: ok

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
sche: ok

Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
sche: there is a warning, I propose to refactor it with fixing the warning; I will add this note to the warning

PR #711, opened 2020-11-12
Gh 701 safe settings

sche:
add more cells to safe settings screen; make address tappable

### DIFF
--- a/Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
+++ b/Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
@@ -12,11 +12,11 @@ import SwiftUI
 fileprivate protocol SectionItem {}
 
 class AppSettingsViewController: UITableViewController {
-    private let app = App.configuration.app
-    private let legal = App.configuration.legal
     private let tableBackgroundColor: UIColor = .gnoWhite
     private let advancedSectionHeaderHeight: CGFloat = 28
     var notificationCenter = NotificationCenter.default
+    var app = App.configuration.app
+    var legal = App.configuration.legal
 
     private typealias SectionItems = (section: Section, items: [SectionItem])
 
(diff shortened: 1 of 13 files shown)


Multisig/UI/App/UIKit/AppSettingsViewController/AppSettingsViewController.swift
```diff
@@ -100,22 +100,31 @@ class AppSettingsViewController: UITableViewController {
         switch item {
         case Section.General.importKey(let name):
             return basicCell(name: name, indexPath: indexPath)
+
```
DmitryBespalov, line 103: thank you! 👏

Multisig/UI/App/UIKit/UI Library/ReusableCells/AddressDetailsCell.swift
```diff
@@ -0,0 +1,44 @@
...
+
+class AddressDetailsCell: UITableViewCell {
+    @IBOutlet private weak var addressWithTitleView: AddressWithTitleView!
+
+    private var style: Style = .nameAndAddress
+    var onViewDetails: (() -> Void)?
+
+    static let rowHeight: CGFloat = 68
+
+    enum Style {
```
DmitryBespalov, line 19: do you think that option when the name is optional and not set, then this would mean to hide it - that would simplify this class? 

I.e. if the name is not set, it will not be shown.

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressWithTitleView.swift
```diff
@@ -12,19 +12,61 @@ class AddressWithTitleView: UINibView {
     @IBOutlet private weak var identiconView: UIImageView!
     @IBOutlet private weak var textLabel: UILabel!
     @IBOutlet private weak var detailLabel: UILabel!
+    @IBOutlet private weak var button: UIButton!
+
+    private var style: Style = .nameAndAddress
+    private var address: Address!
+
+    enum Style {
```
DmitryBespalov, line 20: same question here: maybe by default make this style as address, and only when the name is set to not nil it will be another style?

DmitryBespalov (commented):
I have some questions

Multisig/UI/App/UIKit/UI Library/ReusableCells/AddressDetailsCell.swift
sche: I thought about it. I used style here to make it easier to adjust it there will be a third/fourth style.
But with name = nil it could also be achieved. 

Do you think it worth refactoring at this moment?

DmitryBespalov: No, I think it's not worth it, the scope is small

DmitryBespalov (approved):
let's merge this! 👍

Issue #700, opened 2020-11-09
[UIKit] App Settings


DmitryBespalov:
- Advanced App Settings
- Get In Touch
- Import key / Remove owner key

liliya-soroka:
Reopened:
1. app version doesn't contain full information about the build
current result: app version 2.6.0  . (199) is missed

PR #721, opened 2020-11-19
gh-700 small adjustments

sche:


DmitryBespalov:
can you please check that tapping on addresses in settings is copying the address and showing snackbar?

DmitryBespalov (approved):

PR #730, opened 2020-11-25
Gh 716 push notifications

sche:


Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -13,8 +13,8 @@ import Firebase
 
 class RemoteNotificationHandler {
 
-    @UserDefault(key: "io.gnosis.multisig.deviceID")
-    private var storedDeviceID: String?
+    @UserDefaultWithDefault(key: "io.gnosis.multisig.deviceID", defaultValue: UUID().uuidString)
```
DmitryBespalov: I am concerned about this current code. This new default value must be saved on the app startup, otherwise, it will not be persisted between app starts, and we will have different UUID for the same app / token. I think there's a bug here.

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -67,9 +66,7 @@ class RemoteNotificationHandler {
 
     func safeAdded(address: Address) {
         logDebug("Safe added: \(address)")
-        if authorizationStatus == nil {
```
DmitryBespalov: Can you tell more about why this was removed? Will we have the push notification permission when the safe is added?

Multisig/Logic/Models/Signer.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /// Signs the hash of the provided string with a stored private key.
+    /// Currently the app can store only one private key.
+    /// - Parameters:
+    ///   - string: string to hash and sign
+    /// - Throws: errors during sisgning process
+    /// - Returns: Signature object containing hex(r) hex(s) hex(v + 27) as one strig of secp256k1 signature
+    static func sign(_ string: String) throws -> Signature {
+        let hash = EthHasher.hash(string.data(using: .ascii)!)
```
DmitryBespalov: I would use `.utf8` encoding instead of ascii encoding

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -7,27 +7,58 @@
...
 
-    init(deviceID: UUID? = nil, safes: [Address], token: String, bundle: String, version: String, buildNumber: String) {
-        self.uuid = deviceID?.uuidString.lowercased()
+    init(deviceID: UUID,
+         safes: [Address],
+         token: String,
+         bundle: String,
+         version: String,
+         buildNumber: String,
+         timestamp: String = String(Int(Date().timeIntervalSince1970 * 1_000))) throws {
```
DmitryBespalov: I would move this conversion from date to string inside of the init. Can we make timestamp as Date or as TimeInterval?

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -7,27 +7,58 @@
...
         self.cloudMessagingToken = token
         self.bundle = bundle
         self.version = version
         self.buildNumber = buildNumber
+        self.timestamp = timestamp
+
+        let string = [
+            "gnosis-safe",
+            self.uuid,
+            self.safes.reduce("") { $0 + $1 },
```
DmitryBespalov: I think what is meant here is `.joined()`

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -7,27 +7,58 @@
...
+            self.cloudMessagingToken,
+            self.bundle,
+            self.version,
+            self.deviceType,
+            self.buildNumber,
+            self.timestamp
+        ]
+        .reduce("") { $0 + $1 }
```
DmitryBespalov: I think what is meant here is `.joined()`

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -7,27 +7,58 @@
...
+
+        if let signature = try? Signer.sign(string).value {
+            self.signatures = [signature]
+        } else {
+            self.signatures = nil
```
DmitryBespalov, line 66: if there are no signatures, then `timestamp` must be omitted

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -7,27 +7,58 @@
...
 struct RegisterNotificationTokenRequest: JSONRequest {
-    let uuid: String?
+    let uuid: String
     let safes: [String]
     let cloudMessagingToken: String
     let bundle: String
     let version: String
     let deviceType: String = "IOS"
     let buildNumber: String
+    let timestamp: String
```
DmitryBespalov: Timestamp must be optional because it is not required (https://docs.google.com/document/d/1RqIfZkqdS_TZ4IQVjeWUhNI1ZguD3NfpohSSRLVxB5Q/edit#heading=h.h09ziydayww)

MultisigTests/Data/Services/Safe Transaction Service/RegisterNotificationTokenRequestTests.swift
```diff
@@ -0,0 +1,33 @@
...
+    func testRequestInitCalculatesProperSignature() throws {
+        let request = try RegisterNotificationTokenRequest(
+            deviceID: UUID(uuidString: "bb30cd3e-e0ad-4e9a-b726-44db67a0820b")!,
+            safes: [Address("0xEefFcdEAB4AC6005E90566B08EAda3994A573C1E")],
+            token: "erXBYb-CxU1jtSvwfZrxqW:APA91bH0IWkMWOGizlbNAwxV6OVjEmNR1feRs2WBT7BE6aVMm2C-x1COKqNYq19t5YNjIzVBKDyVVEqFojlkvEtiSaJA0lCZL0LfuEwfc8p9jfBuM6HG82pczVbnMev1J0gXlB3bIlAP",
+            bundle: "io.gnosis.multisig.dev.rinkeby",
+            version: "2.6.0",
+            buildNumber: "1",
+            timestamp: "1606319110027")
+        XCTAssertEqual(request.signatures, ["460ab62322407376576be061a6bfaaaa78cd1be4e0421d88cd635d0568ff2d473280d0edfd898bf1fd73f3fea8206ef91d6fbf6d9dc63b5d7d1378b8e4059f691c"])
```
DmitryBespalov, line 31: can you also please add a test case for the example data from the spec? https://docs.google.com/document/d/1RqIfZkqdS_TZ4IQVjeWUhNI1ZguD3NfpohSSRLVxB5Q/edit#

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
DmitryBespalov: Also, this argument should be optional

DmitryBespalov (commented):
I have several questions that need resolution

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: This is optional for the backend, but we can always send it. Or should I remove it?

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: ok

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: ok

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: Ah, that what you mean under "optional"? I think it is not required to do so for proper backend calculations, because it will just not participate in the logic. There is no such thing a strict rule not to send some optional params for backends, but there is a rule to always send them if backend expects them to have.

Multisig/Logic/Models/Signer.swift
sche: Ok, I can change. It should not make any change in logic.

Multisig/UI/App/RemoteNotificationHandler.swift
sche: I am not sure what do you mean? The first time we will try to send a register request it will be created and stored between sessions. Default value is taken only when nothing is stored.
Or did you have smth else in mind?

Multisig/UI/App/RemoteNotificationHandler.swift
sche: I removed it because I found that it is requested on every app start anyways. But you are right, before it was better 👍 
I will returen it

MultisigTests/Data/Services/Safe Transaction Service/RegisterNotificationTokenRequestTests.swift
sche: Uxio commented there "I proposed some changes first, if we all agree I will edit and add a real example" and I didn't see a signature in the doc, so I created my test case.
Does it make sense to add another one? If smth wrong with signing, then this one will definitely fail.

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: Having it as string in init param is easier for unit testing and it should generally not be passed manually in production code. Is it ok if I leave it like this?

Multisig/UI/App/RemoteNotificationHandler.swift
DmitryBespalov: I think it is better to set UUID explicitly on the first app launch instead of lazy saving it

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
DmitryBespalov: fine

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
DmitryBespalov: it is not meant to send the timestamp without the signature, please make it optional

MultisigTests/Data/Services/Safe Transaction Service/RegisterNotificationTokenRequestTests.swift
DmitryBespalov: Yes, it makes sense because that example test case must pass on all systems (iOS, Android, Web, Backend)

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
DmitryBespalov: this is how it was agreed upon in the spec, we'd better not assume backend implementation details when we implement our code

MultisigTests/Data/Services/Safe Transaction Service/RegisterNotificationTokenRequestTests.swift
sche: ok. I am adding the test and adding signature data to the requirements

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -50,11 +50,14 @@ struct RegisterNotificationTokenRequest: JSONRequest {
...
             self.deviceType,
             self.buildNumber,
-            self.timestamp
+            self.timestamp ?? ""
         ]
         .joined()
 
         if let signature = try? Signer.sign(string).value {
+            guard timestamp != nil else {
+                throw "'timestamp' parameter is required if signing key exists"
```
DmitryBespalov: maybe `preconditionFailure()` is better because this would be a programmer's error

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: ok

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: ok

Multisig/UI/App/RemoteNotificationHandler.swift
sche: ok

Multisig/Data/Services/Notification Service/UnregisterNotificationTokenRequest.swift
```diff
@@ -17,9 +17,9 @@ struct UnregisterNotificationTokenRequest: JSONRequest {
 
     typealias ResponseType = EmptyResponse
 
-    init(deviceID: UUID, address: Address) {
+    init(deviceID: String, address: Address) {
         self.address = address.checksummed
-        self.deviceID = deviceID.uuidString.lowercased()
+        self.deviceID = deviceID
```
DmitryBespalov: the `lowercased()` got lost - either when this init is called or in here it should be present

Multisig/UI/App/UserDefaultPropertyWrapper.swift
```diff
@@ -17,6 +17,16 @@ struct UserDefault<T> {
     }
 }
 
+@propertyWrapper
+struct UserDefaultWithDefault<T> {
```
DmitryBespalov: not sure if we need this anymore

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
```diff
@@ -7,27 +7,64 @@
...
+    init(deviceID: String,
+         safes: [Address],
+         token: String,
+         bundle: String,
+         version: String,
+         buildNumber: String,
+         timestamp: String?) throws {
+
+        guard UUID(uuidString: deviceID) != nil else {
+            throw "'deviceID' should be UUID string"
```
DmitryBespalov: I would make it preconditionFailure() or just precondition instead of guard because this is developer error

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: ok

Multisig/Data/Services/Notification Service/UnregisterNotificationTokenRequest.swift
sche: I checked - it worked fine. But I will make it lowercased()

Multisig/UI/App/UserDefaultPropertyWrapper.swift
sche: 👍

Multisig/Data/Services/Notification Service/RegisterNotificationTokenRequest.swift
sche: ok

DmitryBespalov (approved):
thank you! 👏

PR #743, opened 2020-12-01
Gh 725/tx details stub

DmitryBespalov:
- Added stylized buttons
- After this PR I'll remove the BackupLoadSafeViewController (will merge it with existing)
- This is just some valuable left overs from the last week's work


Multisig/UI/App/UIKit/TransactionDetailsViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+    private enum TransactionSource {
+        case id(String)
+        case safeTxHash(Data)
+        case data(TransactionViewModel)
+    }
+
+    private var txSource: TransactionSource!
+
+    // disable reacting to change of safes reactsToSelectedSafeChanges
+    override var reactsToSelectedSafeChanges: Bool { false }
```
sche, line 28: why? What is the use case?

Multisig/UI/App/UIKit/TransactionlListViewController/TransactionListViewController.swift
```diff
@@ -37,6 +37,8 @@ class TransactionListViewController: LoadableViewController, UITableViewDelegate
         tableView.registerHeaderFooterView(BasicHeaderView.self)
 
         tableView.sectionHeaderHeight = BasicHeaderView.headerHeight
+        tableView.rowHeight = UITableView.automaticDimension
```
sche, line 40: 👍

sche (approved):

Multisig/UI/App/UIKit/TransactionDetailsViewController.swift
DmitryBespalov: because I think that transaction details vc should not reload when safe changes, it should reload when transaction changes

MouazAlzahabi (approved):

PR #744, opened 2020-12-02
Gh 735 address info view

sche:


Multisig/UI/App/UIKit/UI Library/ReusableCells/AddressDetailsCell.swift
```diff
@@ -9,49 +9,30 @@
...
+        addressInfoView.onDisclosureButtonAction = viewDetails
     }
 
-    func setAddress(_ value: Address) {
-        addressWithTitleView.setAddress(value)
+    func setAddressInfo(_ addressInfo: AddressInfo) {
+        addressInfoView.setAddressInfo(addressInfo)
     }
 
-    func setName(_ value: String) {
```
DmitryBespalov, line 38: I thik the code will be easier to read on the call site if we provide additional helper methods

```
setAddress(_ address: Address, label: String? = nil)
setAddress(_ address: String, label: String? = nil)
```

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
```diff
@@ -8,18 +8,29 @@
...
+    let label: String?
+}
+
+class AddressInfoView: UINibView {
     @IBOutlet private weak var identiconView: UIImageView!
     @IBOutlet private weak var textLabel: UILabel!
     @IBOutlet private weak var detailLabel: UILabel!
-    @IBOutlet private weak var button: UIButton!
+    @IBOutlet private weak var copyButton: UIButton!
+    @IBOutlet private weak var disclosureButton: UIButton!
```
DmitryBespalov: How come do we have disclosure button in address info view? What is it for?

sche: This is for view address details or delete button (imported key address)

DmitryBespalov: I think the name is misleading - can we name it "detailButton"?

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
```diff
@@ -8,18 +8,29 @@
...
+
+    private (set) var addressInfo: AddressInfo!
+
+    var onAddressInfoSelection: (() -> Void)?
+    var onDisclosureButtonAction: (() -> Void)?
 
-    private var style: Style = .nameAndAddress
-    private (set) var address: Address!
+    @IBAction private func onDisclosureButton() {
```
DmitryBespalov: In our code base the code layout is usually this:

```
 public vars
 private vars
 init
public methods
private methods
```

Can you please move these function below init or at the end somewhere?
Also, public vars would be better in the beginning of the class

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
```diff
@@ -8,18 +8,29 @@
...
-class AddressWithTitleView: UINibView {
+struct AddressInfo {
+    let address: Address
+    let label: String?
+}
+
+class AddressInfoView: UINibView {
     @IBOutlet private weak var identiconView: UIImageView!
     @IBOutlet private weak var textLabel: UILabel!
     @IBOutlet private weak var detailLabel: UILabel!
```
DmitryBespalov: I would name this label "addressLabel" then

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
```diff
@@ -8,18 +8,29 @@
...
-    @IBOutlet private weak var button: UIButton!
+    @IBOutlet private weak var copyButton: UIButton!
```
DmitryBespalov, line 22: this is named "copyButton" but the action is named "onAddressInfoButton" - which is the correct one?

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
```diff
@@ -8,18 +8,29 @@
 
 import UIKit
 
-class AddressWithTitleView: UINibView {
+struct AddressInfo {
+    let address: Address
+    let label: String?
+}
+
+class AddressInfoView: UINibView {
```
DmitryBespalov, line 16: I think this class is better to implement the copy and open in safari functionality by default; Plus, a bool flag to hide the "open in browser" button.

That way it will be truly reusable and we can just drop it in everywhere where needed

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
```diff
@@ -30,35 +41,31 @@ class AddressWithTitleView: UINibView {
...
 
-    func setStyle(_ value: Style) {
-        style = value
-        update()
-    }
-
-    func setAddress(_ value: Address) {
-        address = value
-        identiconView.setAddress(value.hexadecimal)
+    func setAddressInfo(_ addressInfo: AddressInfo) {
```
DmitryBespalov, line 36: I think it would be good to have a helper method that could assign address and label right away - from address or string

Multisig/UI/App/UIKit/UI Library/ReusableCells/AddressDetailsCell.swift
sche: this cell is refactored into DetailAccountCell with one method `func setAccount(addressInfo: AddressInfo, title: String?)`
I think there is no need in helper methods at this point of time.

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
sche: ok

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
sche: Ok

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
sche: ok

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
sche: renamed action into `copyAddress`

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
sche: Let's add it in future if there is such need. I have a feeling that for consistency and readability it is better always to operate with view model struct `AddressInfo`

Multisig/UI/App/UIKit/UI Library/ReusableViews/AddressInfoView.swift
sche: I added copy by default. Let's discuss open in Safari by voice

DmitryBespalov: ok

DmitryBespalov (approved):

PR #745, opened 2020-12-02
Gh 725/tx details models

DmitryBespalov:
Refactored the client gateway models so that it takes less code + made use of sub-types - I think it is better now.

I'll gradually change the current use of SCG models with the new models


Multisig/Data/Services/Safe Client Gateway Service/TransactionDetailsRequest.swift
```diff
@@ -44,3 +44,26 @@ extension SafeClientGatewayService {
         asyncExecute(request: TransactionDetailsRequest(safeTxHash: safeTxHash), completion: completion)
     }
 }
+
+struct TransactionDetailsRequestV2:  JSONRequest {
```
sche, line 48: extra space

sche (approved):

Multisig/Data/Services/Safe Client Gateway Service/SCG.swift
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    struct ExecutionInfo: Decodable {
+        var nonce: UInt256String
+        var confirmationsRequired: UInt64
+        var confirmationsSubmitted: UInt64
+        var signers: [AddressString]?
+        var confirmations: [Confirmation]?
+        var missingSigners: [AddressString]?
+
```
MouazAlzahabi, line 41: Extra space

MouazAlzahabi (approved):

Multisig/Data/Services/Safe Client Gateway Service/SCG.swift
DmitryBespalov: will fix in the latest PR

Issue #742, opened 2020-12-01
Transaction Builder and View Controller


DmitryBespalov:
![Screenshot 2020-12-01 at 11.28.14.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/34a7adc0-1dc3-4cf5-b159-f4ff914e876a)

liliya-soroka:
closed as no QA required

PR #755, opened 2020-12-04
gh-742 Added building of transaction data cells

DmitryBespalov:
Implemented building of the screen - all in one function 😬


sche:
It is unusual to read such amount of functions :)

I would prefer old-style with separate functions rather than functions inside functions.

As I understand cells selection is not implemented, right? I also notices that now tx details has more data than in SwiftUI
E.g.
Screenshot 2020-12-04 at 12 40 55

sche (approved):

MouazAlzahabi (approved):

DmitryBespalov:
ok, will adapt. Actions for transfers should not be shown - will change that

DmitryBespalov:
I'll merge now and will continue changes in the next PR

Issue #726, opened 2020-11-24
[UIKit] Push Notifications


DmitryBespalov:
Opening of the transaction details on push notification tap

liliya-soroka:
Verified
app version 2.7.0 (229)

PR #756, opened 2020-12-04
gh-726 Open TX Details for pushes

sche:


Multisig/UI/App/UIKit/UI Library/UIWindow+Extension.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Created by Andrey Scherbovich on 04.12.20.
+//  Copyright © 2020 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+extension UIWindow {
+
+    // https://stackoverflow.com/questions/6131205/how-to-find-topmost-view-controller-on-ios
+    static func topMostController() -> UIViewController? {
```
DmitryBespalov, line 14: just asking: did you test this with various app states?
- switch safes opened
- detail screen opened
- safari vc opened
- safe info overlay opened
- import owner screen opened

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -168,8 +168,8 @@ class RemoteNotificationHandler {
             let appConfig = App.configuration.app
             var timestamp: String?
             if let _ = try? App.shared.keychainService.data(forKey: KeychainKey.ownerPrivateKey.rawValue) {
-                // add timestamp if there is a signing key
-                timestamp = String(Int(Date().timeIntervalSince1970 * 1_000))
+                // add timestamp in seconds if there is a signing key
```
DmitryBespalov: now this comment is not needed, the code says it

Multisig/UI/App/RemoteNotificationHandler.swift
```diff
@@ -168,8 +168,8 @@ class RemoteNotificationHandler {
...
+                timestamp = String(Int(Date().timeIntervalSince1970))
```
DmitryBespalov: I think that potentially the Int(Double) can crash because of the big value of Double (scientific notation) not fitting into Int. It is better to format the Double as a value  without decimal point

Multisig/UI/App/UIKit/UI Library/UIWindow+Extension.swift
sche: I have not tested it properly. We just managed to fix pushes, so I will continue on Monday with it

Multisig/UI/App/RemoteNotificationHandler.swift
sche: ok

Multisig/UI/App/RemoteNotificationHandler.swift
sche: agree

Multisig/UI/App/UIKit/UI Library/UIWindow+Extension.swift
sche: I have checked all mentioned use cases.

DmitryBespalov (approved):

Issue #723, opened 2020-11-24
Don't scroll up tx list when coming back from tx details


DmitryBespalov:
**UIKit only.**

When transaction list is not scrolled, and user coming back from the transaction details, then reload the list automatically.

liliya-soroka:
Reopened:
The tx list should be scrolled up and updated on the tab switch 

app version 2.7.0 (227)

liliya-soroka:
closed
verified, the tx list is not scrolled up and not updated if the tx details are opened

PR #757, opened 2020-12-04
gh-723 Add reload feature

MouazAlzahabi:
Data will be reloaded tx list is not scrolled, even when switch tabs


DmitryBespalov:
@MouazAlzahabi can it be reloaded only if coming back from the tx details? We don't need to reload on switching tabs

DmitryBespalov (approved):

Issue #761, opened 2020-12-07
Unified tx list: Load data


MouazAlzahabi:
- [x] Create new structure for both History&Queued tx endpoint
  - [x] separator header with text label (for next tx and for queued tx's list, for history tx's grouping)
  - [x] item type for conflict header 
    - [x] nonce
- [x] Load queued and next tx's using client gw endpoint for queued tx's
- [x] Load history tx's using client gw endpoint for history tx's
- [x] Support pagination
- [x] tx object's should be extended with conflictType field (values e.g. None, WithNext, WithoutNext)

- Endpoints:
  - https://safe-client.rinkeby.staging.gnosisdev.com/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/transactions/history
  - https://safe-client.rinkeby.staging.gnosisdev.com/v1/safes/0x1230B3d59858296A31053C1b8562Ecf89A2f888b/transactions/queued

liliya-soroka:
Verified
app version 2.8.0 (259)

PR #781, opened 2020-12-10
gh-761 Load data from backend

MouazAlzahabi:


DmitryBespalov:
@MouazAlzahabi can you bump version number to 2.8.0 on this PR?

Multisig/Data/Services/Safe Client Gateway Service/HistoryTransactionsSummaryListRequest.swift
```diff
@@ -0,0 +1,46 @@
...
+        try execute(request: HistoryTransactionsSummaryListRequest(address))
+    }
+
+    func historyTransactionsSummaryList(address: String) throws -> HistoryTransactionsSummaryListRequest.ResponseType {
+        try execute(request: HistoryTransactionsSummaryListRequest(safeAddress: address))
+    }
+
+    func historyTransactionsSummaryList(pageUri: String) throws -> HistoryTransactionsSummaryListRequest.ResponseType {
+        return try execute(request: PagedRequest<TransactionSummaryItemWrapper>(pageUri))
+    }
```
DmitryBespalov: I don't think these 3 methods will be used anymore - we'll transition to async calls. Can we remove them?

Multisig/Data/Services/Safe Client Gateway Service/QueuedTransactionsSummaryListRequest.swift
```diff
@@ -0,0 +1,46 @@
...
+        try execute(request: QueuedTransactionsSummaryListRequest(address))
+    }
+
+    func queuedTransactionsSummaryList(address: String) throws -> QueuedTransactionsSummaryListRequest.ResponseType {
+        try execute(request: QueuedTransactionsSummaryListRequest(safeAddress: address))
+    }
+
+    func queuedTransactionsSummaryList(pageUri: String) throws -> QueuedTransactionsSummaryListRequest.ResponseType {
+        return try execute(request: PagedRequest<TransactionSummaryItemWrapper>(pageUri))
+    }
```
DmitryBespalov: here as well

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -14,6 +14,68 @@ struct Page<T: Decodable>: Decodable {
     let results: [T]
 }
 
+protocol TransactionSummaryItem: Decodable { }
+
+
+struct TransactionSummaryItemDateLabel: TransactionSummaryItem {
```
DmitryBespalov: I wanted to transition from this file's models to the SCG.swift models. Does it make sense to make it in this release or make it as a separate refactoring?

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -14,6 +14,68 @@ struct Page<T: Decodable>: Decodable {
...
+struct TransactionSummaryItemTransaction: TransactionSummaryItem {
+    let transaction: TransactionSummary
+    let conflictType: ConflictType
+}
+
+struct TransactionSummaryItemConflictHeader: TransactionSummaryItem {
+    let nonce: UInt256String
+}
+
+enum TransactionSummaryItemWrapper: Decodable {
```
DmitryBespalov: There's a new way I found how to use this decoding, without a protocol but using just enum directly - please see the SCG.swift or let's have a call and I can assist you

Multisig/Data/Services/Safe Client Gateway Service/SCG.swift
```diff
@@ -12,6 +12,49 @@ import Foundation
...
+        case dateLabel(TransactionSummaryItemDateLabel)
+        case label(TransactionSummaryItemLabel)
+        case transaction(TransactionSummaryItemTransaction)
+        case conflictHeader(TransactionSummaryItemConflictHeader)
+        case unknown
+
+        init(from decoder: Decoder) throws {
+            enum Keys: String, CodingKey { case type }
+            let container = try decoder.container(keyedBy: Keys.self)
+            let type = try container.decode(TransactionSummaryItemType.self, forKey: .type)
```
DmitryBespalov: I would use plan strings in the Switch instead of the enum because if new enum strings are introduced on the backend, then the app would just decode them as unknown (in default case), otherwise this line will throw exception

MouazAlzahabi: In that case how about make type as optional and add default case to switch ?

DmitryBespalov: works for me but then it is inconsistent with other code

sche (approved):

DmitryBespalov (approved):

PR #791, opened 2020-12-16
Gh 787/snackbar crash debug

DmitryBespalov:
Crash report: https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/ios:io.gnosis.multisig.prod.mainnet/issues/b351e84478f859565c2f2f974d673d1f?time=last-seven-days&sessionEventKey=6f18296829174cde9e15a2722c6a01b7_1485135971032942665

I don't know what is the reason. 

Things that I thought of:
- Threading issue? I checked - it looks like it should always be called on main thread. I added precondition checks.
- Optional unwrapping issue? - I changed the implicit to explicit unwrapping
- What else could it be?


sche:
hm. The use was on Enter seed phrase screen (pop over). I don't know how it could happed that a snackbar message appeared there. Can it be some glitch with timer? Or could a message be stuck while user was switching tabs/screens?

DmitryBespalov:
snackbar is shown on top of everything else, including any opened screens, so the message is not covered by modal screens.

Issue #816, opened 2020-12-23
[iOS] Uniform tx list:Wrong name for Queue and Next transaction sections on Queue tab


liliya-soroka:
**Bug description**
Wrong name for Queue and Next transaction sections on Queue tab:
next instead of NEXT TRANSACTION
Queued instead of QUEUE
**Steps To Reproduce**
1. Go to the Queue tab in the transactions when there are txs in both sections
Current result:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/8e82a00b-4429-4c40-a8d9-8128ba26d1b0)
**Expected Result**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/a5a2f912-0627-4a21-ad62-ec2f7fe5af26)

**Device & App version**
 - Device/OS: iPhone X/14.2
 - App version: 2.8.0 (244)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.8.0 (259)

PR #818, opened 2021-01-04
gh-816 Fix issue

MouazAlzahabi:
I didn't consider labels as types here because I thought it's something related more to ui rather than model, WDYT about such implementation


DmitryBespalov:
I think, semantically, that "label" field is actually type identifier and not a string or header, therefore it's like an enum to me. While the approach in this PR works, it feels a bit confusing with what backend is expecting

DmitryBespalov (approved):
however, I'm ok with that since this is easy to fix if needed

sche (approved):

Issue #807, opened 2020-12-22
[iOS] Old view of internet connection error on the Enter ENS name page


liliya-soroka:
**Bug description**
Old view of internet connection error on the Enter ENS name page
**Steps To Reproduce**
1. Turn the internet connection off
2. Go to the Add safe->enter ENS
3. Enter ENS name
Current result: The message in the old format is displayed
**Expected Result**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/4d301686-8151-4f41-96bf-fea2ee948b5a)
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/cdf8a6d0-582f-42af-a7b4-ccc9ce2ee497)
**Device & App version**
 - Device/OS: Iphone X/14.2
 - App version: 2.8.0 (244)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.8.0 (259)

PR #823, opened 2021-01-05
gh-807 detailed web3 response errors

sche:


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -340,4 +340,12 @@ enum GSError {
         let loggable = true
     }
 
+    struct ThirdPartyError: DetailedLocalizedError {
+        let description = "Third party library error"
+        let reason: String
+        let howToFix = "Please try again later or contact Safe support if this issue persists"
```
DmitryBespalov, line 346: Just to make sure, are you also updating the spec doc to include the new error reasons & codes?

sche: Yes, I do

DmitryBespalov (approved):

Issue #805, opened 2020-12-22
[iOS] Multisig:The app is crashed if the "Learn more" is clicked on Queue for conflicted txs


liliya-soroka:
**Bug description**
The app is crashed if the "Learn more" is clicked on Queue for conflicted txs - https://console.firebase.google.com/u/1/project/safe-firebase-staging/crashlytics/app/ios:io.gnosis.multisig.staging.rinkeby/issues/19328bf8eaab9655cb84b72a881efe5e?time=last-seven-days&sessionEventKey=d16b778d2fab47728e09c9b46ba25a75_1487420650391082227
**Steps To Reproduce**
1. Create a few groups of conflicted txs in Queue section:
- 1st option reject as one of the user tx
- 2nd option : 
 -select send funds , enter all required information and click submit ->don't sign in MM, leave the signing request for later
 - select send funds again and create new signing request in MM ( you may create many such requests)
 - sign all requests in MM one by one
Current result: all signing txs have the same nonce in the app
2. Click "Learn more" in each of the conflicted group 

Current result: The app is crashed

**Expected Result**
If no page with info, nothing should happen 
**Screenshots**

**Device & App version**
 - Device/OS: iPhone X/ 14.2
 - App version: 2.8.0 (244)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.8.0 (259)

PR #824, opened 2021-01-05
gh-805 Fix issue

MouazAlzahabi:


Multisig/UI/App/UIKit/Transaction/TransactionListViewController/Cells/TransactionsListConflictHeaderTableViewCell.swift
```diff
@@ -26,6 +26,15 @@ class TransactionsListConflictHeaderTableViewCell: UITableViewCell {
     }
 
     @IBAction func learnMoreButtonTouched(_ sender: Any) {
-        UIApplication.shared.sendAction(#selector(UIViewController.didTapExternalURLCell(_:)), to: nil, from: self, for: nil)
+        UIApplication.shared.sendAction(#selector(UIViewController.didTapTransactionsListConflictHeaderTableViewCell(_:)), to: nil, from: self, for: nil)
```
DmitryBespalov: hm... instead of this functionality duplication, why don't we create a protocol "ExternalURLSource" that has the "url" getter variable, and make the didTapExternalURLCell's argument Any. Inside that function, then, we can if-let cast the sender to the ExternalURLSource and obtain the url.

That way it would be generic and extendable.

MouazAlzahabi: Honestly I thought to make it this way first :D, but I changed my mind to minimize the changes, but I will change right awat

Multisig/UI/App/UIKit/Transaction/TransactionDetailsViewController/Transaction Detail Cells/ExternalURLCell.swift
```diff
@@ -22,14 +22,6 @@ class ExternalURLCell: UITableViewCell {
...
-        UIApplication.shared.sendAction(#selector(UIViewController.didTapExternalURLCell(_:)), to: nil, from: self, for: nil)
-    }
-}
-
-extension UIViewController {
-    @objc func didTapExternalURLCell(_ sender: ExternalURLCell) {
-        if let url = sender.url {
-            openInSafari(url)
-        }
+        UIApplication.shared.sendAction(#selector(UIViewController.didTapExternalURL(_:)), to: nil, from: self, for: nil)
```
DmitryBespalov: even more cool, we can now have an extension on the "ExternalURLSource" and wrap this call in a method, so that we can just call "openExternalURL()"

DmitryBespalov (approved):

Issue #814, opened 2020-12-23
[iOS] 42200 error is displayed in case of 42201 and 42250


liliya-soroka:
**Bug description**
42200 error is displayed in the case of 42201 and 42250
**Steps To Reproduce**
1. Simulate the error 42201 and 42250 
Current result:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/c066e94d-8148-4c33-bc0c-788464b13b56)
**Expected Result**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/36a58616-23a5-4f02-b427-45589c3c9149)
**Screenshots**

**Device & App version**
 - Device/OS: iPhone X/14.2
 - App version: 2.8.0 (5)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.8.0 (277)

PR #826, opened 2021-01-06
gh-814 small improvement

sche:


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -116,8 +116,9 @@ enum GSError {
             case 50:
                 return UnprocessableEntity(reason: "Safe info is not found.", code: 42250)
             default:
-                LogService.shared.error("Unrecognised error with code: \(error.code); message: \(error.message)",
-                                        error: unexpectedError)
+                LogService.shared.error(
+                    "Unrecognised error with code: \(error.code); message: \(String(describing: error.message))",
```
DmitryBespalov: if the message is optional, will the text have the "message: Optional<nil>" or will it be empty?

sche: good catch

sche: fixed

DmitryBespalov (approved):

Issue #649, opened 2020-10-16
[iOS]Valid seed phrase is defined as invalid if it contains more then 12  and less than 24 words


liliya-soroka:
**Bug description**
The valid seed phrase is defined as invalid if it contains more then 12  and less than 24 words

**Steps To Reproduce**
1. Go to the app settings
2. Import owner key
3. use https://iancoleman.io/bip39/#english to generate seed phrase with 15,16,18,21 words
4. enter seed phrase from step 3 and click Next
5. Invalid seed phrase error appears

**Expected Result**
The seed phrase may contain from 12 to 24 words

**Screenshots**

**Device & App version**
 - Device/OS: Iphone X/ 13.7
 - App version: 2.3.0 (165)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

rmeissner:
>  with 15,16,18,21 words

16 should not be valid, right? afaik it needs to be a multiple of 3 ;)

liliya-soroka:
Verified
app version 2.8.0 (266)

PR #829, opened 2021-01-06
gh-649 fix error

sche:


DmitryBespalov (approved):

Multisig/Logic/Models/PrivateKey/web3swift/BIP39.swift
```diff
@@ -117,7 +117,7 @@ public class BIP39 {
     static public func mnemonicsToEntropy(_ mnemonics: String,
                                           language: BIP39Language = BIP39Language.english) -> Data? {
         let wordList = mnemonics.components(separatedBy: " ")
-        guard wordList.count >= 12 && wordList.count.isMultiple(of: 4) else {return nil}
+        guard wordList.count >= 12 && wordList.count.isMultiple(of: 3) else {return nil}
```
DmitryBespalov: really? wasn't it the library code? are we sure this is valid?

DmitryBespalov: maybe comment/ explanation or extracting this into parameter/constant would help

MouazAlzahabi (approved):

Multisig/Logic/Models/PrivateKey/web3swift/BIP39.swift
sche: yes, this is a bug in the lib. I created an issues there and a pr to fix it https://github.com/matter-labs/web3swift/pull/301

I think there is no need in a separate variable.

DmitryBespalov (approved):

PR #835, opened 2021-01-07
Gh 783/uikit action details

DmitryBespalov:
Implemented the tx action details, multi send tx list and multi send tx details

### DIFF
--- a/Multisig/UI/App/UIKit/Transaction/ActionDetailViewController.swift
+++ b/Multisig/UI/App/UIKit/Transaction/ActionDetailViewController.swift
@@ -13,6 +13,20 @@ class ActionDetailViewController: UITableViewController {
     private var data: DataString?
     private var customTitle: String?
 
+    private static let indentWidth: CGFloat = 8.0
+
+    /// Container for all cells in the table
+    private var cells = [UITableViewCell]()
+
+    /// Cells' index stack.
+    ///
+    /// The index will be popped during cell creation in builder methods.
+    ///
+    /// This allows to temporary change the insertion point of the cells
+    /// and go back to previous insertion point
+    /// by popping the pushed index.
+    private var builderIndexStack = [0]
+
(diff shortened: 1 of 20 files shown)


Multisig/UI/App/UIKit/Transaction/ActionDetailViewController.swift
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    private func insert(_ cell: UITableViewCell, at index: Int) {
+        cells.insert(cell, at: index)
+    }
+
+    private func insert(_ value: [UITableViewCell], at index: Int) {
+        cells.insert(contentsOf: value, at: index)
+    }
+
```
MouazAlzahabi, line 122: extra line

Multisig/UI/App/UIKit/Transaction/ActionDetailViewController.swift
```diff
@@ -0,0 +1,309 @@
...
+                self.tableView.beginUpdates()
+                let insertedPaths = (insertionIndex..<insertionIndex.advanced(by: cell.subcells.count)).map { IndexPath(row: $0, section: 0) }
+                self.tableView.insertRows(at: insertedPaths, with: .bottom)
+                self.tableView.endUpdates()
+
+            case .expanded:
+                // then collapse the contents
+                cell.state = .collapsed
+
+                // visit all subcell tree and collapse all
```
MouazAlzahabi, line 260: Not sure, but from UX perspective I think we can keep the subcells expanded and just collapse the parent.

MouazAlzahabi (approved):

Multisig/UI/App/UIKit/Transaction/ActionDetailViewController.swift
DmitryBespalov: agree, but it'll be more complicated to implement, I'm leaving this as is

PR #877, opened 2021-01-21
Gh 865 enter seed screen

sche:


DmitryBespalov:
@sche is this the final iteration fo this screen or more work needed on UI?
- The "Next" button is always enabled, but it should enable only when the input is valid
- The error is not showing at all
- On iPod touch the text view is covered by the keyboard and not visible at all
- Please enable the keyboard opening on the screen open automatically (`textView.becomeFirstResponder()`)

sche:
@DmitryBespalov strange, I will take a look regarding not showing error and enables Next button. It worked fine for me.
I will enable keyboard automatically

sche:
@DmitryBespalov I checked - it works fine for me in iPod touch. No issues with next button and displaying errors.
I decided not to use textView.becomeFirstResponder() as then user needs to scroll to read the text.

DmitryBespalov:
@posthnikova is the keyboard should be enabled automatically when opening the seed phrase screen?

DmitryBespalov (approved):

posthnikova:
@DmitryBespalov Yes, that would be nice

MouazAlzahabi (approved):

MouazAlzahabi (approved):

PR #878, opened 2021-01-21
Gh 861/signing v2 discoverability safe loaded

MouazAlzahabi:


Multisig/Logic/Models/Multisig.xcdatamodeld/Multisig.xcdatamodel/contents
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17511" systemVersion="19H2" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="17709" systemVersion="20C69" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
```
DmitryBespalov: we need to test upgrade from previous version because the data model changed

Multisig/UI/Safe Management/Add Safe/EnterSafeNameViewController.swift
```diff
@@ -48,7 +48,16 @@ class EnterSafeNameViewController: UIViewController {
     @objc private func didTapNextButton() {
         guard let name = name, let address = address else { return }
         Safe.create(address: address.checksummed, name: name)
-        completion()
+        if let _ = App.shared.settings.signingKeyAddress {
+            completion()
+        } else if AppSettings.hasShownImportKeyOnboarding() {
```
DmitryBespalov: I'm confused... why do we need to show the Enter seed phrase screen after entering safe name?

DmitryBespalov:
- I've got a push notification request - I think we should delay it until this flow is finished completely (skip / import key)
- screen title is missing
- please double-check the fonts - they look different from designs
- the address label is not centered 
- spacing to the skip button is too big
- skip button font should not be bold
- Maybe it's a good idea to hide the tab bar when entering "load safe" flow and then show it again? now it looks different from the designs, but design assumes we show the whole thing in a modal. I think this needs to be clarified.

Multisig/Logic/Models/AppSettings.swift
```diff
@@ -33,6 +33,19 @@ extension AppSettings {
         let appSettings = current()
         return appSettings.termsAccepted
     }
+
+    static func showImportKeyOnboarding() {
```
DmitryBespalov: do you think it is better to have these as vars or it's ok like this?

Multisig/UI/Safe Management/Add Safe/EnterSafeNameViewController.swift
MouazAlzahabi: You are right

Multisig/Logic/Models/Multisig.xcdatamodeld/Multisig.xcdatamodel/contents
MouazAlzahabi: Do you have preferred alternative ?

Multisig/Logic/Models/AppSettings.swift
MouazAlzahabi: I tried to do same as we used to have "termsAccepted", but I agree that properties are better, shall I change for both?

DmitryBespalov: I think yes. @sche what do you think?

sche: I am fine with vars

MouazAlzahabi:
@DmitryBespalov 
Regarding push notification request, I think it's better to keep it like this as at this point safe will be already created and it should be registered for PN, and this screen is part of importing key feature not creating safe, if you don't agree let's create this in another ticket because I think this needs to be tested more

Multisig/Logic/Models/Multisig.xcdatamodeld/Multisig.xcdatamodel/contents
MouazAlzahabi: I just kept it as we have another new field added in this sprint

sche (approved):

PR #879, opened 2021-01-21
Gh 867/key picker vc

DmitryBespalov:

### DIFF
--- a/Multisig/UI/Settings/KeyPickerController/ButtonTableViewCell.swift
+++ b/Multisig/UI/Settings/KeyPickerController/ButtonTableViewCell.swift
@@ -12,6 +12,15 @@ class ButtonTableViewCell: UITableViewCell {
     var onTap: () -> Void = {}
 
     @IBOutlet private weak var button: UIButton!
+    @IBOutlet private weak var buttonHeight: NSLayoutConstraint!
+
+    var height: CGFloat {
+        get { buttonHeight.constant }
+        set {
+            buttonHeight.constant = newValue
+            setNeedsLayout()
+        }
+    }
 
     func setText(_ text: String, onTap: @escaping () -> Void) {
         self.onTap = onTap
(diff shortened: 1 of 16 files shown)


Multisig/UI/Settings/KeyPickerController/KeyPickerController.swift
```diff
@@ -0,0 +1,188 @@
...
+            viewModel.generateAddressesPage()
+
+            if !viewModel.canLoadMoreAddresses {
+                deleted = [IndexPath(row: Index.showMore, section: Section.showMore)]
+            }
+        }
+
+        tableView.beginUpdates()
+        do {
+            tableView.deleteRows(at: deleted, with: .automatic)
```
sche, line 167: why so advanced? :) You could just redraw whole table view

sche (approved):
👍🏻

Multisig/UI/Settings/KeyPickerController/KeyPickerController.swift
DmitryBespalov: because it has good animation :)

Multisig/UI/Settings/KeyPickerController/ButtonTableViewCell.swift
```diff
@@ -0,0 +1,33 @@
...
+        }
+    }
+
+    func setText(_ text: String, onTap: @escaping () -> Void) {
+        self.onTap = onTap
+        button.setText(text, GNOButtonStyle.plain.font(.gnoFont(forTextStyle: .body)))
+    }
+
+    @IBAction func didTapButton(_ sender: Any) {
+        onTap()
```
MouazAlzahabi, line 31: I would create delegate for the cell instead of callback

MouazAlzahabi (approved):

PR #885, opened 2021-01-25
Gh 866 confirm pk screen

sche:


Multisig/Assets.xcassets/ico-confirm-thin.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
```
MouazAlzahabi, line 2: This one is exactly same as checkmark.circle right?

Multisig/UI/Settings/ImportOwnerKey/PrivateKeyController.swift
```diff
@@ -0,0 +1,31 @@
...
+//  PrivateKeyController.swift
+//  Multisig
+//
+//  Created by Andrey Scherbovich on 25.01.21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+class PrivateKeyController {
```
MouazAlzahabi, line 11: Do we need to have this separated in controller? 
Isn't app settings enough?

If we want to go with this, how about moving remove key here also ?

Multisig/Assets.xcassets/ico-confirm-thin.imageset/Contents.json
sche: not exactly. It is thinner and has small margins 🤷‍♂️

Multisig/UI/Settings/ImportOwnerKey/PrivateKeyController.swift
sche: Yes, placing removing key logic is a good idea

Multisig/Assets.xcassets/ico-confirm-thin.imageset/Contents.json
MouazAlzahabi: Ah I see, If you want let's make sure if we can use the same icon

MouazAlzahabi (approved):

PR #890, opened 2021-01-25
gh-872 Add read-only label

MouazAlzahabi:


Multisig/UI/App/HeaderViewController/SafeBarView.swift
```diff
@@ -16,11 +16,13 @@ class SafeBarView: UINibView {
     @IBOutlet private weak var textLabel: UILabel!
     @IBOutlet private weak var detailLabel: UILabel!
     @IBOutlet private weak var button: UIButton!
+    @IBOutlet weak var readOnlyLabel: UILabel!
```
sche, line 19: could you make it private for consistency?

Multisig/UI/App/HeaderViewController/HeaderViewController.swift
```diff
@@ -78,7 +78,7 @@ final class HeaderViewController: ContainerViewController {
 
             if let safe = selectedSafe {
                 safeBarView.setAddress(safe.addressValue)
-                safeBarView.setName(safe.displayName)
+                safeBarView.setName(safe.displayName, readOnly: App.configuration.toggles.signing)
```
sche, line 81: What is the logic behind here? I think this toggle can be removed at all.
We need to check it an owner is imported for this safe. This might be tricky because we do not store safe owners, but not we need to store them and update periodically (probably on app start & safe settings screen loading).

We need to align on update logic with Android here - might be a good topic for requirements refinement meeting.

Issue #863, opened 2021-01-18
[iOS] Signing V2: Discoverability: Banner


biafra23:
## Banner
- Display banner above assets 
   - if the following conditions are met:
     - For app update: No owner key currently installed && banner has never been dismissed && have a safe loaded
     - For new installs: No owner key was ever installed && banner has never been dismissed && have a safe loaded
  - **Import owner key**  We added signing support to the app! Now you can import your owner key and sign transactions on the go.
  - Buttons: 
    - Import owner key -> Onboarding
    - Cross in the upper left corner to dismiss banner
- Store whether the banner has been dismissed

## Tracking 
(see: https://github.com/gnosis/safe/wiki/Tracking)
- `user_banner_owner_skip` Triggered when clicking the x on the banner to ask users to import their owner key.
- `user_banner_owner_import` Triggered when clicking the Import on the banner to ask users to import their owner key.

## Design
https://zpl.io/b6XOZ8R

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/15e5b4d2-db53-41ed-aaf0-4f56bc460063)

**Updates**: 
When the user starts the import process but does not complete it, the banner should not be shown again (Dirk, 25.1.2021)

liliya-soroka:
Verified
app version 2.9.0 (292)

All tracking will be tested together after the app is intalled locally

PR #891, opened 2021-01-26
gh-863 display banner logic

sche:


Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
```diff
@@ -24,7 +24,8 @@ class BalancesViewController: LoadableViewController, UITableViewDelegate, UITab
 
     override var isEmpty: Bool { results.isEmpty }
 
-    private var showingBanner = true
+    @UserDefault(key: "io.gnosis.multisig.importKeyBannerWasShown")
```
MouazAlzahabi, line 27: We have this UserDefault in multi places of the app, are you up to create separated manager for it now same as we have for the settings and gradually refactor it?

MouazAlzahabi (approved):

Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
sche: I am not sure what do you mean. We can discuss it and lets see if we can make it better

Issue #893, opened 2021-01-27
[iOS]Import owner key doesn't work for the first added safe


liliya-soroka:
**Bug description**
Nothing happens if Import owner key is clicked for the first added safe after clean app install
**Steps To Reproduce**
1. Uninstall app
2. Install the lates version
3. Accept T&C and load safe
4. Use valid ens name "liltestsafe.eth" to add safe
5. Enter safe name and go to onboarding screen
6. Click "Import owner key"
Current result: Nothing happens

**Expected Result**
The first intro page of import owner key flow should be opened
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/14.3
 - App version: 2.9.0 (289)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.9.0 (292)

Issue #894, opened 2021-01-27
[iOS] Possible to add the same safe many time 


liliya-soroka:
**Bug description**
The user stucks on the Enter safe name page and after pressing Next many time the same safe is added multiple times
**Steps To Reproduce**
1. Add the first safe and skip import owner key flow
2. try to add the second safe ( I used ens name tttk.eth)
3. Click Next on the enter safe name
Current result: Nothing happens

4. Click Next many times and swipe the page down to close it
5. Open  switch safe
Current result: The same safe is added many times

**Expected Result**
Next should link on the page from the switch safe was called to add new safe
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/f7a20bb5-6c0f-4c96-ab10-b60911d08de3)
**Device & App version**
 - Device/OS: iPhone 11/14.3
 - App version: 2.9.0 (289)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.9.0 (292)

PR #895, opened 2021-01-27
gh-893 fix bug

sche:


Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
```diff
@@ -52,6 +52,14 @@ class BalancesViewController: LoadableViewController, UITableViewDelegate, UITab
         }
 
         emptyView.setText("Balances will appear here")
+
+        NotificationCenter.default.addObserver(
+            self, selector: #selector(ownerKeyImported), name: .ownerKeyImported, object: nil)
+    }
+
+    @objc private func ownerKeyImported() {
```
MouazAlzahabi, line 60: I believe after all discussions all what we need to do here is just dismiss the banner on any of its keys press, import/close.

MouazAlzahabi (approved):

Multisig/UI/Safe Management/Add Safe/EnterSafeNameViewController.swift
```diff
@@ -48,14 +48,14 @@ class EnterSafeNameViewController: UIViewController {
...
-        if let _ = App.shared.settings.signingKeyAddress {
-            completion()
-        } else if !AppSettings.hasShownImportKeyOnboarding {
+        if !AppSettings.hasShownImportKeyOnboarding && App.shared.settings.signingKeyAddress == nil {
             let vc = SafeLoadedViewController()
             vc.completion = completion
             vc.hidesBottomBarWhenPushed = true
             show(vc, sender: self)
             AppSettings.hasShownImportKeyOnboarding = true
+        } else {
```
MouazAlzahabi, line 57: @sche this will solve the bug #894 also, please just close the issue once you merge it

Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
sche: BalancesViewController created before SafeLoadedViewController is closed. So if the key was imported, we need to refresh it.

Issue #905, opened 2021-02-01
Show app review dialog on app start


elgatovital:
- [x] trigger in-app review dialog on 3rd app start (for the first time)
don't count app starts triggered by a push notification
- [x] after the dialog is triggered for the first time use the same logic as before for subsequent in-app reviews (to handle cases when the user canceled the flow). This means count to 3 again, not including starts from push notifications

DmitryBespalov:
>  how to check if the user left a review?

There is no way to check that user left a review.

> when to show dialog again if the review wasn't submitted? Does API handle it?

The app does not have control over the dialog showing, it only can request it from the iOS at certain points in the app flow and hope that the OS will show the dialog. The showing of the dialog is subject to iOS policies.

If app needs to open the review from a menu, it should navigate to the App Store app's review screen.

elgatovital:
>The app does not have control over the dialog showing, it only can request it from the iOS at certain points in the app flow and hope that the OS will show the dialog. The showing of the dialog is subject to iOS policies.

Would it be reasonable to invoke in-app review ui on each app start and let iOS policies do the rest?

> If app needs to open the review from a menu, it should navigate to the App Store app's review screen.

there is already a ticket for that: https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/904

DmitryBespalov:
@elgatovital we can invoke the review, and per the policies:

> The system automatically limits the display of the prompt to three occurrences per app within a 365-day period.

https://developer.apple.com/design/human-interface-guidelines/ios/system-capabilities/ratings-and-reviews/

Other guidelines:
- Try to make the request at a time that will not interrupt what the user is trying to achieve in your app. For example, at the end of a sequence of events that the user has successfully completed.
- Avoid showing a request for a review immediately when a user launches your app, even if it is not the first time that it launches.

elgatovital:
Then I propose this for iOS
- trigger in-app review dialog on 3rd app start (for the first time)
don't count app starts triggered by a push notification
- after dialog is triggered for the first time use same logic as before for subsequent in-app reviews (to handle cases when user canceled the flow). This means count to 3 again, not including starts from push notifications
@DmitryBespalov

DmitryBespalov:
@elgatovital I'm OK with that

elgatovital:
description updated

liliya-soroka:
[...]

Issue #904, opened 2021-02-01
Add rate the app item to app settings


elgatovital:
- [x] Add `Rate the app` menu item to app settings
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/0ee58513-8697-412b-98f8-69859ab26fda)

  - [https://zpl.io/2jA1J4W]( https://zpl.io/2jA1J4W)
  - [zpl://screen?pid=5e8f27b9963e55bb79b13634&sid=600969288f9503a2c54c0a02](zpl://screen?pid=5e8f27b9963e55bb79b13634&sid=600969288f9503a2c54c0a02)

- [x] Navigate to AppStore, when clicked

liliya-soroka:
Verified
app version 2.9.0 - (302)

PR #907, opened 2021-02-03
Gh 904/settings rate app

DmitryBespalov:
Implements #904 and #905

### DIFF
--- a/Multisig/Cross-layer/Configuration/Config.Example.xcconfig
+++ b/Multisig/Cross-layer/Configuration/Config.Example.xcconfig
@@ -131,6 +131,8 @@ HELP_CENTER_URL         = https:/$()/help.gnosis-safe.io/
 FEATURE_SUGGESTION_URL  = https:/$()/safe.cnflx.io/
 CONTACT_EMAIL           = safe@gnosis.io
 
+APP_STORE_REVIEW_URL = https:/$()/apps.apple.com/app/id1515759131?action=write-review
+
 // App name on the home screen (will be used for CFBundleDisplayName)
 DISPLAY_NAME = $(_DISPLAY_NAME_$(SERVICE_ENV)_$(NETWORK):default=$(_DISPLAY_NAME_PROD_MAINNET))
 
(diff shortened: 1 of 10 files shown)


Multisig/Logic/AppReviewController.swift
```diff
@@ -0,0 +1,37 @@
...
+    /// Trigger in-app review dialog on 3rd app start (for the first time)
+    ///      - don't count app starts triggered by a push notification
+    ///
+    /// After the dialog is triggered for the first time, use the same logic
+    /// as before for subsequent in-app reviews
+    /// (to handle cases when the user canceled the flow).
+    ///
+    /// This means count to 3 again, not including starts from push notifications.
+    func pullAppReviewTrigger() {
+        guard AppSettings.termsAccepted else { return }
```
sche, line 26: nice :)

Multisig/Logic/Models/AppSettings.swift
```diff
@@ -21,58 +21,49 @@ extension AppSettings {
...
+        self.path = path
     }
 
-    static var importedOwnerKey: Bool {
+    var wrappedValue: T {
         get {
             dispatchPrecondition(condition: .onQueue(.main))
-            return current().importedOwnerKey
+            let object = AppSettings.current()
+            return object[keyPath: path]
```
sche, line 60: advanced magic 👍🏻

sche (approved):
👍🏻

Multisig/Cross-layer/Configuration/Config.Example.xcconfig
```diff
@@ -131,6 +131,8 @@ HELP_CENTER_URL         = https:/$()/help.gnosis-safe.io/
 FEATURE_SUGGESTION_URL  = https:/$()/safe.cnflx.io/
 CONTACT_EMAIL           = safe@gnosis.io
 
+APP_STORE_REVIEW_URL = https:/$()/apps.apple.com/app/id1515759131?action=write-review
```
MouazAlzahabi: Just wanted to note that using app store id will make this url work only for current app but not others, ie. staging Rinkeby..

you can do same thing using the bundle id using the following url formate 
https://itunes.apple.com/@StoreID/lookup?bundleId=@BundleID

Easy way to get storeID as explained here: 
https://stackoverflow.com/a/61597181/5093995

MouazAlzahabi (approved):

Multisig/Cross-layer/Configuration/Config.Example.xcconfig
DmitryBespalov: nice catch, will update the setting

PR #910, opened 2021-02-04
Gh 798 known addresses in tx list

sche:


Multisig/App/AppDelegate.swift
```diff
@@ -25,6 +25,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         Tracker.shared.setSafeCount(Safe.count)
         Tracker.shared.setNumKeysImported(SafeTransactionSigner.numberOfKeysImported())
 
+        Safe.updateCachedNames()
```
DmitryBespalov, line 28: Does it increase app loading time when launched on the device?

Multisig/Logic/Models/Safe.swift
```diff
@@ -47,6 +49,17 @@ extension Safe: Identifiable {
         return (try? context.fetch(Safe.fetchRequest().all())) ?? []
     }
 
+    static func updateCachedNames() {
+        guard let safes = try? Safe.getAll() else { return }
+        cachedNames = safes.reduce(into: [AddressString: String]()) {
+            $0[AddressString($1.address!)!] = $1.name!
```
DmitryBespalov: it is hard to read, can we have some names please? also, so many exclamation marks 😄

Multisig/Logic/Models/Safe.swift
```diff
@@ -159,6 +176,8 @@ extension Safe: Identifiable {
         if let addressString = deletedSafeAddress, let address = Address(addressString) {
             App.shared.notificationHandler.safeRemoved(address: address)
         }
+
+        updateCachedNames()
```
DmitryBespalov, line 180: interesting to see the effect on the device - if it makes these operations slower

Multisig/UI/Transaction/TransactionListViewController/Cells/TransactionListTableViewCell.swift
```diff
@@ -35,6 +35,12 @@ class TransactionListTableViewCell: SwiftUITableViewCell {
         confirmationsCountLabel.setStyle(.footnote2)
     }
 
+    override func prepareForReuse() {
+        super.prepareForReuse()
+        typeImageView.layer.cornerRadius = 0
+        typeImageView.clipsToBounds = false
```
DmitryBespalov, line 41: I'd prefer to have the image processed - please see the UIImageViewExtension.swift line 20 for example

The same is relevant for the line 59 of this file as well

Multisig/App/AppDelegate.swift
sche: I think not much. I could do it on background thread as well

Multisig/Logic/Models/Safe.swift
sche: I will add the names, but explanation marks will stay, ok? :)

Multisig/Logic/Models/Safe.swift
sche: did not notice any slowdown, but could do it on a background thread

Multisig/UI/Transaction/TransactionListViewController/Cells/TransactionListTableViewCell.swift
sche: I tried to use RoundCornerImageProcessor, but it does not work for placeholder image. So the option was to round UIImage that is passed to placeholder, or to round UIImageView. I selected the second option.
Or do you mean smth different?

Multisig/App/AppDelegate.swift
sche: But I remember I had some issues when working with db queries on different threads. This might require a bit more effort.

Multisig/Logic/Models/Safe.swift
```diff
@@ -10,6 +10,8 @@ import Foundation
 import CoreData
 import FirebaseAnalytics
 
+fileprivate var cachedNames = [AddressString: String]()
```
MouazAlzahabi, line 13: I prefer to have structure that reflect address info returned from backend, maybe later we have more info for the address for example image.

sche: For now we have only safe name. Icon is always an identicon.

Multisig/UI/Transaction/TransactionListViewController/Cells/TransactionListTableViewCell.swift
DmitryBespalov: oh I see

Multisig/Logic/Models/Safe.swift
DmitryBespalov: ok

Multisig/Logic/Models/Safe.swift
DmitryBespalov: ok

Multisig/App/AppDelegate.swift
DmitryBespalov: afaik our core data must be done on the main thread :/

sche: At the moment yes - otherwise the app will crash as we use one context. But there is a possibility to work on different threads. I created #915 issue

PR #920, opened 2021-02-10
Gh 797 known addresses in txdetails

sche:
I discarded last PR and created this one for both tx list and tx details.


Multisig/Logic/Models/Safe.swift
```diff
@@ -47,6 +49,17 @@ extension Safe: Identifiable {
         return (try? context.fetch(Safe.fetchRequest().all())) ?? []
     }
 
+    static func updateCachedNames() {
```
DmitryBespalov, line 52: question: do we really need to cache names from the core data? Maybe using core data directly is not a problem?

Can you explain why the cached names are needed?

Multisig/UI/UI Library/Views/AddressInfoView.swift
```diff
@@ -45,16 +45,11 @@ class AddressInfoView: UINibView {
         titleLabel.isHidden = text == nil
     }
 
-    func setAddress(_ address: Address, label: String?, showIdenticon: Bool = true) {
+    func setAddress(_ address: Address, label: String?, imageUri: URL?, showIdenticon: Bool = true) {
```
MouazAlzahabi: how about set the default value for imageURI = nil instead of passing it in many places

MouazAlzahabi (approved):

Multisig/UI/Transaction/ActionDetailAddressCell.swift
```diff
@@ -11,6 +11,6 @@ import UIKit
 class ActionDetailAddressCell: ActionDetailTableViewCell {
     @IBOutlet private weak var addressView: AddressInfoView!
     func setAddress(_ value: Address) {
-        addressView.setAddress(value, label: nil)
+        addressView.setAddress(value, label: nil, imageUri: nil)
```
DmitryBespalov: agree with Mouaz, default value = nil (also for label) will make it nicer

DmitryBespalov:
@sche issues that I found:

- Tx List Queue still has square logos (liltestsafe.eth)
- Tx Details has square logos (liltestsafe.eth)
- Tx List with long names (MultiSend) is making the action ellipsized - I forgot if we decided to ellipsize action or the name

Multisig/Logic/Models/Safe.swift
sche: It was because you mentioned that we can cache them, and because it is easy to do so. Probably working directly with database will not be an issue at all. But I planned to refactor it to AddressBook service anyways once we have multiple signing keys with names, safe names and etc.

Multisig/UI/UI Library/Views/AddressInfoView.swift
sche: done

Multisig/UI/Transaction/ActionDetailAddressCell.swift
sche: done

sche:
> @sche issues that I found:
> 
> * Tx List Queue still has square logos (liltestsafe.eth)
> * Tx Details has square logos (liltestsafe.eth)
> * Tx List with long names (MultiSend) is making the action ellipsized - I forgot if we decided to ellipsize action or the name

- All logos are round now
- Actions are not ellipsized anymore

PR #953, opened 2021-02-19
Gh 929/submit rejection transaction

MouazAlzahabi:


Multisig/Data/Services/Safe Client Gateway Service/ProposeTransactionRequest.swift
```diff
@@ -0,0 +1,68 @@
...
+//  RejectTransactionRequest.swift
+//  Multisig
+//
+//  Created by Moaaz on 2/17/21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct ProposeTransactionRequest: JSONRequest {
```
sche, line 11: We already have a Transaction object. I would suggest to re-use it.

I implemented similar request for safe transaction service in wc-prototype branch. You can check it here https://github.com/gnosis/safe-ios/blob/wc-prototype/Multisig/Data/Services/Safe%20Transaction%20Service/CreateTransactionRequest.swift

DmitryBespalov (approved):
I think Andrey's comment needs to be fixed, but otherwise ok!

Multisig/Data/Services/Safe Client Gateway Service/ProposeTransactionRequest.swift
MouazAlzahabi: Thanks for such nice catch, in order to achieve it I had to confirm Codable protocol for couple of entities

sche (approved):

PR #956, opened 2021-02-22
Gh 943/passcode UI

DmitryBespalov:
- New passcode view controller that is base for all passcode-related views
- Create passcode
- Repeat passcode
- Trigger passcode from the Import Owner Key completion


Multisig/UI/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -0,0 +1,144 @@
...
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 2/22/21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class PasscodeViewController: UIViewController, UITextFieldDelegate {
+    @IBOutlet weak var scrollView: UIScrollView!
```
MouazAlzahabi, line 12: Don't we need all of those outlets to be private?

MouazAlzahabi (approved):

Multisig/UI/Settings/Passcode/PasscodeViewController.swift
DmitryBespalov: I wanted them to be accessible in the subclasses, so didn't add private

PR #963, opened 2021-02-24
Gh 962 app password tracking

sche:


Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
```diff
@@ -200,7 +200,9 @@ class BalancesViewController: LoadableViewController, UITableViewDelegate, UITab
             updateSection(indexPath.section)
         }
         cell.onImport = { [unowned self] in
-            let vc = CreatePasscodeViewController { [weak self] in
+            AppSettings.passcodeBannerDismissed = true
```
DmitryBespalov, line 203: I think this is wrong, as per spec the banner should be dismissed when user imports the key

Multisig/UI/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -225,7 +257,18 @@ class EnterPasscodeViewController: PasscodeViewController {
     }
 }
 
-class ChangePasscodeViewController: PasscodeViewController {
+class ChangePasscodeViewController: EnterPasscodeViewController {
```
DmitryBespalov: instead of subclassing it is better to have a variable that is configured at the call site - for the title and for the tracking event. This way we don't have too many classes (it becomes confusing)

Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
sche: Sorry if the spec confusing in this place. Here how it is stated:
```
User is reminded to set up a passcode with a banner in coins assets screen if all the following cases apply:
user has imported signing key
passcode was not set
banner was not closed / skipped
```
Starting setting up and not finishing it is "skipped" here. So the logic is the same as with "import signing key" banner.

DmitryBespalov: I see.

DmitryBespalov (approved):

Multisig/UI/Settings/Passcode/PasscodeViewController.swift
sche: changed

PR #998, opened 2021-03-10
Gh 935/owner key list

DmitryBespalov:


Multisig/Logic/PrivateKeyController.swift
```diff
@@ -39,6 +39,28 @@ class PrivateKeyController {
...
+        } catch {
+            App.shared.snackbar.show(
+                error: GSError.error(description: "Failed to remove imported key", error: error))
+        }
+    }
+
+    static func edit(keyInfo: KeyInfo, name: String) {
+        keyInfo.rename(newName: name)
+        App.shared.snackbar.show(message: "Owner key updated")
+        NotificationCenter.default.post(name: .ownerKeyRemoved, object: nil)
```
DmitryBespalov: why is this notification that the key removed? It was renamed, I think we need a new event

Multisig/Logic/PrivateKeyController.swift
```diff
@@ -39,6 +39,28 @@ class PrivateKeyController {
...
+        //Tracker.shared.track(event: TrackingEvent.ownerKeyRemoved)
```
DmitryBespalov: this one not needed, let's delete this line

Multisig/UI/Settings/OwnerKeyManagement/OwnerKeysListViewController/OwnerKeysListTableViewCell.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import UIKit
+
+protocol OwnerKeysListTableViewCellDelegate {
+    func ownerKeysListTableViewDidEdit(cell: OwnerKeysListTableViewCell)
+    func ownerKeysListTableViewCellDidRemove(cell: OwnerKeysListTableViewCell)
+}
+class OwnerKeysListTableViewCell: UITableViewCell {
+    @IBOutlet private weak var addressInfoView: AddressInfoView!
+    var delegate: OwnerKeysListTableViewCellDelegate?
```
DmitryBespalov: delegate must be weak, otherwise we'll have a retain cycle.

To enable this, you need to mark protocol as class:

```
protocol OwnerKeysListTableViewCellDelegate: class {
```

Multisig/UI/Settings/OwnerKeyManagement/OwnerKeysListViewController/OwnerKeysListViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        present(vc, animated: true)
+    }
+
+    override func reloadData() {
+        super.reloadData()
+        keys = (try? KeyInfo.all()) ?? []
+        setNeedsReload(false)
+        onSuccess()
+        tableView.reloadData()
+
```
DmitryBespalov: blank line, not needed

Multisig/UI/Settings/OwnerKeyManagement/OwnerKeysListViewController/OwnerKeysListViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        keys.count
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueCell(OwnerKeysListTableViewCell.self, for: indexPath)
+        let keyInfo = keys[indexPath.row]
+        cell.set(address: keyInfo.address!, title: keyInfo.name ?? "key")
```
DmitryBespalov: Let's make a default title like

```
"Key \(keyInfo.address!.ellipsized())"
```

DmitryBespalov: potentially, the address might be nil if the database stores trash string that is not an address.

Can we handle that case somehow somewhere?

DmitryBespalov: You know what... maybe we should encapsulate this default name logic into a new property - "displayName" on the KeyInfo

Issue #1015, opened 2021-03-15
[iOS] Support fiat currencies 


MouazAlzahabi:
We need to support fiat currencies as follows:

- [x] Create fiat currencies list
- [x] Update app settings screen to add select currency row 
- [x] Update balances screen based on selected fiat

liliya-soroka:
Tracking verification is required to close the task

liliya-soroka:
Verified
app version 2.13.0 (355)

PR #1021, opened 2021-03-17
GH-1015 Fix issue

MouazAlzahabi:


Multisig/Logic/Models/AppSettings.swift
```diff
@@ -39,6 +39,18 @@ extension AppSettings {
     @AppSetting(\.displayMode)
     static var displayMode: Int32
 
+    @AppSetting(\.fiatCode)
+    private static var fiatCode: String?
```
DmitryBespalov, line 43: what is the difference between fiat code and selected fiat code? I see there's a default value, but maybe we could extend the property wrapper to accept a default value and not have 2 properties for the same thing?

Multisig/Logic/Models/TokenBalance.swift
```diff
@@ -18,29 +18,29 @@ struct TokenBalance: Identifiable, Hashable {
...
 extension TokenBalance {
-    init(_ response: SafeBalancesRequest.Response) {
+    init(_ response: SafeBalancesRequest.Response, code: String) {
         self.init(address: response.tokenAddress?.address ?? .ether,
                   symbol: response.token?.symbol,
                   logoUri: response.token?.logoUri,
                   tokenBalance: response.balance,
                   decimals: response.token?.decimals,
-                  fiatBalance: response.balanceUsd)
+                  fiatBalance: response.balanceUsd, code: code)
```
DmitryBespalov: code should be on the next line

Multisig/Logic/Models/TokenBalance.swift
```diff
@@ -18,29 +18,29 @@ struct TokenBalance: Identifiable, Hashable {
...
 
-    init(_ item: SCGBalance) {
+    init(_ item: SCGBalance, code: String) {
         self.init(address: item.tokenInfo.address.address,
                   symbol: item.tokenInfo.symbol,
                   logoUri: item.tokenInfo.logoUri,
                   tokenBalance: item.balance,
                   decimals: item.tokenInfo.decimals,
-                  fiatBalance: item.fiatBalance)
+                  fiatBalance: item.fiatBalance, code: code)
```
DmitryBespalov: code should be on the next line

DmitryBespalov (approved):

Multisig/Logic/Models/AppSettings.swift
MouazAlzahabi: I couldn't updated it now, it seems this will need refactor for were we used it

Issue #800, opened 2020-12-21
[iOS] Known Addresses V2: Safe Settings screens


sche:
Known addresses are displayed for:
- Safe owner address
- Imported owner key (on web it comes from address book, on mobile imported owner key has a default name ‘Imported owner key’, and a custom name once “Multiple keys” feature is implemented)
- Safe version
- Fallback handler
- Enabled modules

Zeplin
https://zpl.io/VxZl9v3, https://zpl.io/2pGBD7Q

liliya-soroka:
Reopened:
the wrong design links are used in the task description

The correct link for advanced settings - Known addresses:
https://app.zeplin.io/project/5e8f27b9963e55bb79b13634/screen/6006e483d7aa7f96034fe79d

The task is reopened:
1. Fallback handler:
- Default fallbackhendler is not taken from the known addresses list
to check: staging, mainnet
safe= 0xB260dD872c57205C9f308ddbd1C3B038c5c69a5F
- Instead of "Not set"  Unknown 0x00000 fallbackhandler is used for safes with safe version = 1.0.0
staging, mainnet, safe=alice.eth
2. Modules logo is not taken from known addresses list 
to check: staging, rinkeby , safe=liltestsafe.eth

app version 2.13.0 (355)

liliya-soroka:
Verified
app version 2.13.0 (360)

PR #1022, opened 2021-03-17
gh-800 Added address info from backend

DmitryBespalov:


Multisig/UI/UI Library/Views/AddressInfoView.swift
```diff
@@ -29,6 +29,7 @@ class AddressInfoView: UINibView {
         titleLabel.setStyle(.headline)
         textLabel.setStyle(.headline)
         addressLabel.setStyle(.tertiary)
+        textLabel.numberOfLines = 1
```
MouazAlzahabi: Can you set this in interface builder

MouazAlzahabi (approved):

Issue #1016, opened 2021-03-16
[iOS] Changes to Passcode settings


DmitryBespalov:
Passcode settings should allow:

### Turn on/off biometry 
  - **ON**: user can enter passcode using a biometry (Touch ID/Face ID)
  - **OFF**: biometry is not used to enter passcode. 
  - If user did not authorize biometry before, system "permission" alert is shown.
  - If user denied the biometry use, then show alert that leads to device settings.
### Turn on/off require to open app
  - **ON**: passcode is asked after every time user leaves the app and comes back.
  - **OFF**: passcode is not asked on app opening
### Turn on/off require to confirm transactions
  - **ON**: passcode is asked to confirm transaction and reject transaction
  - **OFF**: passcode is not asked to confirm or reject transaction.
### Other
- If only one of the “require” switches is ON and user disables it, then it is equivalent to disabling the “Use passcode” switch.
- If the device does not support biometrics, then the “Login with biometrics” is hidden.
- If passcode is enabled, then changing any switch requires entering the current passcode.

[Spec section link](https://docs.google.com/document/d/1NegSN4t1dsysUewY_K7v3FLi_sORpS-QpfiTAUZkrn8/edit#bookmark=id.ri0dpviez61p)

liliya-soroka:
The issues:
1. https://app.zenhub.com/workspaces/safe-multisig---mobile-5c1b98250e13551e8aae81eb/issues/gnosis/safe-ios/1069

liliya-soroka:
Verified
app version 2.14.0 (4)

PR #1063, opened 2021-03-31
gh-1016 Added functionality for the settings

DmitryBespalov:
- Biometry enable/disable
- App login on/off
- Confirmation on/off


Multisig/Logic/AuthenticationController.swift
```diff
@@ -102,6 +109,29 @@ class AuthenticationController {
         }
         return Data(derivedKey).toHexString()
     }
+
+
+    // MARK: - Biometry
+
+    /// Is device hardware supports the biometry
+    var isBiometricsSupported: Bool {
```
sche, line 117: I would be for adding it to the lib, but okay :)

Multisig/UI/Settings/Passcode/PasscodeSettingsViewController.swift
```diff
@@ -80,42 +86,105 @@ class PasscodeSettingsViewController: UITableViewController {
...
         }
-
-        present(nav, animated: true, completion: nil)
     }
 
-    private func changePasscode() {
+    private func withPasscodeAuthentication(
+        for reason: String,
+        tracking: TrackingEvent? = nil,
+        authenticated: @escaping (Bool, UINavigationController?, @escaping () -> Void) -> Void
```
sche: omg. This is a bit confusing

Multisig/UI/Settings/Passcode/PasscodeSettingsViewController.swift
```diff
@@ -80,42 +86,105 @@ class PasscodeSettingsViewController: UITableViewController {
...
+                }
+                nav?.pushViewController(changeVC, animated: true)
+            } else {
+                finish()
+            }
+        }
+    }
+
+    private func toggleUsage(option: PasscodeOptions) {
+        withPasscodeAuthentication(for: "Require to open app") { [unowned self] success, _, finish in
```
sche: should this string depend on the option?

Multisig/UI/Settings/Passcode/PasscodeSettingsViewController.swift
```diff
@@ -80,42 +86,105 @@ class PasscodeSettingsViewController: UITableViewController {
...
+            }
+
+            finish()
+        }
+    }
+
+    private func toggleBiometrics() {
+        withPasscodeAuthentication(for: "Login with biometrics") { success, nav, finish in
+            if AppSettings.passcodeOptions.contains(.useBiometry) {
+                AppSettings.passcodeOptions.remove(.useBiometry)
```
sche, line 148: I remember we discussed it some time ago, but I have not found it in Passcode v2 requirements is that if action does not reduce security, it should not require confirmation with passcode. Disabling biometry actually increases security, so I would suggest not to ask confirmation for this case. I will add it to the refinement meeting agenda to discuss

Multisig/UI/Settings/Passcode/PasscodeSettingsViewController.swift
DmitryBespalov: yes, good catch

Multisig/UI/Settings/Passcode/PasscodeSettingsViewController.swift
DmitryBespalov: weren't you saying before that we should request biometry in any case when entering passcode?

Or you're saying that we should not request passcode for disabling biometry?

sche:
Bug:
1. Try to enable biometry.
2. Close popover window
=> System suggestion to enable biometry appears

Multisig/UI/Settings/Passcode/PasscodeSettingsViewController.swift
sche: 2nd

sche:
Bug: Disabling biometry in Phone settings for the app does not influence the switch
Steps:
1. Enable biometry
2. Go to phone settings and select the app
3. Disable biometry in global settings for the app
4. Navigate to app's passcode settings
Expected result: Login with biometrics option should be disabled

Multisig/Logic/AuthenticationController.swift
DmitryBespalov: will move it after this is merged

Multisig/UI/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -57,6 +59,9 @@ class PasscodeViewController: UIViewController, UITextFieldDelegate {
         trackEvent(.userPasscodeSkipped)
     }
 
+    @IBAction func didTapBiometry(_ sender: Any) {
```
sche, line 62: I would add a comment here that it is meant for override

sche (approved):

PR #1065, opened 2021-04-01
Gh 1019 ask biometry after passcode setup

sche:


Multisig/UI/App/ViewControllerFactory.swift
```diff
@@ -18,15 +18,29 @@ enum ViewControllerFactory {
...
             .environment(\.managedObjectContext, App.shared.coreDataStack.viewContext)
-            let vc = UIHostingController(rootView: start)
-            vc.modalPresentationStyle = .fullScreen
-            vc.modalTransitionStyle = .crossDissolve
+
+            let startVC = UIHostingController(rootView: start)
+            let createPasswordVC = CreatePasscodeViewController {
+                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(100)) {
+                    if let activated = try? App.shared.auth.activateBiometry(), activated {
+                        AppSettings.passcodeOptions.insert(.useBiometry)
```
DmitryBespalov: I think this is a wrong place to do this. I would rather put this kind of thing into the CreatePasscodeVC flow itself on the last screen after passcode is successful - this way the biometry is activated every time the passcode is created.

sche: Initially I did it like that, but then
1) if you activated it before, it shows face id, checks it -  and nothing happens, so it looks like a glitch, but actually it is not
2) later you can always activate it in passcode settings
So I decided to offer it only after terms, so UX wise it looks better.

DmitryBespalov: what about other places?
- After importing a key
- After opening a banner

I think there should always be an option to use biometry after the passcode setup. If the biometry failed or cancelled, then we should not set this setting, otherwise we should set it. 

Somehow  I think it is also a bad ux not to use biometry automatically after passcode is created - in any place where it can be created. 

So maybe a solution would be at the end of the passcode creation, check if the biometry is "not yet permitted", and if so, then request it so that user authorizes it. 

Otherwise, if it is permitted, then set the useBiometry to yes?

sche: Ok. I played around with it a bit, and it seems fine even without "not yet permitted" check.

DmitryBespalov (approved):

MouazAlzahabi (approved):

PR #1067, opened 2021-04-02
Unstoppabledomains/resolution

JohnnyJumper:
This PR adds Unstoppable domains integration into the flow of adding new safe addresses. 

Key points:
* Updates the UI from ENS only to the general "blockchain domain" 
* Uses previous ENS integration in the case when a domain is not ending on .crypto or .zil
* Uses UnstoppableDomainResolution to fetch the eth address on a stored domain.


DmitryBespalov:
Hi @JohnnyJumper thanks for opening a pr!

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,70 @@
...
+import UnstoppableDomainsResolution
+
+class BlockchainDomainManager {
+    
+    let ens: ENS;
+    let resolution: Resolution;
+    
+    init() {
+        ens = ENS(registryAddress: "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e");
+        self.resolution = try! Resolution();
```
DmitryBespalov: is this guaranteed not to throw?

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,70 @@
...
+        
+        let dispatchGroup = DispatchGroup();
+        dispatchGroup.enter();
+        resolution.addr(domain: domain, ticker: "eth") { result in
+            switch result {
+                case .success(let returnValue):
+                    address = returnValue;
+                case .failure(let error):
+                  err = error;
+          }
```
DmitryBespalov: the indentation seem to be a bit off

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,70 @@
...
+        };
+        dispatchGroup.wait();
+        
+        guard err == nil else {
+            if (err is ResolutionError) {
+                throw self.throwCorrectUdError(err as! ResolutionError, domain);
+            }
+            throw err!;
+        }
+        return Address(address)!;
```
DmitryBespalov: we could use `Address(exactly: address)` here to remove force unwrap

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,70 @@
...
+                case .success(let returnValue):
+                    address = returnValue;
+                case .failure(let error):
+                  err = error;
+          }
+            dispatchGroup.leave();
+        };
+        dispatchGroup.wait();
+        
+        guard err == nil else {
```
DmitryBespalov: I think having 2 if-lets would be more readable here:

```swift
if let error = err as? ResolutionError {
  throw self.throwCorrectUdError(error, domain)
} else if let error = err {
  throw error
}
```

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,70 @@
...
+    let resolution: Resolution;
+    
+    init() {
+        ens = ENS(registryAddress: "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e");
+        self.resolution = try! Resolution();
+    }
+    
+    func resolveUD(_ domain: String) throws -> Address {
+        var address: String = "";
+        var err: Error? = nil;
```
DmitryBespalov: Let's rename `err` to `resolutionError`

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
DmitryBespalov: however, if the `address` is not guaranteed to be an address String, there's an option to use `try Address(from: address)`

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,70 @@
...
+            throw err!;
+        }
+        return Address(address)!;
+    }
+    
+    func resolve(domain: String) throws -> Address {
+        return domain.isUDdomain() ? try self.resolveUD(domain) : try ens.address(for: domain);
+    }
+    
+    func throwCorrectUdError(_ err: ResolutionError, _ domain: String) -> DetailedLocalizedError {
```
DmitryBespalov: Please rename `err` to `error` to use a full name

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
DmitryBespalov: Alternative is to use `try Address(from: address)` if there's a possibility that the address string is not an Ethereum address

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
JohnnyJumper: @DmitryBespalov, As long as we are using default settings this won't throw, this, however, does limit the app to the mainnet only. 
I can reconfigure the library to use testnet as well for future development. We support Rinkeby for .crypto. 

Where can I get network configurations such as chainId and blockchain provider (I assume it is going to be infura as in gnosis-android) to do so?

DmitryBespalov: Got it, yes, that would be good to have because we have versions of the app working on Rinkeby network (in the future also on other chains). 

To get the blockchain provider url: `App.configuration.services.ethereumServiceURL`

To get the network name (we don't have chainId specified) - `App.configuration.app.network`

JohnnyJumper: Awesome, I have connected the library to the potential testnets and watched the error in case if network supplied is not supported by UD.

DmitryBespalov:
@JohnnyJumper ok, so far it looks good to me. I'll ping other devs for review. 

One other thing though is that we're about to transition from CocoaPods to SwiftPM, does UD support it? (I see no Package.swift there).

JohnnyJumper:
> @JohnnyJumper ok, so far it looks good to me. I'll ping other devs for review.
> 
> One other thing though is that we're about to transition from CocoaPods to SwiftPM, does UD support it? (I see no Package.swift there).

Should be relatively easy to update by adding this: 
```swift
package.dependencies.append(
    .package(url: "https://github.com/unstoppabledomains/resolution-swift", from: "0.3.5")
)
```
When I launched the project via xcode the first time I didn't notice any other packages in SwiftPM and went with cocoapods.

DmitryBespalov:
Great! Yes, it's in progress in the current sprint (I'll work on it next) https://github.com/gnosis/safe-ios/issues/1041

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,82 @@
...
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import UnstoppableDomainsResolution
+
+class BlockchainDomainManager {
+    
+    let ens: ENS;
+    let resolution: Resolution?;
```
sche: We don't use ';'. Could you please remove it everywhere in this file?

Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -0,0 +1,82 @@
...
+                cns: NamingServiceConfig(
+                    providerUrl: App.configuration.services.ethereumServiceURL.absoluteString,
+                    network: App.configuration.app.network.rawValue.lowercased()
+                )
+            )
+        );
+    }
+    
+    func resolveUD(_ domain: String) throws -> Address {
+        guard resolution != nil else {
```
sche: guard let resolution = resolution else { ... }
No need to use force unwrapping below.

sche (changes requested):
Lgtm after fixing minor things.
The major question here is alignment with Android app. Here the proposal is to use "Enter Blockchain domain", but on Android it seems a split of ENS/UD when selecting how to enter address.
We will clarify with product how to keep it and one of the platform realisations would need to change it.

JohnnyJumper:
> Lgtm after fixing minor things.
> The major question here is alignment with Android app. Here the proposal is to use "Enter Blockchain domain", but on Android it seems a split of ENS/UD when selecting how to enter address.
> We will clarify with product how to keep it and one of the platform realisations would need to change it.

That is true. It slipped from my mind that both apps should have the same implementation. Let me know which one should I fix.

sche:
@tschubotz what option would you prefer?

tschubotz:
> @tschubotz what option would you prefer?

I'd prefer the Android way, i.e. keeping both separate. That way everything is just more explicit and it names the supported type (ENS, UD) instead of "Blockchain domains" which could mean many things :)

JohnnyJumper:
@sche  @tschubotz I split the ens and ud into separate views, also updated the library to the latest version, and changed a few error messages to make them more explicit. Can you take a look?

DmitryBespalov:
@JohnnyJumper we've also moved from CocoaPods to SwiftPM, can you please update this PR accordingly?

Multisig/UI/Safe Management/Add Safe/EnterENSNameViewController.swift
```diff
@@ -28,13 +28,13 @@ class EnterENSNameViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        navigationItem.title = "Enter ENS Name"
+        navigationItem.title = "Enter ENS domain"
```
DmitryBespalov: I would keep the wording as it was before

Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
```diff
@@ -78,7 +78,7 @@ class EnterSafeAddressViewController: UIViewController {
             self?.present(vc, animated: true, completion: nil)
         }))
 
-        vc.addAction(UIAlertAction(title: "Enter ENS Name", style: .default, handler: { [weak self] _ in
+        vc.addAction(UIAlertAction(title: "Enter ENS Domain", style: .default, handler: { [weak self] _ in
```
DmitryBespalov: here too, I'd keep the existing wording intact

Multisig/UI/Safe Management/Add Safe/EnterENSNameViewController.swift
JohnnyJumper: Done

JohnnyJumper:
@DmitryBespalov 
Merged the main branch into here, removed all cocoapods, and added ud library to SwiftPM
Also updated the text from domain to name in both ens and ud version of UI

Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -278,10 +278,10 @@ enum GSError {
         let loggable = false
     }
 
-    struct ENSAddressNotFound: DetailedLocalizedError {
+    struct BlockhainAddressNotFound: DetailedLocalizedError {
```
DmitryBespalov: sorry, one more thing that I overlooked - let's revert this struct (and `howToFix` below, too)

DmitryBespalov (approved):
one minor fix left, otherwise looks good!

Multisig/Data/Services/Utils/GSError.swift
JohnnyJumper: What a great catch! I completely forgot to rename it back. thank you!

Multisig.xcodeproj/project.pbxproj
```diff
@@ -353,6 +353,13 @@
...
 		0AFA72A024BF368900A370EE /* SlicedText+Address.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AFA729F24BF368900A370EE /* SlicedText+Address.swift */; };
 		0AFC33ED25DE9C3200067D41 /* AuthenticationControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AFC33EC25DE9C3200067D41 /* AuthenticationControllerTests.swift */; };
+		3E373F6F263155CD00D8107A /* UnstoppableDomainsResolution in Frameworks */ = {isa = PBXBuildFile; productRef = 3E373F6E263155CD00D8107A /* UnstoppableDomainsResolution */; };
+		3E373F762631781A00D8107A /* EnterUnstoppableNameViewController.xib in Resources */ = {isa = PBXBuildFile; fileRef = 3E373F742631781900D8107A /* EnterUnstoppableNameViewController.xib */; };
+		3E373F772631781A00D8107A /* EnterUnstoppableNameViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E373F752631781A00D8107A /* EnterUnstoppableNameViewController.swift */; };
+		3E373F7D2631784200D8107A /* BlockchainDomainManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E373F7C2631784200D8107A /* BlockchainDomainManager.swift */; };
+		3E59F14D25B9667F00BB2788 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
sche: (null) looks like smth broken

Multisig.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -33,44 +33,53 @@
...
       },
       {
         "package": "CryptoSwift",
         "repositoryURL": "https://github.com/krzyzanowskim/CryptoSwift.git",
         "state": {
           "branch": null,
-          "revision": "79ac632c85df81f2618dd838fc3869633e163370",
-          "version": "1.2.0"
+          "revision": "3a2acbb32ab68215ee1596ee6004da8e90c3721b",
+          "version": "1.0.0"
```
sche: why was it downgraded? Could you please update in your package `.package(url: "https://github.com/krzyzanowskim/CryptoSwift.git", from: "1.0.0")` to 1.4.0?

sche: probably it will be properly resolved without the update too

Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -313,6 +313,43 @@ enum GSError {
         let code = 1110
         let loggable = false
     }
+    
+    struct UDUnsuportedName: DetailedLocalizedError {
+        let description = "Can't use this name"
+        let reason = "Unstoppable name is invalid."
+        let howToFix = "Name should end with .crypto or .zil"
+        let domain = clientErrorDomain
+        let code = 1120
```
sche: Error codes should be aligned between iOS and Android. I did not check, are they?

Multisig/UI/Safe Management/Add Safe/EnterUnstoppableNameViewController.swift
```diff
@@ -0,0 +1,130 @@
...
+        textField.textField.delegate = self
+        textField.textField.becomeFirstResponder()
+
+        addressFoundStackView.isHidden = true
+        foundHeaderLabel.setStyle(.primary)
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        trackEvent(.safeAddEns)
```
sche: We need to add `.safeAddUD` event. And also please assure that it is the same on Android.

sche:
Looks good! 👍🏻 
Please resolve conflicts and align error messages & tracking event for both iOS and Android and we are good to go.

Multisig.xcworkspace/xcshareddata/swiftpm/Package.resolved
JohnnyJumper: At the moment we can't update the CryptoSwift package.

```Because no versions of EthereumAddress match the requirement 1.3.1..<2.0.0 and EthereumAddress 1.3.0 depends 
on CryptoSwift 1.0.0..<1.1.0, EthereumAddress >=1.3.0 requires CryptoSwift 1.0.0..<1.1.0.
And because root depends on CryptoSwift 1.4.0..<2.0.0 and root depends on EthereumAddress 1.3.0..<2.0.0, 
version solving failed.
```

As I understood this doesn't cause an issue for the gnosis project.  We will work on the update soon

JohnnyJumper:
@sche  There is something strange happening when I resolve the conflicts. Solving the conflicts triggered codecov report above saying that the coverage dropped. I believe I made some cruel mistakes in merging as the files are mostly xcode generated. Can you help me out here? 

P.S once I reverted the merge commit codecov edited its report saying that now the coverage drop is 0.40 instead of 63 something %. I am not sure why this happens.

Other than that should be good now. Errors are aligned with Android version and I removed (null) lines from project.pbxproj

sche:
Thanks @JohnnyJumper 
There is still a conflict. Probably because we make quite some merges currently. Could you please fix conflicts next Tuesday (this will be a planning day for us with probably no merges), and we will merge them?

sche:
@JohnnyJumper do you plan to resolve conflicts this week?

JohnnyJumper:
@sche After resolving the conflicts and updating to the gnosis:main branch I start to receive a firebase error on local build. Probably some missing config causes that. 
To be specific here is the error.
Screen Shot 2021-05-12 at 4 09 21 AM

DmitryBespalov:
@JohnnyJumper @sche I think this is ok to merge, we'll fix that thing on our end. @MouazAlzahabi it'd be good to create an improvement to not use remote config if the Firebase config files are not available (I think it is already implemented in our code for the Firebase Analytics)

sche (approved):

DmitryBespalov:
This will be released in approx. 3 weeks

Issue #1017, opened 2021-03-16
[iOS] Changes to "Enter passcode" screen for biometrics


DmitryBespalov:
### When to show the screen
- Changing passcode 
- Changing any switch in the passcode settings when the passcode is enabled.
- Confirming, rejecting (or in the future, submitting) transaction if "require for confirmations" switch is ON
- When app enters foreground if "require to open app" switch is ON

### Functionality
- “Forgot your passcode?” button will delete all app data.

**Touch ID/Face ID**
- Touch ID/Face ID button will request the biometry login instead of entering passcode.
- Touch ID/Face ID button may be hidden due to operating system reasons (wrong entry of Touch ID 3 times, or other reason). In this case, the button will not be shown.
- Touch ID / Face ID button must be hidden if the "Login with biometrics" switch is OFF.
- The button is only for requesting TouchID/FaceID, it doesn't change or navigate to the settings.

[Spec link](https://docs.google.com/document/d/1NegSN4t1dsysUewY_K7v3FLi_sORpS-QpfiTAUZkrn8/edit#bookmark=id.jjvcy2jl69dz)

liliya-soroka:
Verified
app version 2.14.0 (4)

PR #1068, opened 2021-04-06
gh-1017 Added passcode on app login

DmitryBespalov:


Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -33,6 +34,13 @@ class MainTabBarViewController: UITabBarController {
             object: nil)
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        guard appearsFirstTime else { return }
+        appearsFirstTime = false
```
sche, line 40: Is it needed because of the lock screen?

sche (approved):

Multisig/UI/App/MainTabBarViewController.swift
DmitryBespalov: yes, because otherwise rate app appears on lock screen

PR #1072, opened 2021-04-07
Gh 1069/fix bug passcode

DmitryBespalov:


sche (approved):

Multisig/App/SceneDelegate.swift
```diff
@@ -58,35 +82,30 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         // Called when the scene has moved from an inactive state to an active state.
         // Use this method to restart any tasks that were paused (or not yet started) when the scene was inactive.
         App.shared.clientGatewayHostObserver.startObserving()
-        hidePrivacyProtectionWindow()
+
+        if windowState == .privacy {
```
MouazAlzahabi, line 86: Is it possible to create didSet event for windowState property, which inside we decide what to show depending on the new value?

MouazAlzahabi (approved):

Multisig/App/SceneDelegate.swift
DmitryBespalov: it is possible, but I'd keep it as is for now as it's Thursday already

PR #1077, opened 2021-04-08
Gh 1041 swift pm

sche:


Multisig/Logic/Models/TokenBalance.swift
```diff
@@ -8,6 +8,7 @@
 
 import Foundation
 import SwiftCryptoTokenFormatter
+import UIKit
```
MouazAlzahabi, line 11: Is it needed ?

Multisig/Logic/Models/Address.swift
```diff
@@ -7,6 +7,8 @@
 //
 
 import Web3
+import Foundation
```
MouazAlzahabi, line 10: Is it needed ?

MouazAlzahabi (approved):

DmitryBespalov:
Oh Yeah! 👍 

Screenshot 2021-04-08 at 17 13 29

DmitryBespalov (approved):

Multisig/Logic/Models/Address.swift
sche: yes

Multisig/Logic/Models/TokenBalance.swift
sche: yes

Issue #1084, opened 2021-04-13
[iOS] Forced update dialog


DmitryBespalov:
### Zeplin UI

- https://zpl.io/2GDBRmd
- https://zpl.io/VQPedqn
- https://zpl.io/agxKOnO

### Firebase config variables

- `newestVersion` - newest app version. Default is null.
- `deprecatedSoon` - app versions to be deprecated soon. Comma-separated list of ranges or version numbers (mixing is possible). Each range consists of lower and upper version (inclusive). Values in range separated by minus sign “-”. Default is empty/null
  - `x1.y1.z1-x2.y2.z2,x3.y3.z3-x4.y4.z4,x5.y5.z5`
    - This encodes 2 ranges and 1 explicit version:
    - From `x1.y1.z1 `to `x2.y2.z2`
    - From `x3.y3.z3` to `x4.y4.z4`
    - Just `x5.y5.z5`
- `deprecated` - app versions that are deprecated. Comma-separate list of ranges or version numbers (see above). Default is empty / null.

### The logic for showing the alert

- **IF** app version is in the deprecated list, **THEN** show forced update
- **ELSE IF** app version is in deprecated soon list, **THEN** show the dismissable update alert
- **ELSE IF** the app version is lower (sem version semantics lower) than newest version, **THEN** show the dismissable update alert once for this app version.

### Strategy for loading the remote config parameters 

- Fetch new values in the background
- Activate on app start so that the values are applied
- On app start react on currently stored (or default) values.
- We use Firebase’s default options for updating the remote configuration (once per 12 hours)

 ### Change The Release Process Documentation

- We will have the variables documented in a new sheet in the following doc
  - Note: try to merge the ranges to make them simpler.
- When making a new release candidate, the PO should decide the status and document it in the related doc. (if to deprecate or deprecate old versions soon)
- The release person should update the parameters on the firebase console first on staging (Mainnet/Rinkey) in accordance to the documented parameters, it’s preferable to ask for review from other team members, then notify QA about the release candidate.
- The QA should test that the app is acting in accordance with the defined parameter values.
- After making sure that the app is available to be downloaded, the release person copies the same parameters from staging to release project.

MouazAlzahabi:
@tschubotz Do we need tracking for screen ID, skip, update actions ?

tschubotz:
@MouazAlzahabi just screen tracking would be enough for me right now!

I suggest:

- https://zpl.io/2GDBRmd: `update_deprecated`
[...]

PR #1108, opened 2021-04-19
GH-1084 Create AppUpdateViewController

MouazAlzahabi:


Multisig/UI/App/ViewControllerFactory.swift
```diff
@@ -17,7 +17,14 @@ enum ViewControllerFactory {
     static func rootViewController() -> UIViewController {
         let tabBarVC = MainTabBarViewController()
 
-        if !AppSettings.termsAccepted {
+        if let updateViewController = App.shared.updateController.appUpdateStatus() {
```
DmitryBespalov: it is strange to have the method named `appUpdateStatus` to return a view controller 🤪

Multisig/Logic/UpdateController.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Multisig
+//
+//  Created by Moaaz on 4/16/21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+class UpdateController {
+    func appUpdateStatus() -> UpdateAppViewController? {
```
DmitryBespalov: better name would be `makeUpdateAppViewController()`

Multisig/Logic/UpdateController.swift
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        return UpdateAppViewController(style: style)
+    }
+
+    private func check(value: String, in range: String) -> Bool {
+        let ranges = range.split(separator: ",")
+        for range in ranges {
+            let minVersion = range.components(separatedBy: "..")[0]
+            let maxVersion = range.components(separatedBy: "..")[1]
```
DmitryBespalov: Did you change the requirements for the range syntax? This implementation is not correct per requirements:

  - `x1.y1.z1-x2.y2.z2,x3.y3.z3-x4.y4.z4,x5.y5.z5`
    - This encodes 2 ranges and 1 explicit version:
    - From `x1.y1.z1 `to `x2.y2.z2`
    - From `x3.y3.z3` to `x4.y4.z4`
    - Just `x5.y5.z5`

The separator must be minus sign "-"

Also, what if there's no minus sign? The range list can be just one version number and not a lower/upper bound.

Multisig/Logic/UpdateController.swift
```diff
@@ -0,0 +1,43 @@
...
+
+            if value >= minVersion && value <= maxVersion {
```
DmitryBespalov: String comparison is not a semantic version comparison and will fail for some of the data. This can be unit tested, by the way :) 

Here's how to compare semantic versions: https://semver.org/#spec-item-11

Multisig/Cross-layer/FirebaseRemoteConfig.swift
```diff
@@ -0,0 +1,34 @@
...
+class FirebaseRemoteConfig {
+    static let shared = FirebaseRemoteConfig()
+    enum Key: String {
+        case newestVersion
+        case deprecatedSoon
+        case deprecated
+    }
+
+    private var remoteConfig: RemoteConfig!
+    private let defaultValues: [String : NSObject] = [Key.newestVersion.rawValue : App.configuration.app.marketingVersion as NSObject,
```
DmitryBespalov: this is not correct - the default newest version value must be null or empty

Multisig/Logic/UpdateController.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+
+import Foundation
+
+class UpdateController {
+    func appUpdateStatus() -> UpdateAppViewController? {
+        let remoteConfig = FirebaseRemoteConfig.shared
+        let appVersion = App.configuration.app.marketingVersion
+
+        let latestAppVersion: String? = remoteConfig.value(key: .newestVersion)
```
DmitryBespalov: If the "newest version" is not set, then we also don't need to do anything

DmitryBespalov (changes requested):
Thanks, some changes needed 🙏

Multisig/Cross-layer/FirebaseRemoteConfig.swift
MouazAlzahabi: I think it's fine to have value for it like this, but I change it anyway

DmitryBespalov (approved):

Issue #1105, opened 2021-04-16
[iOS]The user has to do cold app start if it was opened from push notification


liliya-soroka:
**Bug description**
Open app from push notification is not taken into account during "request on app start" implementation
1. Push notification opens tx details without passcode requesting
2. When the tx details is closed the app is blocked with splash screen and no way to remove it without app cold start 

**Steps To Reproduce**
Precondition:
1. Add safe and import 1 owner
2. set up passcode on app start ( with or without biometric)
Steps
1. create tx on web that will require a confirmation from the imported owner
2. Follow push notification on mobile
Current result: Tx details is opened without passcode requesting 

3. Confirm or reject tx
Current result: the passcode or biometric data is requested

4. Close tx details to go to the app
Current result: Splash screen is displayed and there is no way to close it without cold app start

**Expected Result**
To discuss:
the passcode before tx details is displayed or the passcode when the tx details is closed befor
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/6f80318a-288c-40ac-bf6f-30a91775b497)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/59e74c6c-e096-4bb7-a034-a19948ea9e96)
**Device & App version**
 - Device/OS: iPhone 11/iphone 8 /14.3
 - App version: 2.14.0 (10)
 
**Environment**
prod rinkeby
**Additional info**
(Optional)

sche:
nice one

tschubotz:
yea, nice one. I'd say we shouldn't block the release, but fix this for the next one -> `major` for me

liliya-soroka:
Verified
app version 2.15.0 (406)

PR #1112, opened 2021-04-21
gh-1105 incapsulate presentation logic for main window in scene delegate

sche:


Multisig/App/SceneDelegate.swift
```diff
@@ -121,6 +121,14 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
     // MARK: - Window Management
 
+    func presentForMain(_ controller: UIViewController) {
+        if windowState == .main {
+            mainWindow?.rootViewController?.present(controller, animated: true)
+        } else {
+            shouldBePresentedController = controller
```
DmitryBespalov: can we rename for a better name? This seems not specific -- what is the purpose of this view controller?

`delayedMainRootViewController` maybe

DmitryBespalov (approved):

Multisig/App/SceneDelegate.swift
sche: renamed to delayedMainPresentedViewController

MouazAlzahabi (approved):

DmitryBespalov (approved):

PR #1125, opened 2021-04-28
Fix issue

MouazAlzahabi:


Multisig/Cross-layer/FirebaseRemoteConfig.swift
```diff
@@ -26,9 +26,20 @@ class FirebaseRemoteConfig {
...
-    func value<T>(key: Key) -> T? {
-        remoteConfig.configValue(forKey: key.rawValue) as? T
+    func value(key: Key) -> String? {
+        remoteConfig[key.rawValue].stringValue
+    }
+
+    func fetchConfig() {
+        remoteConfig.fetchAndActivate { status, error in
+            if [RemoteConfigFetchAndActivateStatus.successFetchedFromRemote, .successUsingPreFetchedData].contains(status) {
+                print("Config fetched!")
```
sche: Could we use LogService.shared.info here and below?

sche (approved):

Issue #1140, opened 2021-05-05
Add App Tracking Transparency


sche:
https://developer.apple.com/documentation/apptrackingtransparency

1. We need to have one button on terms screen saying "Agree" and the other one saying "No, thanks" (or what it was before?) to close the "app terms" sheet
2. After agreeing, we present the system App Tracking alert and depending on user choice:
  - if "Allow" - then enable tracking and enable crashlytics
  - if "Ask App Not To Track" - then do not enable tracking or crashlytics
3. In the advanced app settings, we may need to rephrase it to "Share Usage and Crash Data" instead of "Share Usage Data"
4. For existing users updating to iOS 14.5, the app should ask the system permission with the same logic as above.
5. If user changes the permission to **deny** tracking in the phone/device settings, then the app must disable tracking. 
6. If the user changes permission to **allow** tracking for the app without first enabling it in the app settings, then the app still does not track, because the user would need to enable tracking in app settings.
7. If user enables the tracking in the app settings but the system permission is denied, then the app should show alert redirecting to the settings. The alert would say
  - Please allow tracking in Settings
  - "Cancel" button - closes alert
  - "Settings" button - opens device settings
  - If user then goes to settings, enables the tracking, and goes back to the app, then (if technically possible), the app is allowed to enable tracking. Please verify that this requirement and requirement 6 above do not conflict with each other.

liliya-soroka:
Verified
app version 2.16.0 (429)

PR #1141, opened 2021-05-05
gh-1140 add app tracking transparency

sche:
Initially I tried to make system allowance bounded to our switch, but it appeared to be simpler.
This request is only for activating IDFA that we do not use.
image

### DIFF
--- a/Multisig/Info.plist
+++ b/Multisig/Info.plist
@@ -19,7 +19,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.15.1</string>
+	<string>2.16.0</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>CLIENT_GATEWAY_URL</key>
@@ -59,7 +59,7 @@
 	<key>NSFaceIDUsageDescription</key>
 	<string>This will allow Safe Multisig to access your Face ID to open app and sign transactions.</string>
 	<key>NSUserTrackingUsageDescription</key>
-	<string></string>
+	<string>We collect anonymized app usage data and crash reports to ensure the quality of our app.</string>
 	<key>PAY_FOR_CANCELLATION_URL</key>
 	<string>$(PAY_FOR_CANCELLATION_URL)</string>
 	<key>PRIVACY_URL</key>
@@ -158,5 +158,7 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>FirebaseCrashlyticsCollectionEnabled</key>
+	<false/>
 </dict>
 </plist>
--- a/Multisig/Info.plist
+++ b/Multisig/Info.plist
@@ -19,7 +19,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.15.0</string>
+	<string>2.15.1</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>CLIENT_GATEWAY_URL</key>
@@ -58,6 +58,8 @@
 	<string>Safe need to use camera to scan QR code.</string>
 	<key>NSFaceIDUsageDescription</key>
 	<string>This will allow Safe Multisig to access your Face ID to open app and sign transactions.</string>
+	<key>NSUserTrackingUsageDescription</key>
+	<string></string>
 	<key>PAY_FOR_CANCELLATION_URL</key>
 	<string>$(PAY_FOR_CANCELLATION_URL)</string>
 	<key>PRIVACY_URL</key>
--- a/Multisig/UI/App/ViewControllerFactory.swift
+++ b/Multisig/UI/App/ViewControllerFactory.swift
@@ -23,20 +23,24 @@ enum ViewControllerFactory {
             nav.modalPresentationStyle = .fullScreen
             nav.modalTransitionStyle = .crossDissolve
 
-            let start = LaunchView(acceptedTerms: .constant(false), onStart: { [weak nav] trackingEnabled in
-                AppSettings.trackingEnabled = trackingEnabled
-                if trackingEnabled {
-                    // https://firebase.google.com/docs/ios/supporting-ios-14
-                    // This is required only for IDFA that we do not use
-                    if #available(iOS 14, *) {
-                        ATTrackingManager.requestTrackingAuthorization { status in
-                            DispatchQueue.main.async {
-                                nav?.popViewController(animated: false)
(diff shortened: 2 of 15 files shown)


Multisig/Info.plist
```diff
@@ -58,6 +58,8 @@
 	<string>Safe need to use camera to scan QR code.</string>
 	<key>NSFaceIDUsageDescription</key>
 	<string>This will allow Safe Multisig to access your Face ID to open app and sign transactions.</string>
+	<key>NSUserTrackingUsageDescription</key>
+	<string></string>
```
DmitryBespalov: The actual description string is missing. I think we should take the string from our terms screen, but it's better to check wording with @tschubotz:

```Anonymized app usage data will be used to improve the app```

File: Multisig/UI/App/ViewControllerFactory.swift [hunk 2]
```diff
@@ -22,7 +23,20 @@ enum ViewControllerFactory {
...
             nav.modalTransitionStyle = .crossDissolve
 
-            let start = LaunchView(acceptedTerms: .constant(false), onStart: { [weak nav] in
+            let start = LaunchView(acceptedTerms: .constant(false), onStart: { [weak nav] trackingEnabled in
+                AppSettings.trackingEnabled = trackingEnabled
+                if trackingEnabled {
+                    // https://firebase.google.com/docs/ios/supporting-ios-14
+                    // This is required only for IDFA that we do not use
+                    if #available(iOS 14, *) {
+                        ATTrackingManager.requestTrackingAuthorization { status in
```
DmitryBespalov: in this case, I would expect that if the user says "Ask app not to track", then we need to disable tracking, even if user pressed "Agree" button. Do you think this is correct?

Multisig/UI/SwiftUI/Settings/App Settings/AdvancedAppSettings.swift
```diff
@@ -58,6 +59,13 @@ struct AdvancedAppSettings: View {
             VStack {
                 Toggle(isOn: $trackingEnabled.didSet { enabled in
                     AppSettings.trackingEnabled = enabled
+                    // https://firebase.google.com/docs/ios/supporting-ios-14
+                    // This is required only for IDFA that we do not use
+                    if #available(iOS 14, *) {
+                        if ATTrackingManager.trackingAuthorizationStatus == .notDetermined && enabled {
+                            ATTrackingManager.requestTrackingAuthorization { _ in }
```
DmitryBespalov: same here, I would expect that if the user denied tracking authorization, we would need to disable the tracking.

Multisig/UI/App/ViewControllerFactory.swift
sche: This is what I did initially, but then if to read system message on the alert it says about tracking _across_ different apps/sites that is not related to tracking within the app

DmitryBespalov: the point is that we share the tracking information and it travels from the device to other servers, so Apple treats it as tracking and that is what expected case. I think it is misleading if user disallows and we still keep the "tracking" enabled, even if the user tried to enable the tracking before

Multisig/Info.plist
tschubotz: Where will this be displayed?

DmitryBespalov: In the alert (https://github.com/gnosis/safe-ios/pull/1141#issue-630607161) between the text and button on that screenshot

sche: https://developer.apple.com/app-store/user-privacy-and-data-use/
I made it empty on purpose. So it will only display the bold part

DmitryBespalov: @sche on that site, it clearly says:

```You must also include a purpose string in the system prompt that explains why you’d like to track the user.```

DmitryBespalov: And more:

```Use the purpose string to explain what this data will be used for to help the user understand what they’re opting in to share. ```

Multisig/UI/App/ViewControllerFactory.swift
sche: This was my way of thinking too before I dug into more details. If google tracking should be bound to this permission, then it would be logical to make this check within google lib. 
After reading some info (e.g. here https://firebase.google.com/docs/ios/supporting-ios-14) - "Analytics event logging, event reporting, and conversion measurement are unaffected, but attribution is impacted if IDFA is not accessible."
Taking into account apple system message stating only cross-apps tracking I did not bind them. So this permission is only about sharing IDFA info.

sche: @tschubotz please let us know what do you think we should do

Multisig/Info.plist
tschubotz: ah ok, I think 


> `Anonymized app usage data will be used to improve the app.`

sounds good then!

Multisig/UI/App/ViewControllerFactory.swift
DmitryBespalov: This is inconsequential, with this reasoning, why do we have to use this tracking manager at all, if it only required for IDFA tracking?

DmitryBespalov: By the Apple's User Privacy and Data Use, `Tracking also refers to sharing user or device data with data brokers`, so we must ask permission with this alert.

However, if user chooses "Ask App Not To Track", the Firebase Analytics event collection is not affected, i.e. we still get anonymized data. 

In my opinion, this situation - if user selects that option but we still have analytics enabled - is going to be confusing for the user. The natural question is: why does the app still has tracking enabled even if I asked the app not to track.

So the user's choice would be not to track.

For that reason, we should disable the tracking if that choice was given

tschubotz: hmmm. After reading https://github.com/firebase/firebase-ios-sdk/issues/5928, I think as long as this IDFA not set, they don't match the user's data on their, no? Hence we perhaps don't even need this popup?

sche: IMO this require more product thinking.
If to follow this logic, then on terms screen I would remove "Agree without tracking" button and would alway request this dialog. If user says "no" later she can enable it in advanced settings, but this will be not related to this permission. And what to do if user goes to phone settings and switches tracking off for the app (like with push notifications)?

DmitryBespalov: ok, after discussion:

- We need to have one button on terms screen saying "Agree" and the other one saying "No, thanks" (or what it was before?) to close the "app terms" sheet
- After agreeing, we present the system App Tracking alert and depending on user choice:
  - if "Allow" - then enable tracking and enable crashlytics
  - if "Ask App Not To Track" - then do not enable tracking or crashlytics
- In the advanced app settings, we may need to rephrase it to "Share Usage and Crash Data" instead of "Share Usage Data"
- For existing users updating to iOS 14.5, the app should ask the system permission with the same logic as above.
- If user changes the permission in the phone/device settings, then the app should disable tracking if the user denied the permission. If the user changes permission to allow tracking in the device, then the app can start tracking only if it is also enabled in the app settings.

tschubotz: For users on iOS 13, things should still work as before?

PR #1142, opened 2021-05-06
Gh 1140 add app tracking transparency

sche:


Multisig/UI/App/ViewControllerFactory.swift
```diff
@@ -17,19 +18,29 @@ enum ViewControllerFactory {
     static func rootViewController() -> UIViewController {
         let tabBarVC = MainTabBarViewController()
 
-        if let updateViewController = App.shared.updateController.makeUpdateAppViewController() {
```
sche, line 20: @MouazAlzahabi this was disabled in release branch, and I did not return it back because this is a buggy implementation as it can allow to start using app without accepting terms

Multisig/Logic/Models/AppSettings.swift
```diff
@@ -25,9 +26,20 @@ extension AppSettings {
     static var termsAccepted: Bool
 
     @AppSetting(\.trackingEnabled)
+    private static var _trackingEnabled: Bool
```
DmitryBespalov: maybe let's name it `trackingEnabledSetting` as the underscoring is not by the coding guidelines we adopted

Multisig/UI/SwiftUI/Onboarding/LaunchView.swift
```diff
@@ -39,7 +39,7 @@ struct LaunchView: View {
     @State
     var showTerms = false
 
-    var onStart: () -> Void = { }
+    var onStart: () -> Void
```
DmitryBespalov: maybe this can be reverted and we don't need to add empty `{}` below?

DmitryBespalov (approved):

Multisig/Logic/Models/AppSettings.swift
sche: changed

Multisig/UI/SwiftUI/Onboarding/LaunchView.swift
sche: reverted

PR #1143, opened 2021-05-06
Change app update presentation logic

MouazAlzahabi:


Multisig/App/SceneDelegate.swift
```diff
@@ -127,6 +130,14 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         }
     }
 
+    private func showAppUpdate() {
+        if appUpdateWindow != nil {
```
DmitryBespalov: instead of relying on the presence of the app update window, it is better to introduce new state in the WIndowState and modify the logic accordingly

sche: +1

MouazAlzahabi: How I see this is that displaying app update is separated from Window state.
Even if we introduced new state we will use it when we could create appUpdateWindow or in other words if there is new update. 

I update the implementation based on your comment, please verify

DmitryBespalov (approved):

sche (approved):

PR #1146, opened 2021-05-09
Gh 1144/now newversion flow

MouazAlzahabi:


sche (approved):

Multisig/Cross-layer/FirebaseRemoteConfig.swift
```diff
@@ -25,6 +25,10 @@ class FirebaseRemoteConfig {
         remoteConfig = RemoteConfig.remoteConfig()
         let settings = RemoteConfigSettings()
         remoteConfig.configSettings = settings
+        // we have this codition to make it faster to test on staging
+        if Bundle.main.isStaging {
```
DmitryBespalov: Sorry but this extension on Bundle.main is not needed because we already have this thing implemented:

`App.configuration.environment != .production`

I would keep it as "not production" so that it would work both in dev and staging.

Multisig/Logic/UpdateController.swift
```diff
@@ -18,11 +18,19 @@ class UpdateController {
...
             style = .required
-        } else {
+            AppSettings.lastIgnoredUpdateVersion = nil
+        } else if let deprecatedSoonVersionsRange = deprecatedSoonVersions, check(value: appVersion, in: deprecatedSoonVersionsRange) {
             style = .recommended
+            AppSettings.lastIgnoredUpdateVersion = nil
+        } else if latestAppVersion == AppSettings.lastIgnoredUpdateVersion {
+            return nil
+        }
+        else {
```
DmitryBespalov: should be `} else {`

Multisig/Cross-layer/FirebaseRemoteConfig.swift
MouazAlzahabi: I forgot about this one, thank you

DmitryBespalov (approved):

Issue #1089, opened 2021-04-14
Signing V2: Adjust "read only" label visibility logic


elgatovital:
Show "read only" label only after backend response is received and it was verified, that safe owners are not among owners imported on the device.
![image.png](https://images.zenhubusercontent.com/5ccabf56658809463cff37ed/6d2c1157-0b08-4a12-8865-2ea73ed30522)

PR #1150, opened 2021-05-10
gh-1089 update read-only label after response from backend

sche:


Multisig/UI/App/HeaderViewController/HeaderViewController.swift
```diff
@@ -103,7 +103,9 @@ final class HeaderViewController: ContainerViewController {
             if let safe = selectedSafe {
                 safeBarView.setAddress(safe.addressValue)
                 safeBarView.setName(safe.displayName)
-                safeBarView.setReadOnly(safe.isReadOnly)
+                if updateReadOnlyLabel {
+                    safeBarView.setReadOnly(safe.isReadOnly)
```
DmitryBespalov: what if instead of this argument we would just change the default `isReadOnly` value to `false`?

In the implementation of that isReadOnly we need to check that if `ownersInfo` is nil, then return false; otherwise decide based on the implemented logic.

MouazAlzahabi (approved):

Multisig/UI/App/HeaderViewController/HeaderViewController.swift
sche: ok

Issue #1176, opened 2021-05-20
Create How to Add Key screen


DmitryBespalov:
**Texts**

*   **Title**
    *   Add Owner Key 
*   Create New Owner Key
    *   Create a new key that you can use as owner of your Gnosis Safe
*   Import Existing Owner Key
    *   Import an existing owner key or a seed phrase

**Zeplin**

https://zpl.io/Vq3ZB9Y

*   Links

**Tracking**

*   Just screen tracking: screen_owner_options

liliya-soroka:
Verified
app version 2.17.0 (450)

PR #1187, opened 2021-05-26
gh-1176 add key type; update deps

sche:


Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -10,13 +10,24 @@ import Foundation
 import CoreData
 import Web3
 
+/// Enum for storing key type in the presistanse store. The order of existing items should not be changed.
```
DmitryBespalov: can't help but notice: presistanse -> persistence

Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -10,13 +10,24 @@ import Foundation
...
+enum KeyType: Int, CaseIterable {
+    case deviceImported
```
DmitryBespalov: I recommend adding ` = 0` and `= 1` explicitly so that order can be changed

Multisig/Logic/Models/KeyInfo.swift
sche: ok

DmitryBespalov (approved):
Just minor comments, mergeable!

Multisig/Logic/Models/KeyInfo.swift
sche: ok

MouazAlzahabi (approved):

PR #1192, opened 2021-05-26
Create KeyDetailsViewController, update tracking key count logic

MouazAlzahabi:


Multisig/Cross-layer/Analytics/TrackingEvent.swift
```diff
@@ -24,10 +25,15 @@ extension Tracker {
         setUserProperty(status, for: TrackingUserProperty.pushInfo)
     }
 
-    func setNumKeysImported(_ count: Int) {
-        setUserProperty("\(count)", for: TrackingUserProperty.numKeysImported)
+    func setNumKeys(_ count: Int, type: KeyType) {
+        if type == .deviceGenerated {
```
sche: switch ?

Multisig/Logic/PrivateKeyController.swift
```diff
@@ -15,7 +15,7 @@ class PrivateKeyController {
 
             App.shared.notificationHandler.signingKeyUpdated()
 
-            Tracker.shared.setNumKeysImported(KeyInfo.count)
+            Tracker.shared.setNumKeys(KeyInfo.count(.deviceImported), type: .deviceImported)
```
sche, line 18: I will need your changes before merging my pr

MouazAlzahabi: That's what I thought also

sche (approved):

Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -42,12 +42,14 @@ extension KeyInfo {
...
-    static var count: Int {
+    static func count(_ type: KeyType? = nil) -> Int {
         do {
-            let context = App.shared.coreDataStack.viewContext
-            let fr = KeyInfo.fetchRequest().all()
-            let itemCount = try context.count(for: fr)
-            return itemCount
+            let items = try KeyInfo.all()
+            if let type = type {
+                return items.filter({ keyInfo in keyInfo.keyType == type }).count
```
DmitryBespalov, line 49: Just an idea, I think it makes sense to execute a "count" fetch request which will return the count from the database, no need to count manually.

But this would be more complicated. Until we see this has performance issues, I'm ok with this implementation.

DmitryBespalov (approved):

Issue #1178, opened 2021-05-20
Changes to Enter Key Name screen


DmitryBespalov:
*   **Back** button
    *   Goes back to the previous screen. When this screen is opened again, the same key address is displayed. To generate a new key, the user has to close this flow and start again.
*   **Save **button
    *   Disabled if the name is empty or whitespaces
    *   Enables when name is entered
    *   We need to also save that the private key was generated from that seed phrase in order to allow exporting of both later.
    *   **When pressed,** 
        *   will save the seed phrase securely and the private key
        *   Updates tracking user properties and triggers tracking events
        *   Closes the flow and opens the key details
        *   Shows snackbar
            *   The key successfully created. Add it to a Safe on desktop and then restart the mobile app.
    *   

**Zeplin**

*   Links

**Tracking**

*   Existing screen id: screen_owner_enter_name
*   Update user property: [num_keys_generated](https://github.com/gnosis/safe/wiki/Tracking#num_keys_generated)
*   Trigger: [user_key_generated](https://github.com/gnosis/safe/wiki/Tracking#user_key_generated)

liliya-soroka:
Verified
app version 2.17.0(450)

PR #1193, opened 2021-05-26
gh-1178 save generated key

sche:


Multisig/Logic/Models/PrivateKey/PrivateKey.swift
```diff
@@ -19,26 +20,36 @@ struct PrivateKey {
...
     }
 
-    var data: Data {
+    struct PKData: Codable {
+        let key: Data
+        let mnemonic: String?
+    }
+
+    /// needed for PrivateKeyController.migrateLegacySigningKey()
+    var keyData: Data {
```
DmitryBespalov, line 29: maybe calling it `legacyKeyData` would be better

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/OnboardingGenerateKeyViewController.swift
```diff
@@ -43,7 +52,30 @@ class OnboardingGenerateKeyViewController: UITableViewController {
...
+    @objc private func didTapNextButton(_ sender: Any) {
+        let vc = EnterAddressNameViewController()
+        vc.actionTitle = "Save"
+        vc.descriptionText = "Choose a name for the owner key. The name is only stored locally and will not be shared with Gnosis or any third parties."
+        vc.screenTitle = "Enter Key Name"
+        vc.trackingEvent = .enterKeyName
+        vc.placeholder = "Enter name"
+        vc.address = privateKey.address
+        vc.completion = { [unowned self, vc] name in
+            guard PrivateKeyController.importKey(privateKey, name: name, isDrivedFromSeedPhrase: true) else { return }
```
DmitryBespalov, line 64: I'm a bit confused - shouldn't we go directly to the key details after entering key name?

From the task description:
```
When pressed,
will save the seed phrase securely and the private key
Updates tracking user properties and triggers tracking events
Closes the flow and opens the key details
Shows snackbar
The key successfully created. Add it to a Safe on desktop and then restart the mobile app.
```

DmitryBespalov:
@sche ok, as we agreed, please address the comments in another pr 👍

DmitryBespalov (approved):

Multisig/Logic/Models/PrivateKey/PrivateKey.swift
sche: ok

Issue #1181, opened 2021-05-20
Export Key screen


DmitryBespalov:
*   Title
    *   Export 
*   Segmented buttons
    *   **Seed Phrase**
        *   Shows the seed phrase with numbered words
        *   NO QR-code
    *   **Private Key**
        *   Shows the private key as a hexadecimal string
        *   QR code: content is the private key string
*   **Share **button
    *   Opens system “Share” sheet to share either seed phrase or a private key (whatever is currently selected
*   **Back** button
    *   Closes the flow and goes back to previous screen (key details)

**Zeplin**

https://zpl.io/VDgYQMg
https://zpl.io/2j3BdmA

*   Links

**Tracking**

*   Screen id when seed is active: screen_owner_export_seed
*   Screen id when private key is active: screen_owner_export_key

liliya-soroka:
Verified
app version 2.17.0 (450)

the tracking will be checked separately

PR #1197, opened 2021-05-27
gh-1181 Fixed export screen opening

DmitryBespalov:
- Fixed: seed phrase now with 12 words instead of 24
- Fixed: export key screen shows actual seed phrase and/or private key

### DIFF
--- a/Multisig/Data/Services/Utils/GSError.swift
+++ b/Multisig/Data/Services/Utils/GSError.swift
@@ -431,7 +431,7 @@ enum GSError {
         let howToFix = "Please remove this key"
         let domain = iOSErrorDomain
         let code = 1315
-        let loggable = false
+        let loggable = true
     }
 
     struct PrivateKeyFetchError: DetailedLocalizedError {
@@ -440,6 +440,6 @@ enum GSError {
         let howToFix = "Please try again"
         let domain = iOSErrorDomain
         let code = 1316
-        let loggable = false
+        let loggable = true
     }
 }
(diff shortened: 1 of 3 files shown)


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -424,4 +424,22 @@ enum GSError {
...
         let loggable = false
     }
+
+    struct PrivateKeyDataNotFound: DetailedLocalizedError {
+        let description = "Failed to find private key data"
+        let reason: String
+        let howToFix = "Please remove this key"
+        let domain = iOSErrorDomain
+        let code = 1315
+        let loggable = false
```
sche: I would make it loggable

Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -424,4 +424,22 @@ enum GSError {
...
+    }
+
+    struct PrivateKeyFetchError: DetailedLocalizedError {
+        let description = "Failed to fetch private key"
+        let reason: String
+        let howToFix = "Please try again"
+        let domain = iOSErrorDomain
+        let code = 1316
+        let loggable = false
```
sche: same here

sche (approved):

Multisig/Data/Services/Utils/GSError.swift
DmitryBespalov: why so?

PR #1236, opened 2021-06-09
Change naming from Multisig

MouazAlzahabi:


Multisig/Info.plist
```diff
@@ -86,7 +86,7 @@
 	<key>NSCameraUsageDescription</key>
 	<string>Safe need to use camera to scan QR code.</string>
 	<key>NSFaceIDUsageDescription</key>
-	<string>This will allow Safe Multisig to access your Face ID to open app and sign transactions.</string>
+	<string>This will allow Safe to access your Face ID to open app and sign transactions.</string>
```
sche: Should it be "Gnosis Safe"?

Multisig/Info.plist
```diff
@@ -7,7 +7,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
 	<key>CFBundleDisplayName</key>
-	<string>$(DISPLAY_NAME)</string>
+	<string>Safe</string>
```
sche: Should it be Gnosis Safe?

sche:
I would request a review from Tobi / Lukas on these changes. We have mixture of Safe/Gnosis Safe and it would be better to double check where to use what

sche (approved):

Issue #1229, opened 2021-06-07
Add unit tests for the Balances List view controller


DmitryBespalov:
Explore what and where can be tested and covered. Goal is 100% coverage (if possible).
Investigate Snapshot testing.

# QA notes
Nothing to QA

liliya-soroka:
closed as no QA required

PR #1238, opened 2021-06-09
gh-1229 Add tests

DmitryBespalov:
# Setting Up Git LFS for snapshot testing

https://docs.github.com/en/github/managing-large-files/versioning-large-files/about-git-large-file-storage

    brew install git-lfs


MultisigIntegrationTests/UI Snapshot Tests/BalanceTableViewCellTests.swift
```diff
@@ -0,0 +1,48 @@
...
+            cell.setDetailText("")
+            cell.setSubDetailText("")
+            cell.setImage(UIImage())
+        }
+        assertSnapshot(matching: vc, as: .image(on: .iPhoneX))
+    }
+
+    func testLongStrings() {
+        let vc = CellTestViewController<BalanceTableViewCell>(estimatedHeight: 60) { cell in
+            cell.setMainText(String(repeating: "ETH repeating ", count: 15))
```
DmitryBespalov, line 41: this needs to be fixed actually, because only amounts are shown

sche: so what is wrong here?

MultisigIntegrationTests/UI Snapshot Tests/BalanceTableViewCellTests.swift
```diff
@@ -0,0 +1,48 @@
...
+        let vc = CellTestViewController<BalanceTableViewCell>(estimatedHeight: 60) { cell in
+            cell.setMainText("ETH")
+            cell.setDetailText("0,005")
+            cell.setSubDetailText("155,84 USD")
+            cell.setImage(UIImage(named: "ico-ether")!)
+        }
+        assertSnapshot(matching: vc, as: .image(on: .iPhoneX))
+    }
+
+    func testNotConfigured() {
```
sche: do we need to test notConfigured/Empty at all?

sche:
How about a test using BalancesVC with mock data to increase coverage for the whole controller?

MultisigIntegrationTests/UI Snapshot Tests/BalanceTableViewCellTests.swift
DmitryBespalov: These are testing edge cases. Not configured is used because the XIB had some text, so I fixed it. Empty was used to double-check that it works with empty values.

DmitryBespalov:
@sche I want to start small 😉

sche (approved):

PR #1242, opened 2021-06-10
WalletConnect ledger wallet support

sche:
- implemented signing typed data
- added support for Trust Wallet, Rainbow and LedgerLive - now it is possible to sign and execute transactions with Ledger Nano X! 🎉
- small refactoring of signing. I checked that it works well for confirmations and rejections

### DIFF
--- a/Multisig/Logic/Models/Transaction.swift
+++ b/Multisig/Logic/Models/Transaction.swift
@@ -27,7 +27,7 @@ struct Transaction: Codable {
     // can be modified for WalletConnect transactions
     var nonce: UInt256String
     // computed based on other properties
-    var safeTxHash: HashString?
+    var safeTxHash: HashString!
     var transactionHash: HashString?
 }
 
--- a/Multisig/Logic/Models/Transaction.swift
+++ b/Multisig/Logic/Models/Transaction.swift
@@ -81,16 +81,16 @@ extension Transaction {
     }
 
     mutating func updateSafeTxHash() {
-        safeTxHash = safeTxHash(by: safe!.address)
+        safeTxHash = safeTransactionHash()
     }
 
-    func encodeTransactionData(for safe: AddressString) -> Data {
+    func encodeTransactionData() -> Data {
         let ERC191MagicByte = Data([0x19])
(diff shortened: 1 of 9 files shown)


Multisig/Logic/Models/Transaction.swift
```diff
@@ -27,7 +27,7 @@ struct Transaction: Codable {
     // can be modified for WalletConnect transactions
     var nonce: UInt256String
     // computed based on other properties
-    var safeTxHash: HashString?
```
DmitryBespalov, line 30: guaranteed to be not nil?

Multisig/UI/WalletConnect/WalletConnectClientController.swift
```diff
@@ -157,6 +127,53 @@ class WalletConnectClientController {
...
+            } catch {
+                completion(.failure(error))
+            }
+        }
+
+        do {
+            switch session.walletInfo?.peerMeta.name ?? "" {
+
+            // we call signTypedData only for wallets supporting this feature
+            case "MetaMask", "LedgerLive", "🌈 Rainbow", "Trust Wallet":
```
DmitryBespalov, line 160: what if their name changes? do they have some id?

DmitryBespalov (approved):

Multisig/Logic/Models/Transaction.swift
sche: yes. We calculate it everywhere where we initiate transaction

Multisig/UI/WalletConnect/WalletConnectClientController.swift
sche: We have only peerMeta like 
```
"peerMeta":{"description":"MetaMask Mobile app","icons":["https:\/\/raw.githubusercontent.com\/MetaMask\/brand-resources\/master\/SVG\/metamask-fox.svg"],"name":"MetaMask","ssl":true,"url":"https:\/\/metamask.io"},"rpcUrl":""}}
```

So I thought that name is quite reliable. Otherwise eth_sign will be used that is also supported by majority of wallets, but not as user friendly as eth_signTypedData

Issue #1244, opened 2021-06-10
[iOS]Change import owner key banner to Add owner key banner 


liliya-soroka:
**Bug description**
Change import owner key banner to Add owner key banner.
1. Update text from import to add
2. Update tracking :
`user_banner_owner_import` -> `user_banner_owner_add`
`user_onboarding_owner_import` -> `user_onboarding_owner_add`

**Steps To Reproduce**
1. Clean app install
2. add the first safe without importing owner key
3. check the import owner key banner
Current result: The banner contains information about the importing owner key, but should contain information about adding owner key

**Expected Result**
1. Update text from import to add
2. Update tracking :
`user_banner_owner_import` -> `user_banner_owner_add`
`user_onboarding_owner_import` -> `user_onboarding_owner_add`
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/b4b9828f-d8fc-4b78-a0da-d5c85ffc5d03)
**Device & App version**
 - Device/OS: iPhone 11/14.3
 - App version: 2.17.0 (462)
 
**Environment**
prod, mainnet
**Additional info**
(Optional)

sche:
@tschubotz just to confirm. I am changing tracking values as specified here, right?

liliya-soroka:
Verified
app version 2.18.0 (474)

PR #1246, opened 2021-06-14
gh-1244 renamings

sche:


Multisig/Cross-layer/Analytics/TrackingEvent.swift
```diff
@@ -117,7 +117,7 @@ enum TrackingEvent: String, Trackable {
 
     case chooseOwner                                = "screen_owner_choose"
     case bannerImportOwnerKeySkipped                = "user_banner_owner_skip"
-    case bannerImportOwnerKeyImported               = "user_banner_owner_import"
+    case bannerImportOwnerKeyAdd                    = "user_banner_owner_add"
```
DmitryBespalov, line 120: I would not change tracking because it would invalidate previous tracking

DmitryBespalov: I mean tracking string value - the variable change is ok

DmitryBespalov (approved):
good, just one comment

Multisig/Cross-layer/Analytics/TrackingEvent.swift
sche: this was requested in the bug description. I will ask there one more time

sche: The change was requested during refinement meeting

PR #1258, opened 2021-06-17
Gh 1229/balances redesign

DmitryBespalov:
This is still work in progress. 

In this part I wanted to make the balances UIViewController testable with snapshot tests. That's why I went and re-designed it. You can see the [design documentation here](https://github.com/gnosis/safe-ios/blob/586af155efba0cf420857e695365dfe4eb548d8b/docs/Balances.pdf). 

The overall idea (not yet documented) was to go with view controllers for only display and getting user interaction events, and move all of the logic into testable controllers/logic objects (not created yet in this PR).

### DIFF
--- a/Multisig/UI/Assets/BalancesViewController/BalancesV2/BalancesTableViewController.swift
+++ b/Multisig/UI/Assets/BalancesViewController/BalancesV2/BalancesTableViewController.swift
@@ -1,0 +1,202 @@
+//
+//  BalancesTableViewController.swift
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 16.06.21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class BalancesTableViewController: UITableViewController {
+
+    private var data: Balances = .empty
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // configure table view
+        tableView.allowsSelection = false
+        tableView.backgroundColor = .primaryBackground
+        tableView.refreshControl = makeRefreshControl()
+
+        // register cells
+        tableView.registerCell(ImportKeyBannerCell.self)
+        tableView.registerCell(CreatePasscodeBannerCell.self)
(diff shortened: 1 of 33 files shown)


Multisig/UI/Assets/BalancesViewController/BalancesV2/BalancesTableViewController.swift
```diff
@@ -0,0 +1,202 @@
...
+
+        /// If not nil, then a banner will be shown.
+        /// Only one banner at a time.
+        var banner: BannerPlaceholder?
+
+        /// If not nil, then the entries and total cells will be shown
+        var summary: Summary?
+
+        /// Creates cell data from the banner and summary
+        var cellData: [BalancesCellData] {
```
sche, line 107: as this is an array I would make name plural. `cellsData`

MultisigIntegrationTests/UI Snapshot Tests/BalancesTableViewControllerTests.swift
```diff
@@ -0,0 +1,173 @@
...
+                    .init(image: LocalImageData(image: UIImage(named: "ico-token-placeholder")),
+                          name: "OWL",
+                          tokenAmount: "1",
+                          quoteAmount: "1 000 000 EUR")
+                ]
+            )
+        )
+        vc.reload(data)
+        assertSnapshot(matching: vc, as: .image(on: .iPhoneX))
+        assertSnapshot(matching: vc, as: .image(on: .iPhoneX, traits: UITraitCollection(userInterfaceStyle: .dark)))
```
sche, line 81: nice

sche (approved):

Multisig/UI/Assets/BalancesViewController/BalancesV2/BalancesTableViewController.swift
DmitryBespalov: 🤔 but "data" is already a plural

Issue #1264, opened 2021-06-22
[iOS] Add ribbon below the safe info header on all pages


sche:
- Assets, Transactions, Settings pages

- Add network indicator for Safe info / Share safe screen

- Don't show ribbon when no safes loaded

**QA**
- All pages where the safe header presents (safe name + address + safe switcher button) should have a ribbon with network name.
  - [ ] Assets tab
  - [ ] Transactions tab
  - [ ] Settings tab

https://zpl.io/2E4LwwD
https://zpl.io/Vxp7BBg
https://zpl.io/a8QLWZK

- Network name and network colour are correct

- When no safe is loaded, then ribbon is not shown
https://zpl.io/aNOe4pZ

- Safe Info (when tapping on the safe name in the header) should contain the network indicator.

https://zpl.io/2vjwBBr

liliya-soroka:
Verified
app version 2.19.0(482)

PR #1276, opened 2021-06-28
gh-1264 Implemented UI views

DmitryBespalov:
- Network Indicator view for reuse later in the load safe cell
- Safe Info view change
- Ribbon View added to Header controller

Right now it is only views without connection to the safe data. I'm depending on the other changes/prs for that.


Multisig/UI/App/HeaderViewController/HeaderViewController.swift
```diff
@@ -38,6 +40,11 @@ final class HeaderViewController: ContainerViewController {
         addObservers()
         headerBarHeightConstraint.constant = ScreenMetrics.safeHeaderHeight
         reloadSafeData()
+
+        #warning("Enable ribbon after Safe model implementation")
+        ribbonView.text = "Rinkeby"
```
MouazAlzahabi: I have two ideas here, 
1) we need to make this view as component and observe the change of network info, I already implemented post notification when changing the network so instead of doing this in all view controllers how about just do this in the ribbonView?

2) can't we find a way to add this ribbon in all view controllers together instead of adding it in the header in some and at top of others?

sche (approved):

Multisig/UI/App/HeaderViewController/HeaderViewController.swift
DmitryBespalov: this would be the next step when implementing it in other places, for now I kept it here because there's also a shadow that needs to be on top of the content view and ribbon is above the shadow

Multisig/Logic/Models/Network.swift
```diff
@@ -72,13 +72,15 @@ extension Network {
         let theme = NetworkTheme(context: context)
         theme.textColor = themeTextColor
         theme.backgroundColor = themeBackgroundColor
+        theme.network = network
 
         network.theme = theme
```
sche, line 77: do we need this line now?

Multisig/Logic/Models/Network.swift
```diff
@@ -72,13 +72,15 @@ extension Network {
...
 
         let token = NetworkToken(context: context)
         token.name = currencyName
         token.symbol = currencySymbl
         token.decimals = Int32(currencyDecimals)
+        token.network = network
 
         network.nativeCurrency = token
```
sche, line 85: and this

MultisigTests/UIColorStylesTests.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import XCTest
+@testable import Multisig
+import UIKit
+
+class UIColorStylesTests: XCTestCase {
+
+    func test_initHex() {
```
sche, line 15: 👍🏻

sche (approved):

MouazAlzahabi (approved):

PR #1278, opened 2021-06-30
Update ClientGateway&EthereumNodeService to use selected network

MouazAlzahabi:


Multisig/Cross-layer/App.swift
```diff
@@ -53,5 +51,23 @@ class App {
...
+
+        for name in updateNotifications {
+            NotificationCenter.default.addObserver(self,
+                                           selector: #selector(bindNetworkInfo),
+                                           name: name,
+                                           object: nil)
+        }
+    }
+
+    @objc func bindNetworkInfo() {
```
sche: could be private

Multisig/Data/Services/Safe Client Gateway Service/ConfirmTransactionRequest.swift
```diff
@@ -11,8 +11,10 @@ import Foundation
 struct ConfirmTransactionRequest: JSONRequest {
     var safeTxHash: String
     var signedSafeTxHash: String
+    let network: Int
+    
     var httpMethod: String { "POST" }
-    var urlPath: String { "/v1/transactions/\(safeTxHash)/confirmations" }
+    var urlPath: String { "/\(network)/v1/transactions/\(safeTxHash)/confirmations" }
```
rmeissner: The naming seems inconsistent with the other occurrences where it is called `chainId` ;)

Multisig/Cross-layer/App.swift
MouazAlzahabi: No, it's used at AppDelegate

Multisig/Cross-layer/App.swift
```diff
@@ -53,5 +51,23 @@ class App {
...
+        if let network = try? Safe.getSelected()?.network {
+            clientGatewayService.bindNetwork(Int(network.chainId))
```
sche: This approach may break monitoring functionality. Right now we monitor pending WalletConnect transactions, but in future we may monitor more things - e.g. if we add tx execution functionality we might monitor tx status.
On the other hand we can add optional `network` parameter in future.

Multisig/Cross-layer/App.swift
sche: or can't because you use it in AppDelegate?

Multisig/Data/Services/Safe Client Gateway Service/BalancesRequest.swift
```diff
@@ -35,8 +34,8 @@ struct BalancesRequest: JSONRequest {
 }
 
 extension BalancesRequest {
-    init(_ address: Address) {
-        self.init(safeAddress: address.checksummed, fiat: AppSettings.selectedFiatCode)
+    init(_ address: Address, chainId: Int) {
```
sche: As a next refactoring step (our of scope of this PR) I would look into passing everywhere Safe object, that already has it's network, so probably we can remove network binding logic

Multisig/Cross-layer/App.swift
DmitryBespalov: 👍

Multisig/Data/Services/Safe Client Gateway Service/FiatCodesRequest.swift
```diff
@@ -9,19 +9,19 @@
...
-    var httpMethod: String {
-        "GET"
-    }
+    let chainId: Int
+    
+    var httpMethod: String { "GET" }
 
     var urlPath: String {
-        "/v1/balances/supported-fiat-codes"
+        "/\(chainId)/v1/balances/supported-fiat-codes"
```
sche: As I can see here https://github.com/gnosis/safe-client-gateway/blob/b152f3e3d27c85efeb96c9baf8b43c7f5f8a6d93/src/routes/balances.rs#L62 there is no need in chain id here

Multisig/Data/Services/Safe Client Gateway Service/TransactionDetailsRequest.swift
```diff
@@ -10,23 +10,26 @@ import Foundation
...
 
 extension TransactionDetailsRequest {
-    init(safeTxHash: Data) {
+    init(safeTxHash: Data, chainId: Int) {
         id = safeTxHash.toHexStringWithPrefix()
+        self.chainId = chainId
     }
 }
 
 extension SafeClientGatewayService {
```
DmitryBespalov, line 27: @MouazAlzahabi did you run integration tests? I think client gateway tests will break now

Multisig/Data/Services/Safe Client Gateway Service/BalancesRequest.swift
DmitryBespalov: @sche hm, maybe that's not a good one, because this will couple the request to the database model, I think the current approach is good

Multisig/Data/Services/Safe Client Gateway Service/ProposeTransactionRequest.swift
```diff
@@ -67,13 +68,17 @@ extension SafeClientGatewayService {
         return asyncExecute(request: ProposeTransactionRequest(safe: transaction.safe!,
                                                                sender: sender,
                                                                signature: signature,
-                                                               transaction: transaction),
+                                                               transaction: transaction,
+                                                               chainId: chainId),
                             completion: completion)
     }
 
     func proposeTransaction(transaction: Transaction, sender: AddressString, signature: String) throws {
```
sche: e.g. here transaction should already have attached safe, so chainId could be fetched as `transaction.safe.network!.id`

Multisig/Data/Services/Safe Client Gateway Service/ConfirmTransactionRequest.swift
sche: I would rename in other places into `network` as we use this name in our model

MouazAlzahabi: Yes, by mistake :)

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -538,6 +538,28 @@ extension SCGModels {
...
+            case nativeCurrency
+            case theme
+        }
+
+        init(from decoder: Decoder) throws {
+            let values = try decoder.container(keyedBy: CodingKeys.self)
+            var string = try! values.decode(String.self, forKey: .chainId)
+            chainId = Int(string) ?? 0
+            chainName = try values.decode(String.self, forKey: .chainName)
+            string = try! values.decode(String.self, forKey: .rpcUrl)
```
sche: why do you mix `try!` and `try`?

Multisig/Data/Services/Safe Client Gateway Service/SafeClientGatewayService.swift
```diff
@@ -10,6 +10,7 @@ import Foundation
 
 class SafeClientGatewayService {
     var url: URL
+    var chainId: Int! = 1
```
sche: we have a enum: Network.ChainID.mainEthereumNetwork

Multisig/Logic/Models/Network.swift
```diff
@@ -91,7 +91,7 @@ extension Network {
     static func create(_ networkInfo: SCGModels.Network) throws -> Network {
         try Network.create(chainId: networkInfo.chainId,
                      chainName: networkInfo.chainName,
-                     rpcUrl: networkInfo.authenticatedRpcUrl,
+                     rpcUrl: networkInfo.rpcUrl,
```
sche, line 92: why did you change it here?

Multisig/Logic/NetworkManager.swift
```diff
@@ -13,12 +13,12 @@ class NetworkManager {
...
                 switch result {
-                case .success(let networks):
-                    for network in networks {
+                case .success(let data):
+                    for network in data.results {
                         Network.updateIfExist(network)
                     }
                     NotificationCenter.default.post(name: .networkInfoChanged, object: nil)
-                case .failure(_):
+                case .failure(let error):
```
sche: no need if we ignore

Multisig/Logic/Models/Network.swift
MouazAlzahabi: Because by the old why we store Infura key in the database as part of rpc url

Multisig/Data/Services/Safe Client Gateway Service/BalancesRequest.swift
```diff
@@ -9,17 +9,17 @@
...
 
 struct BalancesRequest: JSONRequest {
-    var safeAddress: String
-    var fiat: String
-    var isTrusted: Bool?
-    var isExcludeSpam: Bool?
+    private let safeAddress: String
+    private let networkId: Int
+    private let fiat: String
+    #warning("Check if we need these two")
```
DmitryBespalov: we don't use them because their default values are ok for mobile

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -538,6 +538,42 @@ extension SCGModels {
...
+            self.chainName = chainName
+            self.rpcUrl = rpcUrl
+            self.blockExplorerUrl = blockExplorerUrl
+            self.nativeCurrency = nativeCurrency
+            self.theme = theme
+        }
+
+        init(from decoder: Decoder) throws {
+            let values = try decoder.container(keyedBy: CodingKeys.self)
+            var string = try! values.decode(String.self, forKey: .chainId)
```
DmitryBespalov: I would not use forced try and forced unwrapping everywhere here

DmitryBespalov: I would throw

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -538,6 +538,42 @@ extension SCGModels {
...
+        enum CodingKeys: String, CodingKey {
+            case chainId
+            case chainName
+            case rpcUrl
+            case blockExplorerUrl
+            case nativeCurrency
+            case theme
+        }
+
+        init(chainId: Int,
```
DmitryBespalov: Instead of implementing these here, I would implement extension on URL that would make it Decodable

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
DmitryBespalov: I think if you'd define the value type as UInt256String, it will accept the stringified integer that can be used in other places, so we don't need to implement decodable here

Multisig/UI/WalletConnect/WCRequestsHandler.swift
```diff
@@ -99,7 +102,9 @@ class WCRequestsHandler: RequestHandler {
             }
         } else {
             do {
-                let result = try App.shared.nodeService.rawCall(payload: request.jsonString)
+                #warning("Make sure if this always works @Andrey")
+                let rpcURL = try! Safe.getSelected()!.network!.authenticatedRpcUrl
```
DmitryBespalov, line 106: @sche did you see it?

MouazAlzahabi: He asked me to put this

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -532,12 +532,16 @@ extension SCGModels {
 
 
     struct Network: Decodable {
-        let chainId: Int
+        let chainId: UInt256String
```
sche, line 535: 👍🏻

sche (approved):

bin/test.sh
```diff
@@ -1,4 +1,5 @@
 #! /usr/bin/env bash
+set -e +o pipefail
```
sche, line 2: @DmitryBespalov could it be a reason why unit tests can not run on Jenkins?

DmitryBespalov: I think, this is the reason:
Screenshot 2021-07-01 at 11 45 26

DmitryBespalov (approved):

PR #1280, opened 2021-07-01
Gh 1265/tx details ribbon

DmitryBespalov:
- Added ability to observe selected safe in the RibbonView itself and hide/show based on that
- Added RibbonViewController which is a wrapper/container over any other content controller that adds the ribbon on top of the screen.
- Using ribbon vc in transaction details screens
- Using ribbon vc in the collectible details


Multisig/UI/UI Library/View Controllers/LoadableViewController/LoadableViewController.swift
```diff
@@ -25,6 +25,7 @@ class LoadableViewController: UIViewController {
     @IBOutlet weak var dataErrorView: ScrollableDataErrorView!
     @IBOutlet weak var emptyView: ScrollableEmptyView!
     @IBOutlet weak var tableView: UITableView!
+    @IBOutlet weak var ribbonView: RibbonView!
```
sche: Why did you decide to use here a separate view instead of a wrapper view controller?

sche (approved):

MouazAlzahabi (approved):

Multisig/UI/UI Library/View Controllers/LoadableViewController/LoadableViewController.swift
DmitryBespalov: It's a good point! I've tried with the wrapper VC now, and it works OK, so I'll use that.

I did it because I first implemented ribbonView like that in previous pull request.

MouazAlzahabi (approved):

sche (approved):

PR #1281, opened 2021-07-02
Gh 1268 switch safes

sche:
image


MouazAlzahabi (approved):

MultisigIntegrationTests/CoreData/NetworkTests.swift
```diff
@@ -197,6 +197,40 @@ class NetworkTests: CoreDataTestCase {
             themeBackgroundColor: "#000000")
     }
 
+    func test_networkSafes() throws {
```
DmitryBespalov, line 200: Missing test cases:
- when no safes exist, then should return empty
- when all safes in the same network, should do the correct ordering

Maybe unrelated, but it's good if we fix it here:
- when multiple safes with different network and same address exist, then works fine
- when last safe removed from the network, then selects next safe in order

Multisig/UI/Safe Management/SwitchSafesViewController/SwitchSafesViewController.swift
```diff
@@ -11,16 +11,20 @@ import UIKit
 final class SwitchSafesViewController: UITableViewController {
     var notificationCenter = NotificationCenter.default
 
-    private var safes = [Safe]()
-    private let addSafeRowIndex = 0
+    private var networkSafes = Network.NetworkSafes()
+    private let addSafeSection = 0
```
DmitryBespalov, line 15: I would prefer a different design - with enums, but OK, no need to rewrite now

DmitryBespalov (commented):
tests missing

MultisigIntegrationTests/CoreData/NetworkTests.swift
sche: > when no safes exist, then should return empty
added 

> when all safes in the same network, should do the correct ordering
this is already covered

> when multiple safes with different network and same address exist, then works fine
added

> when last safe removed from the network, then selects next safe in order
I will add it in other already open PR

DmitryBespalov (approved):

PR #1300, opened 2021-07-08
Gh 1286/update chain id to be u int256 instead of u int32

MouazAlzahabi:


Multisig/Data/Services/Safe Client Gateway Service/BalancesRequest.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 
 struct BalancesRequest: JSONRequest {
     private let safeAddress: String
-    private let networkId: Int
+    private let networkId: String
```
DmitryBespalov, line 13: I regret that now we have multiple names interchangeably: network and chain, I will create a separate refactor ticket to align that

MultisigIntegrationTests/CoreData/NetworkTests.swift
```diff
@@ -42,22 +42,22 @@ class NetworkTests: CoreDataTestCase {
...
     }
 
     func test_by() {
-        var network = Network.by(1)
+        var network = Network.by("1")
         XCTAssertNil(network)
         _ = try? makeNetwork(id: 1)
 
-        network = Network.by(2)
+        network = Network.by("2")
```
DmitryBespalov, line 57: just a thought, the better design decision would be to model the chain id type as a separate type, ChainID, and that would have less changes throughout the code base.

Nothing to do because of this comment now :)

Multisig/Data/Services/Safe Client Gateway Service/SCGModels.swift
```diff
@@ -539,9 +539,13 @@ extension SCGModels {
...
         let theme: Theme
         let ensRegistryAddress: AddressString?
+
+        var id: String {
+            chainId.description
+        }
         
-        var id: Int {
-            Int(chainId.value)
+        var idValue: UInt256 {
```
sche: do we need both id and idValue?

Multisig/Data/Services/Safe Client Gateway Service/BalancesRequest.swift
sche, line 13: Would it be more explicit to use UInt256String here?

sche: and in other places

MultisigIntegrationTests/CoreData/NetworkTests.swift
```diff
@@ -230,19 +229,19 @@ class NetworkTests: CoreDataTestCase {
         XCTAssertEqual(networkSafes[0].safes[1].name, "22")
         XCTAssertEqual(networkSafes[0].safes[2].name, "20")
 
-        XCTAssertEqual(networkSafes[1].network, network1)
-        XCTAssertEqual(networkSafes[1].safes.count, 2)
-        XCTAssertEqual(networkSafes[1].safes[0].name, "01")
-        XCTAssertEqual(networkSafes[1].safes[1].name, "00")
+        XCTAssertEqual(networkSafes[2].network, network1)
```
sche: Why the logic was changed here?

MouazAlzahabi: The unit testing was failing with out of range exception

Multisig/Data/Services/Safe Client Gateway Service/BalancesRequest.swift
MouazAlzahabi: This is same as using the String to represent safeAddress instead of Address in the above line

sche: Yes, probably this is for the refactoring ticket that Dima mentioned

sche (approved):

Issue #1299, opened 2021-07-07
[iOS]No Network Ribbon during add safe flow


liliya-soroka:
**Bug description**
No Ribbon during add safe flow
**Steps To Reproduce**
1. Start add safe flow
2. Select network
3. Check each screen related to the safe loading 

Current result: There is no network ribbon ( except QR code screen)

**Expected Result**
Add selected network ribbon on the safe loading screens

**Device & App version**
 - Device/OS: iPhone 11/14.3
 - App version: 2.19.0 (488)
 
**Environment**
staging
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 2.19.0 (494)

PR #1303, opened 2021-07-08
gh-1299 Showing ribbon during add safe

DmitryBespalov:
Handles #1299 

- Using `Network` instead of SCG network in the intermediate screens
- Some adjustments to the ribbon vc

### DIFF
--- a/Multisig/UI/Safe Management/Add Safe/SelectNetworkViewController/SelectNetworkViewController.swift
+++ b/Multisig/UI/Safe Management/Add Safe/SelectNetworkViewController/SelectNetworkViewController.swift
@@ -184,8 +184,10 @@ class SelectNetworkViewController: LoadableViewController, UITableViewDelegate, 
         tableView.deselectRow(at: indexPath, animated: true)
         let vc = EnterSafeAddressViewController()
         vc.completion = completion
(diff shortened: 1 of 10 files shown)


Multisig/UI/Safe Management/Add Safe/SelectNetworkViewController/SelectNetworkViewController.swift
```diff
@@ -184,8 +184,10 @@ class SelectNetworkViewController: LoadableViewController, UITableViewDelegate,
         tableView.deselectRow(at: indexPath, animated: true)
         let vc = EnterSafeAddressViewController()
         vc.completion = completion
-        vc.network = model.models[indexPath.row]
-        show(vc, sender: self)
+        vc.network = Network.createOrUpdate(model.models[indexPath.row])
```
MouazAlzahabi: If add safe flow were canceled, this will introduce new network stored in the database with no safes, which we want to avoid

sche: +1

sche (approved):

MouazAlzahabi (approved):

Issue #1221, opened 2021-06-07
[iOS]Unstoppable domain: Proper error handling ( no internet connection) 


liliya-soroka:
**Bug description**
Long message without information about problem with internet connection is displayed
**Steps To Reproduce**
1. Run app and turn internet connection off
2. Add safe using unstoppable domain (udtestdev-creek.crypto for rinkeby)

**Expected Result**
the same error handling as for ENS

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/fc7ae7a6-227e-4e04-b215-f6ab57e0831c)
**Device & App version**
 - Device/OS: Iphone 11/13.4
 - App version: 2.17.0 (462)
 
**Environment**
staging, rinkeby
**Additional info**
(Optional)

sche:
@liliya-soroka I checked - there is no easy way to make it better. This error comes from 3rd party library that does not handle system errors (like no internet connection) properly, and we don't want to "guess" what could be potential reason of it. So I would keep it as it is.

liliya-soroka:
hm, is it not from the unstoppable domain implementation? I see that on android the proper message is handled. ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/17501563-57b1-4c62-aba6-27a58ca020c7) -   Could they do the same on ios?

sche:
iOS and Android libs are different. So in Android lib it seems like they have a custom "No internet" error, but in iOS lib they don't.
Could you put it a minor label and I will later create an issue in their repo to adjust this?

DmitryBespalov:
@liliya-soroka fixed in build 496

liliya-soroka:
Verified
app version 2.19.0 (500)

PR #1305, opened 2021-07-12
gh-1221 UD error can be handled with a custom network layer

DmitryBespalov:
Handles #1221 

Added error handling for 'unknown'/internal errors


Multisig/Logic/Ethereum/BlockchainDomainManager.swift
```diff
@@ -77,10 +78,95 @@ class BlockchainDomainManager {
...
             return GSError.ThirdPartyError(
                 reason: error.localizedDescription
             )
         }
     }
 }
+
+// Taken from the DefaultNetworkingLayer implementation to replace
+// the error handling with more descriptive error from HTTTPClient
+public struct GSNetworkingLayer: NetworkingLayer {
```
sche, line 105: would it be better to create a PR into their repo to return better errors?

DmitryBespalov: I wouldn't want to spend more time on it now, it's good enough

MouazAlzahabi (approved):

sche (approved):

Issue #1311, opened 2021-07-12
[iOS] The issue with balance page and native currency iew in transactions


liliya-soroka:
**Bug description**
The issue with balance page and native currency iew in transactions:
- error about wrong format is displayed on balance page
- 0 unknown is displayed instead of native network currents in transactions list and on tx details
**Steps To Reproduce**
1. add safe 0x1230B3d59858296A31053C1b8562Ecf89A2f888b on rinkeby
2. go to the balance page and on the transaction history(check the latest receive transaction)
Current result:
1. Error on balance page
2. 0 Unknow is displayed instead of amount+Eth
 
**Expected Result** The balance should be displayed together with valid info about txs with native currency 
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/066978ce-8065-4383-89e7-afcf8d449f30)

![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/0440fd75-e60a-438f-a5a4-4cca6480659a)

**Device & App version**
 - Device/OS: iPhone 11/14.3
 - App version: 2.19.0 (465)
 
**Environment**
staging app
**Additional info**
(Optional)

DmitryBespalov:
fixed in build 501

DmitryBespalov:
Verified 2.19.0 (532)

PR #1312, opened 2021-07-13
gh-1311 Replaced ether with native coin

DmitryBespalov:
Replaced the enums


Multisig/Logic/Models/TokenInfo.swift
```diff
@@ -9,16 +9,9 @@
 import Foundation
 
 struct TokenInfo: Decodable, Hashable {
-    let type: TokenType
```
DmitryBespalov, line 12: never used

sche (approved):

MouazAlzahabi (approved):

Issue #1223, opened 2021-06-07
Show contracts v1.3.0 as "up to date" in Safe Settings


DmitryBespalov:
Show the contract version "1.3.0 up to date" for safes with implementation addresses:
- GnosisSafe `0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552`
- GnosisSafeL2 `0x3E5c63644E683549055b9Be8653de26E0B4CD36E`

Note: 1.1.1 and 1.2.0 versions should still be shown as "up to date".

Depends on the backend issue https://github.com/gnosis/safe-client-gateway/issues/462

biafra23:
See: https://github.com/gnosis/safe-client-gateway/blob/interstellar/src/routes/about.rs#L35

Example (mainnet): https://safe-client.staging.gnosisdev.com/1/about/master-copies
For the endpoint and format for master copies per chain

DmitryBespalov:
Verified 2.19.0 (532)

liliya-soroka:
Opened question about 1.1.1 and 1.2.0 - https://github.com/gnosis/safe-client-gateway/issues/541

PR #1340, opened 2021-07-19
Gh 1223/show contracts v1.3.0 as "up to date" in safe settings

MouazAlzahabi:
Handles #1223


Multisig/Data/Services/Utils/String+Utils.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Moaaz on 7/19/21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+extension String {
+    func compareNumeric(_ value: String) -> ComparisonResult {
+        return compare(value, options: .numeric)
```
DmitryBespalov: I think we already have imported the "Version" swift package, I think this should be used in order to compare versions.

MultisigTests/Logic/Models/GnosisSafeTests.swift
```diff
@@ -10,57 +10,18 @@ import XCTest
...
-            "0xAC6072986E985aaBE7804695EC2d8970Cf7541A2",
-            "0x8942595A2dC5181Df0465AF0D7be08c8f23C93af",
-        ]
-
-        let unknown: [Address] = [
-            "0x3b1c2b0940C85458197E0D18690805d6F89547eE",
-            "0x976DC99c50B916Ea9b5275979059BCe9f1A0B1D1",
-            "0xD5D4763AE65aFfFD82e3aEe3ec9f21171A1d6e0e",
-        ]
+        let supported: [String] = ["1.0.0", "1.1.0", "1.2.3"]
```
DmitryBespalov: so the supported is just between 1.0.0 and 1.3.0?

In that case, tests should test the edge cases, inside the interval, and outside the interval.

- 1.0.0
- 1.3.0
- 1.3.1
- 1.4.0
- 0.1.0
- 1.1.1

DmitryBespalov: Well, actually, I would test with explicit specification the minimum and maximum version, so that the test won't break when we add/change the boundaries.

DmitryBespalov (approved):

Issue #1342, opened 2021-07-21
[iOS] "Guard" section should be shown for all configurations and can have "Not set" state


DmitryBespalov:
Pre previous discussion, we need to show the "guard" section in the safe settings and show it as "not set" similar to the fallback handler.

DmitryBespalov:
Verified 2.19.0 (532)

liliya-soroka:
https://gnosisinc.slack.com/archives/CQB729LLF/p1626708333410700 - It was decided that Guard should be shown only for the safes with master copy version >=1.3.0

PR #1347, opened 2021-07-22
gh-1342 always display guard info

sche:
Handles #1342

### DIFF
--- a/Multisig/Logic/Ethereum/GnosisSafe.swift
+++ b/Multisig/Logic/Ethereum/GnosisSafe.swift
@@ -17,8 +17,10 @@ class GnosisSafe {
         case unknown
     }
 
-    private var minimumSupportedVersion: Version { Version("1.0.0")! }
-    private var maximumSupportedVersion: Version { Version("1.3.0")! }
+    var minimumSupportedVersionValue = "1.0.0"
+    var maximumSupportedVersionValue = "1.3.0"
+    private var minimumSupportedVersion: Version { Version(minimumSupportedVersionValue)! }
+    private var maximumSupportedVersion: Version { Version(maximumSupportedVersionValue)! }
 
     func isSupported(_ version: String) -> Bool {
         if let version = Version(version), version >= minimumSupportedVersion && version <= maximumSupportedVersion {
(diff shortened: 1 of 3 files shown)


DmitryBespalov:
tests are failing

Multisig/Logic/Ethereum/GnosisSafe.swift
```diff
@@ -17,27 +17,8 @@ class GnosisSafe {
         case unknown
     }
 
-    var minimumSupportedVersionValue = "1.0.0"
-    var maximumSupportedVersionValue = "1.3.0"
-    private var minimumSupportedVersion: Version { Version(minimumSupportedVersionValue)! }
-    private var maximumSupportedVersion: Version { Version(maximumSupportedVersionValue)! }
```
DmitryBespalov, line 23: I don't understand why this was deleted. the variables were there for testability

sche: True, sorry

Multisig/UI/Transaction/TransactionDetailsViewController/TransactionDetailCellBuilder.swift
```diff
@@ -198,10 +198,7 @@ class TransactionDetailCellBuilder {
 
             case .setFallbackHandler(let fallbackTx):
                 let handler: Address = fallbackTx.handler.value.address
-                var (label, imageUri) = displayNameAndImageUri(addressInfo: fallbackTx.handler)
-                if label == nil {
-                    label = App.shared.gnosisSafe.fallbackHandlerInfo(AddressInfo(address: handler))?.name ?? "Not set"
```
DmitryBespalov, line 203: this removed the "Not set" functionality from the tx details, I also don't see where it is now implemented :)

sche: So the assumption for "Not set" would be that there is a `setFallbackHandler` tx with zero address?
I would expect the backend to return a label for this case. Or we could return 
```
if label == nil {
  label = handler.isZero ? "Not set" : "Unknown"
}
```
wdyt?

DmitryBespalov: yes, check for the label is good

DmitryBespalov (approved):

Issue #1349, opened 2021-07-24
[iOS] Data Loss: Owner Keys


fhoxh:
**Bug description**

Created or imported Owner Keys become deleted without notice.

**Steps To Reproduce**

1. Create New Owner Key or Import Existing Owner Key.
2. Observe successful creation or successful import, of new or existing Owner Key.
3. Observe that Owner Key is able to successfully participate in transaction signing.
4. Observe that some short time later, following unrelated interactions with iOS, the previously created or imported Owner Key is no longer present under Settings > Owner keys.

**Expected Result**

It is expected that created or imported Owner Keys remain present, until/unless explicitly deleted, manually, by user.

**Device & App version**

 - Device/OS: iPhone12Max/iOS15.0(19A5297e)
 - App version: Gnosis Safe 2.18.0

DmitryBespalov:
@fhoxh thanks for creating an issue. When you say "unrelated interactions with iOS", can we rule out that phone was restored from iCloud backup?

fhoxh:
> @fhoxh thanks for creating an issue. When you say "unrelated interactions with iOS", can we rule out that phone was restored from iCloud backup?

My pleasure! The device was not restored from backup. "Unrelated interactions" is intended to mean: using other apps; installing and deleting other apps; turning the device on and off; and other miscellaneous, unremarkable, commonplace activities. In terms of a foundational reproduction case, I would suggest creating a new Owner Key, force-quitting the app, relaunching to verify that the Owner Key is present (in my example, there was one and only one Owner Key), and then checking back one day later.

DmitryBespalov:
@fhoxh ok, thanks, we'll take a look

DmitryBespalov:
build 3.0.0 (3)

liliya-soroka:
verified
app version 3.0.0(6)

PR #1366, opened 2021-08-04
Gh 1349/data loss attempt1

DmitryBespalov:
Handles #1349

- Key cleanup on app startup seems to be the only place to suspect key removal. What could happen is that for some reason keychain is not accessible during the cleanup function and returned error. Then, because the error was treated as "no key exists", the key info could be deleted.
- Instead of ignoring errors, make the cleanup to only delete the keys for which private key is 100% does not exist after successful read operation.
- Also found a small tracking bug for the key number for the walletConnect

### DIFF
--- a/Multisig/Logic/Models/KeyInfo.swift
+++ b/Multisig/Logic/Models/KeyInfo.swift
@@ -207,12 +207,6 @@ extension KeyInfo {
         try all().forEach { try $0.delete() }
     }
 
-    /// Deletes keys with matching addresses
-    /// - Parameter addresses: addresses of keys to delete
-    static func delete(addresses: [Address]) throws {
-        try keys(addresses: addresses).forEach { try $0.delete() }
-    }
-
     /// Saves the key to the persistent store
     func save() {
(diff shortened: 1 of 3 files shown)


Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -207,12 +207,6 @@ extension KeyInfo {
         try all().forEach { try $0.delete() }
     }
 
-    /// Deletes keys with matching addresses
```
DmitryBespalov, line 210: not used

Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -30,10 +30,6 @@ extension KeyInfo {
         set { type = Int16(newValue.rawValue) }
     }
 
-    var hasPrivateKey: Bool {
-        (try? privateKey()) != nil
```
DmitryBespalov, line 34: `try?` may have caused the data loss

sche (approved):

Issue #1337, opened 2021-07-19
[iOS] [Ledger] Create a hardware database model


DmitryBespalov:
- New KeyInfo type
- Create method to create and store the connection data in the app
  - Think about 'version' for the data if it is stored in binary

liliya-soroka:
Verified
app version 3.1.0 (554)

PR #1382, opened 2021-08-12
Gh 1337 add key in database

sche:
Handles #1337

Import ledger device and store in the database

### DIFF
--- a/Multisig/Logic/Ledger/LedgerController.swift
+++ b/Multisig/Logic/Ledger/LedgerController.swift
@@ -15,21 +15,28 @@ class LedgerController {
         self.bluetoothController = bluetoothController
     }
 
-    func getAddress(deviceId: UUID, completion: @escaping (Address?) -> Void) {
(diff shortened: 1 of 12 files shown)


elgatovital (approved):

Multisig/Logic/Ledger/LedgerController.swift
```diff
@@ -8,17 +8,54 @@
...
+    }
+
+    typealias LedgerKeyInfoCompletion = (LedgerInfo?) -> Void
+
+    func getAddress(deviceId: UUID, at index: Int, completion: @escaping LedgerKeyInfoCompletion) {
+        guard let device = bluetoothController.deviceFor(deviceId: deviceId) else {
+            completion(nil)
+            return
+        }
+        let path = "44'/60'/0'/0/\(index)"
```
elgatovital, line 31: you should supply path as an argument

sche: it will be mitigated in following PRs :)

Issue #1402, opened 2021-08-30
Add "Enter name" screen for the Ledger keys


liliya-soroka:
Add "Enter name" screen for the Ledger keys

liliya-soroka:
Verified
app version 3.1.0 (554)

PR #1406, opened 2021-08-30
gh-1402 enter ledger name screen

sche:
Handles #1402

- Add "Enter ledger screen name" with the default name

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyPickerViewController.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyPickerViewController.swift
@@ -48,9 +48,30 @@ class LedgerKeyPickerViewController: SegmentViewController {
 
     @objc func didTapImport() {
         guard let selectedIndex = selectedIndex,
-              let controller = viewControllers[selectedIndex] as? LedgerKeyPickerContentViewController else { return }
-        controller.importSelectedKey()
-        completion()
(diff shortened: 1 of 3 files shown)


DmitryBespalov (approved):

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyPickerViewController.swift
```diff
@@ -48,9 +48,30 @@ class LedgerKeyPickerViewController: SegmentViewController {
...
-        completion()
+              let controller = viewControllers[selectedIndex] as? LedgerKeyPickerContentViewController,
+              let key = controller.selectedKey else { return }
+
+        var namePrefix = ""
+        switch controller.keyType {
+        case .ledger: namePrefix = "Ledger key #"
+        case .ledgerLive: namePrefix = "Ledger Live key #"
+        }
+        let defaultName = "\(namePrefix)\(key.index + 1)"
```
MouazAlzahabi, line 59: Do we need to have default name? 

I would be up to remove it in order to have same consistency with other import key types

MouazAlzahabi (approved):

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyPickerViewController.swift
sche: Here we can offer a good one, in other places, it is harder to do.
Now WalletConnect keys connected with default name even without this screeen

Issue #1428, opened 2021-09-09
[iOS] Move WalletConnect key feature out of experimental


sche:
Right now, the feature is not discoverable by a majority of users. We want to collect more data faster.
We will move it out of experimental with a possibility to return it.

liliya-soroka:
waiting for the app version with the task

liliya-soroka:
critical bug - [iOS] The app is crashed when key details for the connected via WC key is opened

liliya-soroka:
Verified
app version 3.2.0 (589)

PR #1439, opened 2021-09-14
gh-1428 move WalletConnect keys out of experimental

sche:
Handles #1428

For now, I comment out the code as we might need to return it. We will delete it when the requirements are ready and Android starts working on it.

### DIFF
--- a/Multisig/Logic/Models/KeyInfo.swift
+++ b/Multisig/Logic/Models/KeyInfo.swift
@@ -93,7 +93,7 @@ extension KeyInfo {
         let context = App.shared.coreDataStack.viewContext
         let fr = KeyInfo.fetchRequest().all()
         var items = try context.fetch(fr)
-        if !App.configuration.toggles.walletConnectOwnerKeyEnabled {
+        if /*!App.configuration.toggles.walletConnectOwnerKeyEnabled*/ false {
             items = items.filter { $0.keyType != .walletConnect }
         }
         return items
@@ -112,7 +112,7 @@ extension KeyInfo {
         return try addresses.compactMap { address in
             let fr = KeyInfo.fetchRequest().by(address: address)
             var items = try context.fetch(fr)
-            if !App.configuration.toggles.walletConnectOwnerKeyEnabled {
+            if /*!App.configuration.toggles.walletConnectOwnerKeyEnabled*/ false {
                 items = items.filter { $0.keyType != .walletConnect }
             }
             return items.first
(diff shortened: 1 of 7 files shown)


Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -93,7 +93,7 @@ extension KeyInfo {
         let context = App.shared.coreDataStack.viewContext
         let fr = KeyInfo.fetchRequest().all()
         var items = try context.fetch(fr)
-        if !App.configuration.toggles.walletConnectOwnerKeyEnabled {
+        if /*!App.configuration.toggles.walletConnectOwnerKeyEnabled*/ false {
             items = items.filter { $0.keyType != .walletConnect }
         }
```
DmitryBespalov: I would just comment all 3 lines instead of `if false`

Multisig/UI/Settings/OwnerKeyManagement/OwnerKeysListViewController/OwnerKeysListViewController.swift
```diff
@@ -140,7 +140,7 @@ class OwnerKeysListViewController: LoadableViewController, UITableViewDelegate,
 
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let keyInfo = keys[indexPath.row]
-        if App.configuration.toggles.walletConnectOwnerKeyEnabled {
+        if /*App.configuration.toggles.walletConnectOwnerKeyEnabled*/ true {
```
DmitryBespalov: I'd rather comment out the branches that are not executing, it's better readable that way instead of `/* ... */`

Multisig/UI/Settings/OwnerKeyManagement/ChooseOwnerKeyViewController/ChooseOwnerKeyViewController.swift
```diff
@@ -109,24 +109,24 @@ extension ChooseOwnerKeyViewController: UITableViewDelegate, UITableViewDataSour
...
+//        }
     }
 
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         tableView.deselectRow(at: indexPath, animated: true)
         let keyInfo = owners[indexPath.row]
 
         // For WalletConnect key check that it is still connected
-        if App.configuration.toggles.walletConnectOwnerKeyEnabled && keyInfo.keyType == .walletConnect {
+        if /*App.configuration.toggles.walletConnectOwnerKeyEnabled &&*/ keyInfo.keyType == .walletConnect {
```
DmitryBespalov: can we also use `//` here please (and in the line 168 below)

DmitryBespalov (approved):

Issue #1491, opened 2021-10-04
[iOS] The app is crashed on removing a key connected via WC ( MM)


liliya-soroka:
**Bug description**
The app is crashed on removing a key connected via WC ( MM) from Key details page

**Precondition**
Install MM mobile app on the same device where the safe mobile app

**Steps To Reproduce**
1.open app and connect MM owner using "Connect via WC"
2. Open key details for the connected key and click "Remove the key" option 
3. confirm remove dialog

**Current result** the app is crashed
**Expected Result** the key should be deleted without the crash 

Note!!!:  seems that the signing with MM doesn't work properly if the key was removed and readded: dialog with sign message doesn't appear

**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.2.0 (600)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.2.0 (611)

PR #1492, opened 2021-10-04
gh-1491 fix crash

sche:
Handles #1491


MouazAlzahabi (approved):

Multisig/UI/Transaction/TransactionDetailsViewController/TransactionDetailsViewController.swift
```diff
@@ -329,7 +329,9 @@ class TransactionDetailsViewController: LoadableViewController, UITableViewDataS
                 return
             }
 
-            self?.confirmAndRefresh(safeTxHash: safeTxHash, signature: signature, keyType: keyInfo.keyType)
+            DispatchQueue.main.async {
```
DmitryBespalov, line 332: do we need weak self here?

Issue #1472, opened 2021-09-28
[Delegates] Key Added Screen -  UI only


DmitryBespalov:
Screen to show after "Enter Key Name" to request enabling delegate key

### Wireframes (alternatives)
![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/4e0e29b5-04f6-4867-bfde-bc88b632e1e3)

liliya-soroka:
Verified
app version 3.4.0 (690)

PR #1499, opened 2021-10-05
gh-1472 Refactor: pull up class + Added stub UI

DmitryBespalov:
Handles #1472 (work in progress)

- Extracted super class from the Safe Added VC to reuse in the Key Added VC later
- Added stub UI for the ledger key added vc.


Multisig/UI/Safe Management/Add Safe/SafeLoadedViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+        titleText = "Load Gnosis Safe"
+        headerText = "Safe loaded"
+        primaryActionName = "Add owner key"
+        secondaryActionName = "Skip"
+
+        super.viewDidLoad()
+    }
+
+    override func primaryAction(_ sender: Any) {
+        Tracker.trackEvent(.userOnboardingOwnerAdd)
```
MouazAlzahabi, line 37: for tracking I would define "trackingEvent" in base class and here we just assign the value of it

MouazAlzahabi (approved):

sche (approved):

Multisig/UI/Safe Management/Add Safe/SafeLoadedViewController.swift
DmitryBespalov: I'll handle it as part of a separate issue for tracking, leaving as is for now

PR #1502, opened 2021-10-06
gh-1472 Added simulated ledger

DmitryBespalov:
Added option to add ledger key while in simulator.

The bluetooth controller is mocked and simulated
resposne for "get address" function is added.

Handles #1472 (work in progress)

- Polymorphic Bluetooth controller (via inheritance)
- Added simulated bluetooth response for ledger device - only get address is returned
- Fixed a hidden bug with the `data`: before it was accessing the underlying data slice and it was a very weird code


MouazAlzahabi (approved):

sche:
lgtm!
Did you double-check it with a real device that it still works?

sche:
I checked signing in this branch - it worked

Issue #1495, opened 2021-10-04
[iOS] Found devices are not displayed as a list


liliya-soroka:
**Bug description**
Found devices are not displayed as a list. The user has to reload list to replace one found device by another one 
**Steps To Reproduce**
1. Turn on a few Ledger nano X devices
2. try to add key or confirm tx

Current result: Always only one device is displayed, to see the next device the user has to refresh the page. After refreshing the next found device replaced the previous one in the list 

**Expected Result**
The list all found Ledger nano X device should be displayed instead of just one entity

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 14.8 (600)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
Verified by the team 
Two Ledger devices were listed on one page as a list

PR #1511, opened 2021-10-07
gh-1495 no message

sche:
Handles #1495

We stop scanning for devices once we have at least one device found.
I removed it. We need to stop it manually or it will be automatically stopped by releasing BluetoothController


MouazAlzahabi (approved):

Multisig/Logic/Ledger/BluetoothController.swift
```diff
@@ -95,8 +95,6 @@ extension BluetoothController: CBCentralManagerDelegate {
             devices.append(device)
             delegate?.bluetoothControllerDidDiscover(device: device)
         }
-
-        centralManager.stopScan()
```
DmitryBespalov, line 99: is there a way for a device to disappear? here we always appending

DmitryBespalov (approved):

Issue #1540, opened 2021-10-28
Add/Import/export address book entities 


MouazAlzahabi:
We need to add the ability to Add/Import/export entities to address book

liliya-soroka:
Reopened - The task can not be tested and approved till the issue with importing is fixed
https://app.zenhub.com/workspaces/mobile-development-kanban-60cc78740f488d000e996335/issues/gnosis/safe-ios/1550
app version 3.3.0 (651)

liliya-soroka:
Verified
app version 3.3.0 (659)

Issue #1537, opened 2021-10-28
Address book list 


MouazAlzahabi:
We need to create Address book list screen categorized by network

liliya-soroka:
Verified
app version 3.3.0 (651)

PR #1543, opened 2021-10-28
Gh 1540/add import export address book entities

MouazAlzahabi:
Handles #1537  #1540


Multisig/Logic/FileManagerWrapper.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+import Foundation
+import UIKit
+
+class FileManagerWrapper {
+    static func export(text: String, fileName: String, fileExtension: String) -> URL? {
+        let exportFilePath = NSTemporaryDirectory() + "\(fileName).\(fileExtension)"
+        let exportFileURL = URL(fileURLWithPath: exportFilePath)
+        FileManager.default.createFile(atPath: exportFilePath, contents: NSData() as Data, attributes: nil)
```
DmitryBespalov: dummy question: what is advantage of using file handle API over reading/writing `Data(contentsOf:options:)`?

Multisig/Logic/Models/Safe.swift
```diff
@@ -187,12 +187,7 @@ extension Safe: Identifiable {
 
         let chainId = safe.chain!.id!
 
-        if safe.chain!.safe!.count == 1 {
-            // remove chain with associated safe
-            Chain.remove(chain: safe.chain!)
```
DmitryBespalov, line 192: just curious why this code is deleted

Multisig/UI/Settings/AddressBook/AddressBookListTableViewController.swift
```diff
@@ -0,0 +1,223 @@
...
+        alertController.addAction(cancelButton)
+
+        self.present(alertController, animated: true)
+    }
+
+    private func didTapAddButton() {
+        let vc = SelectNetworkViewController()
+        vc.screenTitle = "New Entry"
+        vc.descriptionText = "Select network on which you want to add entry:"
+        vc.completion = { [unowned self] chain  in
```
DmitryBespalov: just checking: `unowned self` - address book list is guaranteed to exist when the completion called, right?

Multisig/UI/Settings/AddressBook/AddressBookListTableViewController.swift
```diff
@@ -0,0 +1,223 @@
...
+            let vc = CreateAddressBookEntryViewController()
```
DmitryBespalov, line 102: I'd name this `createEntryVC`

Multisig/UI/Settings/AddressBook/AddressBookListTableViewController.swift
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        let cancelButton = UIAlertAction(title: "Cancel", style: .cancel, handler: nil)
+        alertController.addAction(cancelButton)
+
+        self.present(alertController, animated: true)
+    }
+
+    private func didTapAddButton() {
+        let vc = SelectNetworkViewController()
```
DmitryBespalov: I'd name this `selectNetworkVC` for less ambiguity

Multisig/Logic/Models/Chain.swift
```diff
@@ -222,6 +222,30 @@ extension Chain {
 
         return chainSafes
     }
+
+    typealias ChainEntries = [(chain: Chain, entries: [AddressBookEntry])]
+
+    /// Returns safes grouped by chain with the following logic applied:
```
DmitryBespalov: is this comment valid? It says something about safes, but the code is about address book entries

Multisig/Logic/Models/Safe.swift
MouazAlzahabi: because now chain could contains address book entries, so it's not correct to remove a chain when there is no safes related

Multisig/UI/Settings/AddressBook/CreateAddressBookEntryViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+            addressField.setError("Address should not be empty")
+            return
+        }
+        addressField.setInputText(text)
+        do {
+            // (1) validate that the text is address
+            let address = try Address(text, isERC681: true)
+            addressField.setAddress(address)
+
+            // (2) and that there's no such safe already
```
DmitryBespalov: comment is about safe, but the code is about address book entry

DmitryBespalov (approved):

Multisig/Logic/FileManagerWrapper.swift
MouazAlzahabi: good point, no advantage actually, I will change

Multisig/UI/Settings/AddressBook/AddressBookListTableViewController.swift
MouazAlzahabi: right

Issue #1556, opened 2021-11-04
[iOS] Refactor browse address on chain feature 


MouazAlzahabi:
The current implementation of browse address on chain feature is supposed to have selected safe, the problem appeared when there is no selected safe in screens such as Address book list. 

We need to update the way how we implement this in order to accept address+chain info.

liliya-soroka:
Verified during the prefix testing 

app version 3.4.0 (684)

PR #1578, opened 2021-11-10
Gh 1530/eip 3770  add network prefix to the address

MouazAlzahabi:
Handles #1530 #1556


Multisig/Logic/String+Highlight.swift
```diff
@@ -0,0 +1,36 @@
...
+//  Multisig
+//
+//  Created by Moaaz on 11/10/21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+extension String {
```
DmitryBespalov: this file needs to be in UI Library, not in the "logic"

Multisig/UI/UI Library/Views/AddressInfoView.swift
```diff
@@ -80,13 +89,26 @@ class AddressInfoView: UINibView {
...
+    @objc func displayAddress() {
+        if let _ = label {
+            addressLabel.text = prefixString() + self.address.ellipsized()
+        } else {
+            let prefixString = prefixString()
+            addressLabel.attributedText = (prefixString + self.address.checksummed).highlight(prefix: prefixString.count + 4)
+        }
+    }
+
+    private func prefixString() -> String {
```
DmitryBespalov, line 102: should this be extracted and reused? seems repeated several times

DmitryBespalov (approved):

Multisig/UI/UI Library/Views/AddressInfoView.swift
MouazAlzahabi: Actually what should be done is refactor AddressInfoView + SafeBarView, as they are almost identical, but will keep this for later

Issue #1517, opened 2021-10-11
[Delegates] Integrate backend endpoints


DmitryBespalov:
Integrate backend when it is ready

liliya-soroka:
Verified
app version 3.4.0 (720)

PR #1582, opened 2021-11-11
Gh 1517 integrate delegates backend

biafra23:
Handles #1517

- Activate Delegate screen after ledger address import
- Add chainId to CreateDelegateRequest
- Modify endpoint for CGW
- create delegate for all chains

**Known problems**
- When I run his from the IDE there is an exception(?) a little while after the network request has been sent. I'd guess this is a crash. But I can't reproduce it without the IDE being connected.

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyAddedViewController.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyAddedViewController.swift
@@ -193,13 +193,15 @@ class AddDelegateKeyController {
     }
 
     func sendToBackend(delegateAddress: Address, signature: Data, completion: @escaping (Result<Void, Error>) -> Void) {
-        clientGatewayService.asyncCreateDelegate(safe: nil,
-                                                 owner: ownerAddress,
-                                                 delegate: delegateAddress,
-                                                 signature: signature,
-                                                 label: "iOS Device Delegate",
-                                                 chainId: "4") { result in
-            completion(result.map { _ in () })
+        Chain.all.forEach { chain in
+            clientGatewayService.asyncCreateDelegate(safe: nil,
+                    owner: ownerAddress,
+                    delegate: delegateAddress,
+                    signature: signature,
+                    label: "iOS Device Delegate",
+                    chainId: chain.id!) { result in
+                completion(result.map { _ in () })
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyAddedViewController.swift
```diff
@@ -193,12 +193,29 @@ class AddDelegateKeyController {
...
+                    owner: ownerAddress,
+                    delegate: delegateAddress,
+                    signature: signature,
+                    label: "iOS Device Delegate",
+                    chainId: chain.id!) { result in
+                group.leave()
+            }
+        }
+        // wait until all requests finish, then call completion
+        let timeoutResult = group.wait(timeout: .now() + .seconds(60))
```
sche: should we move the magic number to a variable?

sche (approved):

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyAddedViewController.swift
biafra23: Done in 3c6ddd62a6385fae46f4c467ed566ae18f6daf1e

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyAddedViewController.swift
```diff
@@ -193,12 +193,30 @@ class AddDelegateKeyController {
...
+                    owner: ownerAddress,
+                    delegate: delegateAddress,
+                    signature: signature,
+                    label: "iOS Device Delegate",
+                    chainId: chain.id!) { result in
+                group.leave()
+            }
+        }
+        // wait until all requests finish, then call completion
+        let createDelegateRequestTimeoutInSeconds = 60 // one minute
```
DmitryBespalov, line 214: The name and comment do not add any meaning why is it 60 and not, say 49. Let's add the comment here `We use 60 seconds because it's a URLRequest's default timeout and we expect all requests to finish before that`.

And later we need to adjust the UI to show activity indicator or a progress bar to the user with the progress of requests (in a separate ticket and pull request)

MouazAlzahabi: Fixed

Issue #1584, opened 2021-11-15
[iOS] Support EIP-3770 address in address input fields


MouazAlzahabi:
We need to support EIP-3770 addresses in all address input fields, this includes: 

- QR scanners
- Paste from clipboard 

If address is not valid we display the usual invalid address error, if network not mach the selected network we display network mismatch error.

liliya-soroka:
Verified
app version 3.4.0 (684)

PR #1592, opened 2021-11-18
scan qr address with prefix

MouazAlzahabi:
Handles #1584

### DIFF
--- a/Multisig/Data/Services/Utils/GSError.swift
+++ b/Multisig/Data/Services/Utils/GSError.swift
@@ -242,6 +242,15 @@ enum GSError {
         let loggable = false
     }
 
+    struct AddressMismatchNetwork: DetailedLocalizedError {
+        let description = "Can’t use this address"
+        let reason = "This address is not matching the selected network."
+        let howToFix = "Please use different address or switch network"
+        let domain = clientErrorDomain
+        let code = 1104
+        let loggable = false
+    }
+
     struct KeyAlreadyImported: DetailedLocalizedError {
         let description = "Can't use this private key"
         let reason = "This key already imported."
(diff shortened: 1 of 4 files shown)


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -242,6 +242,15 @@ enum GSError {
         let loggable = false
     }
 
+    struct AddressMismatchNetwork: DetailedLocalizedError {
+        let description = "Can’t use this address"
+        let reason = "This address is not matching the selected network."
+        let howToFix = "Please use different address or switch network"
```
sche: "Please use **a** different address or switch network" - checked with Grammarly :)

Shared/Utils/Address.swift
```diff
@@ -95,13 +95,11 @@ struct Address: Hashable, ExpressibleByStringInterpolation, CustomStringConverti
...
 
-    init(_ value: String, isERC681: Bool) throws {
-        var addressString = value
-        if isERC681 {
-            addressString = Self.addressFromERC681(addressString)
-        }
-
-        _store = try EthereumAddress(hex: addressString, eip55: false)
+    // This will check if ERC681 or EIP3770
+    static func addressWithPrefix(text: String) throws -> (prefix: String?, address: Address) {
```
sche: Would it be better to make prefix as a property of `Address`?

sche: I am also not sure how do we add prefixes to addresses where we need to present them in the UI. I would assume it to be a part of the Address description if it has a prefix set.

MouazAlzahabi: as prefix is the short name of corresponding network, is taken from there now

PR #1595, opened 2021-11-18
Add and integrate Intercom SDK with separate app id and api key.

biafra23:
Handles safe#500

- Unswiffle Firebase APNS handling
- registerForRemoteNotifications() on evry app start
- Integrate Intercom APNS handling
- ...

### DIFF
--- a/Multisig/App/SceneDelegate.swift
+++ b/Multisig/App/SceneDelegate.swift
@@ -56,13 +56,6 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             handleUserActivity(userActivity)
         }
 
-        Intercom.setApiKey(App.configuration.services.intercomApiKey, forAppId: App.configuration.services.intercomAppId)
-        Intercom.setLauncherVisible(true)
-        
-        #if DEBUG
-        Intercom.enableLogging()
-        #endif
-        Intercom.registerUnidentifiedUser()
     }
 
     func sceneWillEnterForeground(_ scene: UIScene) {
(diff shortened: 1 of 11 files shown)


Multisig/App/SceneDelegate.swift
```diff
@@ -54,6 +55,14 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
             handleUserActivity(userActivity)
         }
+
+        Intercom.setApiKey(App.configuration.services.intercomApiKey, forAppId: App.configuration.services.intercomAppId)
+        Intercom.setLauncherVisible(true)
+        
+        #if DEBUG
+        Intercom.enableLogging()
+        #endif
+        Intercom.registerUnidentifiedUser()
```
DmitryBespalov: could you please extract all this intercom-related setup into a separate class similar to `FirebaseConfig.swift` and then move the setup call from this `SceneDelegate` to `AppDelegate` instead?

Multisig/Info.plist
```diff
@@ -63,6 +63,10 @@
 	<string>$(HELP_CENTER_URL)</string>
 	<key>INFURA_API_KEY</key>
 	<string>$(INFURA_API_KEY)</string>
+	<key>INTERCOM_APP_ID</key>
```
DmitryBespalov, line 66: I see these variables here, it's good! Can you also add the example values to the `Config.Example.xcconfig` so that we don't forget to set these variables?

DmitryBespalov: we also need to provide these variables in our CI environment, I'll help you with that

biafra23:
Branch has been renamed. Closing this draft

PR #1611, opened 2021-11-25
Load TestingAppDelegate instead of AppDelegate for unit tests

elgatovital:


DmitryBespalov:
I'm a bit confused of why this is needed, could you explain more please?

elgatovital:
by default, AppDelegate is loaded before unit tests are executed with all its dependencies. Having TestingAppDelegate will ensure a cleaner test setup without a boilerplate that is not needed for unit testing.

DmitryBespalov (approved):

Issue #1516, opened 2021-10-11
[Delegates] Error Handling


DmitryBespalov:
Unify error-handling for the delegates feature so that it adheres to our error  handling requirements.

Create new errors and messages where needed, document it in the spec doc.

DmitryBespalov:
Added errors 1320 - 1323 (https://docs.google.com/document/d/1foQ9DSMgO2iN-6u4JsJIDJWB7mUHX-D8ApZWvx6oKAo/edit#)

liliya-soroka:
verified
app version 3.4.0 (735)

PR #1615, opened 2021-11-25
gh-1516 Added error types

DmitryBespalov:
Handles #1516 

- Converted strings to errors

### DIFF
--- a/Multisig/Data/Services/Utils/GSError.swift
+++ b/Multisig/Data/Services/Utils/GSError.swift
@@ -416,9 +416,8 @@ enum GSError {
         let loggable = false
     }
 
-    // MARK: Delegate Key Errors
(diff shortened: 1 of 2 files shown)


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -416,6 +416,8 @@ enum GSError {
         let loggable = false
     }
 
+    // MARK: Delegate Key Errors
+
 	// MARK: Address Book
```
elgatovital, line 419: //MARK: Delegate Key Errors should be moved closer to errors related to delegate key

DmitryBespalov: true

Issue #1655, opened 2021-12-07
[Intercom] Adjust App Settings


sche:
- [x] Add icons and sections
- [x] Renaming "Get In Touch" to "Help Center"
- [x] "Chat with us" opens Intercom in a popover screen

**TODO**
Add links to screenshots

elgatovital:
Get In Touch -> Get Support according to screens in figma
Itercom wiring can be done after https://github.com/gnosis/safe-ios/pull/1660 is merged

liliya-soroka:
Nothing happens on Chat with us click

app version 3.5.0 (732) and 3.4.0 (743)

iPhone 11/14.8

liliya-soroka:
Verified
app version 3.5.0 (741)

PR #1665, opened 2021-12-08
Gh 1655/Adjust app settings

elgatovital:
Handles #1655

- Add icons for app settings items
- Adjust app settings menu
- Add separate page for terms, privacyPolicy, licenses, rateTheApp

simulator_screenshot_8C889669-26FD-45E5-96FD-0F531E4C6D53
simulator_screenshot_7DDEEDA8-B17A-41D7-A37B-D113ED3836FA


lukasschor:
Why is chain prefix not part of appearance like on web?

Also thinking "Fiat currency" might be well place in appearance as well (on web there is no comparison as it's not even in setttings). But less of a strong opinion there. Definitely think fiat currency is too prominent for the fact that this is something users usually set once and then never again. :)

biafra23:
Looks like the left margin of the `App version` is not the same as on the other cells. 

app-version-margin

Other than that it looks great, But I did not do a proper code review.

biafra23:
@lukasschor Appearance is only about dark mode ATM. If we change this, it should be properly planned and IMHO not be part of this PR. Or even this ticket.

elgatovital:
> Looks like the left margin of the `App version` is not the same as on the other cells.
> Other than that it looks great, But I did not do a proper code review.

Fixed: cells had different left margin

elgatovital:
> Why is chain prefix not part of appearance like on web?
> 
> Also thinking "Fiat currency" might be well place in appearance as well (on web there is no comparison as it's not even in setttings). But less of a strong opinion there. Definitely think fiat currency is too prominent for the fact that this is something users usually set once and then never again. :)

@sche, @TanyaEfremova fresh feedback regarding app settings

lukasschor:
> @lukasschor Appearance is only about dark mode ATM. If we change this, it should be properly planned and IMHO not be part of this PR. Or even this ticket.

Yes didn't mean this should be changed in this PR/ticket, just wanted to share an opinion based on the screenshot. Next time I'll provide the feedback on slack instead.

TanyaEfremova:
@lukasschor Agreed, a few items in the app settings should be grouped logically, but as @biafra23 said, it is not yet a part of this ticket, but it's a part of another task that is in backlog for now.

biafra23 (approved):
Meinen Segen hast Du :-) 👍🏻

DmitryBespalov (approved):

Issue #1718, opened 2021-12-20
[iOS] If the first safe tx is initiated using WalletConnect on mobile, it will fail


sche:
**Related backend issue:**
https://github.com/gnosis/safe-client-gateway/issues/757

liliya-soroka:
Verified
app version 3.6.0 (769)

PR #1719, opened 2021-12-20
gh-1718 fix the fist tx initiated with WalletConnect on mobile issue

sche:
Handles #1718

### DIFF
--- a/Multisig/Logic/Models/Transaction.swift
+++ b/Multisig/Logic/Models/Transaction.swift
@@ -79,7 +79,8 @@ extension Transaction {
                     data: data?.data,
                     operation: operation)
 
-            self.nonce = UInt256String(estimationResult.latestNonce.value + 1)
+            self.nonce = UInt256String(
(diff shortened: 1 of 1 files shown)


Multisig/Logic/Models/Transaction.swift
```diff
@@ -79,7 +79,8 @@ extension Transaction {
                     data: data?.data,
                     operation: operation)
 
-            self.nonce = UInt256String(estimationResult.latestNonce.value + 1)
+            self.nonce = UInt256String(
+                estimationResult.latestNonce.value == 0 ? 0 : estimationResult.latestNonce.value + 1)
```
MouazAlzahabi, line 83: I would like to check if there is "recommendedNonce" first, otherwise we do this logic

sche: As I understand it, `recommendedNonce` will be a part of the new endpoint.

MouazAlzahabi (approved):

Issue #1646, opened 2021-12-06
[TransferAssets] Token amount input component


sche:
The component should be used both for chain currency and ERC-20 tokens.

- [x] It is possible to specify the token amount to send manually.
- [x] It is possible to paste the token amount from a clipboard.
- [x] It is possible to enter only digits into the token amount field.
- [x] It has token logo.

**Error handling**
Invalid token amount cases:
- [x] [Insufficient funds](https://www.figma.com/file/BoikwWyX5nBgbfcaBt5wdK/iOS-Initiate-assets-transfer?node-id=373%3A10387)
- [x] [The amount should be greater than 0](https://www.figma.com/file/BoikwWyX5nBgbfcaBt5wdK/iOS-Initiate-assets-transfer?node-id=556%3A6939)
- [x] [The number of decimals exceeds allowed maximum](https://www.figma.com/file/BoikwWyX5nBgbfcaBt5wdK/Initiate-assets-transfer?node-id=717%3A8807)

![image.png](https://images.zenhubusercontent.com/5c62952584111d4401fa50d5/d50bf3e1-b131-459d-987b-9ea17044b811)

liliya-soroka:
Blocked for testing

liliya-soroka:
Reopened:
no amount validation at all 

app version 3.6.0 (796)

liliya-soroka:
Verified
app version 3.6.0 (819)
the issue with send max and validation is reported separately

PR #1722, opened 2021-12-20
gh-1646 Token amount component

elgatovital:
Handles #1646

- Add component for displaying & editing the token amount

### DIFF
--- a/Multisig/UI/UI Library/Views/TokenAmountField.swift
+++ b/Multisig/UI/UI Library/Views/TokenAmountField.swift
@@ -9,6 +9,8 @@
 import UIKit
 
 class TokenAmountField: UINibView {
+    
+    var onTap: () -> Void = { }
 
     @IBOutlet weak var iconImage: UIImageView!
     @IBOutlet weak var amountLabel: UITextField!
@@ -17,5 +19,25 @@ class TokenAmountField: UINibView {
     
     override func commonInit() {
         super.commonInit()
+    }
+    
+    override func awakeFromNib() {
+        symbolLabel.setStyle(.tertiary)
+        amountLabel.setStyle(.primary)
+        amountLabel.placeholder = "Amount"
+        errorLabel.setStyle(.error)
+        errorLabel.isHidden = true
+    }
+    
+    @IBAction func didTapField(_ sender: Any) {
+        onTap()
+    }
+    
+    @IBAction func didTouchUp(_ sender: Any) {
+        
+    }
+    
+    @IBAction func didTouchDown(_ sender: Any) {
+        
     }
 }
--- a/Multisig/UI/UI Library/Views/TokenAmountField.swift
+++ b/Multisig/UI/UI Library/Views/TokenAmountField.swift
@@ -1,0 +1,21 @@
+//
+//  TokenAmountField.swift
+//  Multisig
+//
+//  Created by Vitaly Katz on 20.12.21.
+//  Copyright © 2021 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class TokenAmountField: UINibView {
+
+    @IBOutlet weak var iconImage: UIImageView!
+    @IBOutlet weak var amountLabel: UITextField!
+    @IBOutlet weak var symbolLabel: UILabel!
+    @IBOutlet weak var errorLabel: UILabel!
+    
+    override func commonInit() {
+        super.commonInit()
+    }
+}
--- a/Multisig/UI/UI Library/Views/TokenAmountField.swift
+++ b/Multisig/UI/UI Library/Views/TokenAmountField.swift
@@ -21,11 +21,8 @@ class TokenAmountField: UINibView {
         set { amountTextField.text = newValue }
     }
     
-    override func commonInit() {
-        super.commonInit()
-    }
-    
     override func awakeFromNib() {
+        super.awakeFromNib()
         symbolLabel.setStyle(.tertiary)
         amountTextField.setStyle(.primary)
         amountTextField.placeholder = "Amount"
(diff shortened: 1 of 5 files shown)


Multisig/UI/UI Library/Views/TokenAmountField.swift
```diff
@@ -0,0 +1,78 @@
...
+    @IBOutlet weak var amountTextField: UITextField!
+    @IBOutlet weak var symbolLabel: UILabel!
+    @IBOutlet weak var errorLabel: UILabel!
+    
+    var balance: String {
+        get { amountTextField.text ?? "" }
+        set { amountTextField.text = newValue }
+    }
+    
+    override func commonInit() {
```
DmitryBespalov: you can remove this as it only calls `super`

Multisig/UI/UI Library/Views/TokenAmountField.swift
```diff
@@ -0,0 +1,78 @@
...
+        super.commonInit()
+    }
+    
+    override func awakeFromNib() {
+        symbolLabel.setStyle(.tertiary)
```
DmitryBespalov, line 26: something is missing: you must to call super.awakeFromNib() in this method

Multisig/UI/UI Library/Views/TokenAmountField.swift
```diff
@@ -0,0 +1,78 @@
...
+        errorLabel.isHidden = false
+        borderImage.tintColor = .error
+    }
+}
+
+// allow only decimal numbers to be entered in the amount field
+extension TokenAmountField: UITextFieldDelegate {
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        
+        let decimalSeparator = Locale.autoupdatingCurrent.decimalSeparator ?? "."
```
DmitryBespalov: decimal separator can be 1 of 3 different things (this is how keyboard works), please see more here: https://github.com/gnosis/SwiftCryptoTokenFormatter/blob/6f605fd7bd61327a6006919e28bf3f0632d1c556/SwiftCryptoTokenFormatter/SwiftCryptoTokenFormatter/TokenFormatter.swift#L14

Multisig/UI/UI Library/Views/TokenAmountField.swift
```diff
@@ -0,0 +1,78 @@
...
+        let inverseSet = CharacterSet(charactersIn:"0123456789").inverted
```
DmitryBespalov: there's a `decimalDigit` set already: https://developer.apple.com/documentation/foundation/characterset/1779641-decimaldigits#

Multisig/UI/UI Library/Views/TokenAmountField.swift
DmitryBespalov: and it doesn't adhere to the locale. I killed a lot of time in the past to find out this.

Multisig/UI/UI Library/Views/TokenAmountField.swift
```diff
@@ -0,0 +1,78 @@
...
+    func showError(message: String) {
+        errorLabel.text = message
+        errorLabel.isHidden = false
+        borderImage.tintColor = .error
+    }
+}
+
+// allow only decimal numbers to be entered in the amount field
+extension TokenAmountField: UITextFieldDelegate {
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
```
DmitryBespalov, line 50: will that allow pasting negative numbers?

DmitryBespalov: via copy-paste

Multisig/UI/UI Library/Views/TokenAmountField.swift
```diff
@@ -0,0 +1,78 @@
...
+        
+        let decimalSeparator = Locale.autoupdatingCurrent.decimalSeparator ?? "."
+        let inverseSet = CharacterSet(charactersIn:"0123456789").inverted
+        let components = string.components(separatedBy: inverseSet)
+        let filtered = components.joined(separator: "")
+        if filtered == string {
+            return true
+        } else {
+           if string.contains(decimalSeparator) {
+           let countdots = textField.text!.components(separatedBy: decimalSeparator).count - 1
```
DmitryBespalov: please handle the optional  properly, this actually might crash here I think

Multisig/UI/UI Library/Views/TokenAmountField.swift
elgatovital: no. But we also don't want negative amounts, right?

DmitryBespalov: right, just making sure that it should not accept negative numbers even if I copy-paste a negative number

MouazAlzahabi (approved):

Multisig/UI/UI Library/Views/TokenAmountField.swift
elgatovital: decimal separator used should adhere to locale according to https://github.com/gnosis/safe/wiki/How-to-format-amounts. 
Using Locale.current.decimalSeparator

DmitryBespalov: it doesn't, there are combinations that even if locale is set to one value, the actual decimal separator that keyboard produces is something else

DmitryBespalov: try simulating different locales + regions: US, German, UK, Japanese, Korean, French and see which decimal separator characters you're getting

elgatovital: I have tried different locale/region combinations and each time I was getting the expected decimal separator. The same separator also was used in amounts displayed in the balances list.

elgatovital: separator displayed in the decimalPad keyboard was adhering to locale in my tests.

DmitryBespalov (approved):

PR #1726, opened 2021-12-21
Adjust splash screen to use new christmas image

elgatovital:
image


sche:
@elgatovital the app is already in review. I would not make a new submission at this point.

Issue #1640, opened 2021-12-06
[WhatsNew] What's New screen


DmitryBespalov:
Implement the "What's New" screen.

Design of icons and texts is to be defined.

For the v1 of this feature, use the "[WhatsNewKit](https://github.com/SvenTiigi/WhatsNewKit)" library and store the icons and texts directly in the app.

Enable the screen for the upcoming 3.6.0 release with Initiate Tx, Execute Tx, and Push Notifications for WalletConnect keys features.

Design: [Figma](https://www.figma.com/file/NynYLm0HPB49Q8tApebw0n/What's-new?node-id=2%3A1237)

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/7eb99f1c-b834-4323-a7dd-0de0566f4d68)

lukasschor:
Did you evaluate the Intercom Carousels for this? https://www.intercom.com/mobile-carousels

DmitryBespalov:
@lukasschor not yet, however based on the Tanya's experience with Intercom, we decided not to go for this right now because it's more complicated than it seems.

biafra23:
@TanyaEfremova I have a question regarding the `Read notes` button in the design. Where does that lead to?

biafra23:
@sche Please review and modify the text in https://www.figma.com/file/NynYLm0HPB49Q8tApebw0n/What's-new?node-id=0%3A1

TanyaEfremova:
@biafra23 The button is **only** displayed when when there are release notes that are externally available for reading, otherwise - don't display the link button. I think it should be added to the requirements

biafra23:
@sche @TanyaEfremova Will we have the read more link for the 3.60 release? And if so, what is the URL for the release notes?

sche:
@biafra23 no need in "Read more" link

liliya-soroka:
Verified
app version 3.6.0 (769)

PR #1729, opened 2021-12-22
Whats new

biafra23:
Handles #1640

- Add WhatsNewKit dependency
- Add text and images to screen
- Show screen once per version

**Missing** 
- I was not able to add the space between item title and subtitle
- Also I am not 100% sure about the colors 
- Dark mode support (Design missing)

Implementation | Design
--------------- | -------
image | image


Multisig/UI/App/WhatsNewHandler.swift
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        whatsNewViewController = WhatsNewViewController(
+                whatsNew: whatsNew,
+                configuration: configuration,
+                versionStore: InMemoryWhatsNewVersionStore() // keyValueVersionStore // persist
+        )
+    }
+
+    func getViewController() -> WhatsNewViewController? {
```
sche: `whatsNewViewController` is not private. I think no need for the getter

Multisig/UI/App/WhatsNewHandler.swift
```diff
@@ -0,0 +1,63 @@
...
+        configuration.completionButton.title = "Let's go"
+
+        let keyValueVersionStore = KeyValueWhatsNewVersionStore(
+                keyValueable: UserDefaults.standard
+        )
+
+        whatsNewViewController = WhatsNewViewController(
+                whatsNew: whatsNew,
+                configuration: configuration,
+                versionStore: InMemoryWhatsNewVersionStore() // keyValueVersionStore // persist
```
sche: Is `InMemoryWhatsNewVersionStore` a part of the lib?

biafra23: Yes. It is. For debugging. It means the dismissal is not persisted. So the screen is shown every time.

Multisig/UI/App/WhatsNewHandler.swift
```diff
@@ -0,0 +1,62 @@
...
+        configuration.itemsView.titleColor = .darkText
+        configuration.itemsView.subtitleColor = .gray
+
+        configuration.detailButton?.titleColor = .button
+        configuration.completionButton.backgroundColor = .button
+        configuration.completionButton.title = "Let's go"
+
+        configuration.itemsView.autoTintImage = false
+
+        let keyValueVersionStore = KeyValueWhatsNewVersionStore(
```
elgatovital, line 54: what happens after another update and some changes to WhatsNew structure? How will it distinguish different versions from each other?

biafra23: I will change the code, so that whats new is bound to a version and not shown on later or earlier versions.

biafra23: Fixed in 5f078dc

Multisig/UI/App/WhatsNewHandler.swift
```diff
@@ -0,0 +1,67 @@
...
+            ]
+    )
+
+    var whatsNewViewController: WhatsNewViewController?
+
+    init() {
+        let whatsNews = [whatsNew]
+        var configuration = WhatsNewViewController.Configuration()
+
+        configuration.backgroundColor = .white
```
MouazAlzahabi: We already have predefined styles for labels & buttons could you reuse them ? 
or define the missing one instead of setting values here?

Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -77,6 +78,8 @@ class MainTabBarViewController: UITabBarController {
         appearsFirstTime = false
 
         onFirstAppear(self)
+
+        WhatsNewHandler().whatsNewViewController?.present(on: self)
```
MouazAlzahabi, line 82: Is this takes into consideration the priority for Enter passcode screen, update available screen and terms accept screen ?

biafra23: Terms accept, Update available & passcode all come before this screen.

Multisig/UI/App/WhatsNewHandler.swift
biafra23: I assume you mean the colors in: UIColor+Styles? 

Unfortunately @TanyaEfremova did not use those names. So I will try to find out why colors look similar to those she used. But it would be really helpful if the designs used the color names from the ios Project.

biafra23: For example: is `tertiaryBackground` or `quaternaryBackground` the right color for the background?

biafra23: See f02d0abba1bdde87d8f01040c8e9a1d854aac4fc for my attempt to use the right colors.

MouazAlzahabi (approved):

PR #1734, opened 2021-12-28
Info label with tooltip

elgatovital:
- Add view for labels with info icon and tooltip
image


Multisig/UI/UI Library/Views/InfoLabel.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    private var tooltipSource: TooltipSource?
+    
+    required init?(coder decoder: NSCoder) {
+        super.init(coder: decoder)
+    }
+    
+    func setText(_ text: String = "", description: String? = nil, style: GNOTextStyle = .headline) {
+        
+        let result = NSMutableAttributedString()
```
sche: No need for this variable. You can use `attributedText` var

sche (approved):

Multisig/Assets.xcassets/ico-info.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
```
sche, line 1: You can use `info.circle` system icon

elgatovital: it's a different icon with a different look. I would not use it. Instead, I have created dark mode version for icon from our design and added it to the project.

elgatovital: image

Issue #1657, opened 2021-12-07
[Intercom] Push Notifications


sche:
Implement push notifications for incoming Intercom messages when the App is closed or in the background.

- [x] Display incoming messages
- [x] Opening the push message should open App and show Intercom chat screen in a popover

biafra23:
Regarding: Display incoming message:
- According to @TanyaEfremova we should show it only on certain screens. For example not when the user is in the process of signing (or in the future creating a transfer)

In that case we should increase the badge on the `Chat with us` setting.

This should be the same for any incoming message. Regardless of push or while the app is already open.

biafra23:
@TanyaEfremova said: No number in the badge necessary. Just a small coloured badge.

TanyaEfremova:
Screens where we could show the incoming message:
- Balances;
- Transactions: queue, history;
- Settings: app, safe
- Dapps (if possible, because it conflicts with walletConnect btn)
 
Such screens as tx details should not show the message, as it might conflict with the action buttons (Confirm, Execute).

liliya-soroka:
Reopened: Click on push notification opens app but doesn't open intercom chat. Also nothing happens on "Chat with us" click. So as a user I know that I got a new message from support but I can not check it. The app restart is required to open intercom chat 

Steps:
1. Send app to the background or close it
2. send incoming notification from intercom to the user in the app
3. click notification
Current result: the app is opened but the intercom chat doesn't appear. 
If you go to the settings and click "chat with us" nothing happens
https://drive.google.com/file/d/1985rStTNKEwGUoZc8zGBVPVFZ2tX7kxP/view?usp=sharing
app version 3.6.0 (769)
staging

liliya-soroka:
Reopened with additional step:
The app should be locked with passcodes when the push notification is clicked.
After app unlock the intercome message is not displayed 
If the user opens the app from push notification before it's locked with passcode everything works

app version 3.6.0 (798)

biafra23:
I cannot reproduce the error, @liliya-soroka describes. But I have another one that involves opening from Push notification and having a passcode set: 

(Needs better description)
And after a while the screen is unlocked automatically when I recall the application. But the chat is always shown.
And then, when I close the chat the screen is locked. So that is a bug.

PR #1745, opened 2022-01-05
Gh 1657: Push fix

biafra23:
Handles #1657 

- Messenger can't be opened from PN handler

### DIFF
--- a/Multisig/UI/App/MainTabBarViewController.swift
+++ b/Multisig/UI/App/MainTabBarViewController.swift
@@ -79,11 +79,11 @@ class MainTabBarViewController: UITabBarController {
 
         onFirstAppear(self)
 
-        WhatsNewHandler().whatsNewViewController?.present(on: self)
-
         if (App.shared.intercomConfig.presentMessenger) {
             App.shared.intercomConfig.presentMessenger = false
             App.shared.intercomConfig.startChat()
+        } else {
+            WhatsNewHandler().whatsNewViewController?.present(on: self)
         }
     }
 
--- a/Multisig/UI/App/MainTabBarViewController.swift
+++ b/Multisig/UI/App/MainTabBarViewController.swift
@@ -80,6 +80,11 @@ class MainTabBarViewController: UITabBarController {
         onFirstAppear(self)
 
         WhatsNewHandler().whatsNewViewController?.present(on: self)
+
+        if (App.shared.intercomConfig.presentMessenger) {
+            App.shared.intercomConfig.presentMessenger = false
+            App.shared.intercomConfig.startChat()
+        }
     }
 
     private func balancesTabViewController() -> UIViewController {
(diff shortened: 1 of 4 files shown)


Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -80,6 +80,11 @@ class MainTabBarViewController: UITabBarController {
         onFirstAppear(self)
 
         WhatsNewHandler().whatsNewViewController?.present(on: self)
+
+        if (App.shared.intercomConfig.presentMessenger) {
```
biafra23, line 82: Maybe this is not the best place for this check?

elgatovital: is it possible that intercom could be presented on top of what's new screen with current implementation?

biafra23: Yes. I hope that is what happens if both are displayed.

biafra23: Should we not launch the Messenger if there is a WhatsNew screen? Or have the WhatsNew on top of the Messenger?

DmitryBespalov: in UIKit, only one screen can be presented at the same time from the same controller. This indeed may collide with what's new, and I also think that this should be either mutually exclusive or scheduled with one after another (close intercom --> show what's new).

DmitryBespalov: I think making it mutually exclusive is fine, and we should prefer intercom over the what's new, because if the intercom flag is set, then the user intends to open the intercom, and what's new is not what is expected.

DmitryBespalov: @biafra23 can you please align with  Andrey & Liliya for this to be expected behavior?

Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -79,7 +79,12 @@ class MainTabBarViewController: UITabBarController {
 
         onFirstAppear(self)
 
-        WhatsNewHandler().whatsNewViewController?.present(on: self)
+        if (App.shared.intercomConfig.presentMessenger) {
+            App.shared.intercomConfig.presentMessenger = false
+            App.shared.intercomConfig.startChat()
+        } else {
+            WhatsNewHandler().whatsNewViewController?.present(on: self)
```
biafra23, line 86: @sche & @liliya-soroka Is it ok to not show the WhatsNew sceen if the user opened the app via push notification for Intercom chat? They will see it on the next app opening that is not done via PN.

sche: It is ok not to show

liliya-soroka: I think it's even better not to show. Is it only for Intercome push notifications or for the tx push notifications too ?

DmitryBespalov (approved):

Issue #1743, opened 2022-01-04
Implement executing transaction from the backend data


DmitryBespalov:
Given the backend response (and WC response), implement executing transaction in the tests.

- ~[ ] ERC721 Contract ABI (https://eips.ethereum.org/EIPS/eip-721)~
- ~[ ] ERC712 signTypedData~
  - https://medium.com/metamask/eip712-is-coming-what-to-expect-and-how-to-use-it-bb92fd1a7a26
  - https://eips.ethereum.org/EIPS/eip-712
  - https://eips.ethereum.org/EIPS/eip-191
  - https://eips.ethereum.org/EIPS/eip-155
  - https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.js
  - https://docs.metamask.io/guide/signing-data.html
- [x] Safe Contracts ABI (multiple versions) (https://github.com/gnosis/safe-contracts/releases)
- [x] Safe Client Gateway Transaction to Ethereum Transaction conversion
- ~[ ] WalletConnect Transaction to Ethereum Transaction conversion~ WC key execution is implemented
- [x] Fiat price API - use the balances endpoint to get fiat conversion factor in order to calculate the price

liliya-soroka:
Verified
app version 3.6.0 (822)
Only fiat is not implemented ( moved to the new ticket)

PR #1746, opened 2022-01-05
Gh 1743/safe abi

DmitryBespalov:
Handles #1743 

- current copy of the assets from the https://github.com/gnosis/safe-deployments
- generator of the Swift contract types from the Json Abi description
- generated ABI for the safe deployments (1.0.0 - 1.3.0)


Packages/Ethereum/.swiftpm/xcode/xcshareddata/xcschemes/potato.xcscheme
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elgatovital, line 1: why "potato"? What is the purpose of this scheme?

Packages/Ethereum/Sources/SafeAbi/GnosisSafe_v1_0_0.swift
```diff
@@ -0,0 +1,1127 @@
+// THIS FILE IS GENERATED. DO NOT MODIFY BY HAND.
```
elgatovital, line 1: this class is generated, right? Along with several others in this pr. What were the reasons not to exclude generated code from repository? Could it be excluded and added to .gitignore?

Packages/Ethereum/.swiftpm/xcode/xcshareddata/xcschemes/potato.xcscheme
DmitryBespalov: the potato scheme builds the potato tool which is a generator of the safe ABIs (for now)

Packages/Ethereum/Sources/SafeAbi/GnosisSafe_v1_0_0.swift
DmitryBespalov: yes, but it is not automatically generated and inserted into the build process on every build. I ran the generator manually, so I'm committing this so that we can use these APIs in order to initiate transactions

DmitryBespalov: @elgatovital I'd remove it in the future when/if there's a way to auto-generate and insert in the build process

elgatovital (approved):

Packages/Ethereum/Sources/SafeDeployments/SafeDeployments.swift
```diff
@@ -0,0 +1,101 @@
...
+        case DefaultCallbackHandler = "default_callback_handler"
+        case CompatibilityFallbackHandler = "compatibility_fallback_handler"
+        case MultiSend = "multi_send"
+        case MultiSendCallOnly = "multi_send_call_only"
+        case ProxyFactory = "proxy_factory"
+        case SignMessageLib = "sign_message_lib"
+        case SimulateTxAccessor = "simulate_tx_accessor"
+    }
+
+    public enum Version: String, Codable {
```
MouazAlzahabi, line 46: I'm not sure if we can do this, but I believe we should count on backend to fetch these constants, because now backend controls which safe version supported or not,

Packages/Ethereum/Sources/SafeDeployments/SafeDeployments.swift
```diff
@@ -0,0 +1,101 @@
...
+        case v1_1_1 = "v1.1.1"
+        case v1_2_0 = "v1.2.0"
+        case v1_3_0 = "v1.3.0"
+
+        public var identifier: String {
+            rawValue.replacingOccurrences(of: ".", with: "_")
+        }
+    }
+
+    public enum ChainId: String, Codable, Hashable {
```
MouazAlzahabi, line 57: Same here

Packages/Ethereum/Sources/SafeDeployments/SafeDeployments.swift
DmitryBespalov: We will use backend to decide which actual safe ABI to use, but this is part of the package that generates the ABIs. It doesn't depend on the client gateway.

DmitryBespalov: It is based on the Solidity contracts (or rather Json ABI description of contracts of different versions)

MouazAlzahabi (approved):

Issue #1741, opened 2022-01-03
[iOS]Crash when WC is used for tx creation 


liliya-soroka:
**Bug description**
Crash when WC is used for tx creation  after connection or when one by one txs are created - https://console.firebase.google.com/u/0/project/safe-firebase-staging/crashlytics/app/ios:io.gnosis.multisig.staging.mainnet/issues/390768c9d074effa9794368c86f6860a?time=last-seven-days&type=all&sessionEventKey=c8637046455c47eca4662ed096cf6987_1627392597618573773
**Steps To Reproduce**
1. Go to the dapps
2. Click WallectConnect app on rinkeby
3. connect to the WC app
4. click :eth_sendTransaction
Current result: Multisig app is crashed message appears 

5. Go to the app
Current result: No incoming tx from the WC app

6. restart the safe app and click :eth_sendTransaction again
7. go to the safe app
Current result: The incoming tx is displayed

8.  confirm and go to the WC app again
9. click :eth_sendTransaction
Current result: Multisig app is crashed message appears 

**Expected Result**
It should be possible to create tx without crash and app restart
**Screenshots**
https://drive.google.com/file/d/15XEt7SMaXuaburXRum3G2te-qOJkX-us/view?usp=sharing
**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.6.0 (769)
 
**Environment**
staging
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.6.0 (796)

Issue #1740, opened 2022-01-03
[iOS] Can not load safe settings- Error 429


liliya-soroka:
**Bug description**
Can not load safe settings- Error 429
**Steps To Reproduce**
1. Open safe settings
Current result: The safe settings is displayed and ens name loading is in progress. In some time the error appears instead of safe settings
**Expected Result**
The safe settings should be loaded
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/39b684e0-c0a3-4f3c-a295-b6dcf471045e)
**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.6.0 (769)
 
**Environment**
staging 
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.6.0 (796)

PR #1764, opened 2022-01-11
gh-1741 fix crash

sche:
Handles #1741, #1740

Fix crash

### DIFF
--- a/Multisig/Logic/Models/Safe.swift
+++ b/Multisig/Logic/Models/Safe.swift
@@ -156,7 +156,7 @@ extension Safe: Identifiable {
         self.name = name
 
         App.shared.coreDataStack.saveContext()
-        NotificationCenter.default.post(name: .selectedSafeUpdated, object: nil)
+        NotificationCenter.default.post(name: .selectedSafeUpdated, object: self)
 
         Safe.updateCachedNames()
     }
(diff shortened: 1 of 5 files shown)


Multisig/Logic/Models/Safe.swift
```diff
@@ -206,20 +206,18 @@ extension Safe: Identifiable {
...
         nonce = info.nonce.value
         modulesInfo = info.modules?.map { $0.addressInfo }
         fallbackHandlerInfo = info.fallbackHandler?.addressInfo
         guardInfo = info.guard?.addressInfo
-                            
-        NotificationCenter.default.post(name: .selectedSafeUpdated, object: self)
+        DispatchQueue.main.async {
+            self.contractVersion = info.version
+            App.shared.coreDataStack.saveContext()
+            NotificationCenter.default.post(name: .selectedSafeUpdated, object: self)
```
sche, line 221: @elgatovital this line leads to an inifinit loop on the safe settings screen. What was the reason to add it?

sche: fixed

elgatovital: header with send/receive buttons needs safe info to decide where to navigate when user clicks send button: add owner key screen or send asset screen, depending whether safe is read only. This propagates safe info to the view controller of the header.

elgatovital (approved):

DmitryBespalov (approved):

Issue #1755, opened 2022-01-10
[ExecuteTx] Review screen: overall UI


DmitryBespalov:
Scrollable screen with the header sections and execution options sections, with “Submit” button.

[Review screen in Figma](https://www.figma.com/file/BHlmCT7Qp3jDpBaK39Vh7D/Execute-transactions?node-id=1255%3A11562)

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/fb73ecef-41b9-4a55-86f0-4673455ae356)

liliya-soroka:
Verified
app version 3.6.0 (802)

PR #1766, opened 2022-01-11
gh-1755 Created overall ui files

DmitryBespalov:
Handles #1755 

- Overall UI for the screen
- Difference from design: the screen is opened as a modal (it seems to be more meaningful this way)

### DIFF
--- a/Multisig/UI/Transaction/Execution/ReviewExecutionCellBuilder.swift
+++ b/Multisig/UI/Transaction/Execution/ReviewExecutionCellBuilder.swift
@@ -1,0 +1,43 @@
+//
+//  ReviewExecutionCellBuilder.swift
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 11.01.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+class ReviewExecutionCellBuilder: TransactionDetailCellBuilder {
+
+    override func buildTransaction(_ tx: SCGModels.TransactionDetails) {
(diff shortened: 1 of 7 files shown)


Multisig/UI/Transaction/Execution/ReviewExecutionCellBuilder.swift
```diff
@@ -0,0 +1,43 @@
...
+//  ReviewExecutionCellBuilder.swift
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 11.01.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+class ReviewExecutionCellBuilder: TransactionDetailCellBuilder {
```
MouazAlzahabi, line 11: Do we need CellBuilder files, why don't we just create in the view controller directly

MouazAlzahabi (approved):

sche (approved):

Multisig/UI/Transaction/Execution/ReviewExecutionCellBuilder.swift
DmitryBespalov: Well, this extracts responsibility of building the cells into a separate class and leaves the view controller with gluing things together.

Also, this time this cell builder extends existing one, so it's easy to compose that one. If I would have the TxDetailCellBuilder code in the TxDetailsVC then I would need to inherit from the TxDetailsVC in my review execution controller, which I don't want because it has some other logic.

Issue #1758, opened 2022-01-10
[ExecuteTx] Edit Transaction Fee screen


DmitryBespalov:
# Edit transaction fee screen

This screen gives user a way to change the fee-related parameters of the transaction to some manual values instead of using automatically estimated values.

The actual fields depend on the transaction type which depends on the supported transaction type of the chain (Chain Info -> features -> EIP-1559)

[Figma Link](https://www.figma.com/file/BHlmCT7Qp3jDpBaK39Vh7D/Execute-transactions?node-id=1255%3A11753)

![Advanced settings - Edit_@2x.png](https://images.zenhubusercontent.com/6183a79a848bb3312854c951/67446dcc-cd69-4478-aaec-c6848e0ce032)

* Legacy Transaction type
    * Nonce
    * Gas limit
    * Gas Price (Gwei)
* EIP-1559 Transaction type
    * Nonce
    * Gas limit
    * Base Fee
    * Max Priority Fee (or miner’s tip)
    * Max Fee Per Gas (must equal to base fee + max priority fee)
* Transactions on the Optimism chain have different fee structure and are handled in a separate issue. Some fields will not be editable.

Tooltips for each field?

“Reset to default” button in each field in case the value was manually set? - not implemented here, will be listed in another ticket

* This would revert to the automatically estimated value 

Estimated Fee Price

* Value depends on the fee structure and can be presented in both native coin value and if price is available then in fiat value as well.

Link to the help article - TBD

### Current CGW settings for the EIP-1559 support on each chain
(10.02.2022)

<table>
  <tr>
   <td>Chain 
   </td>
   <td>Transaction Type
   </td>
  </tr>
  <tr>
   <td>Ethereum
   </td>
   <td>EIP1559
   </td>
  </tr>
  <tr>
   <td>Gnosis
   </td>
   <td>EIP1559
   </td>
  </tr>
  <tr>
   <td>Polygon
   </td>
   <td>Legacy
   </td>
  </tr>
  <tr>
   <td>Binance Smart Chain
   </td>
   <td>Legacy
   </td>
  </tr>
  <tr>
   <td>Energy Web Chain
   </td>
   <td>Legacy
   </td>
  </tr>
  <tr>
   <td>Arbitrum
   </td>
   <td>Legacy
   </td>
  </tr>
  <tr>
   <td>Avalanche
   </td>
   <td>EIP1559
   </td>
  </tr>
  <tr>
   <td>Optimism
   </td>
[...]

PR #1774, opened 2022-01-13
gh-1758 Edit fee screen

DmitryBespalov:
Handles #1758

Together with @biafra23 💪 

- FormVC which allows for constructing forms of multiple inputs with validation
- Fee form for the legacy fee structure with value validation
- Not connected to the fee estimation logic yet.

### DIFF
--- a/Multisig/UI/Transaction/Execution/FeeLegacyFormModel.swift
+++ b/Multisig/UI/Transaction/Execution/FeeLegacyFormModel.swift
@@ -1,0 +1,195 @@
+//
+//  FeeLegacyFormModel.swift
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 13.01.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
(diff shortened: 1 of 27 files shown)


Multisig/UI/Transaction/Execution/FeeLegacyFormModel.swift
```diff
@@ -0,0 +1,195 @@
...
+//  Created by Dmitry Bespalov on 13.01.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import Solidity
+import Ethereum
+import UIKit
+
+class FeeLegacyFormModel: FormModel {
```
elgatovital, line 14: why `legacy`?

DmitryBespalov: because there're 2 types of fees we'll work with: fee for a legacy-type EVM transaction, and a fee for an EIP-1559 type EVM transaction which will have different parameters

Issue #1749, opened 2022-01-06
[iOS] Balances are displayed without decimals


sche:
**Bug description**
On iOS, we currently display balances without decimals.

**Steps To Reproduce**
Open assets screen

**Expected Result**
Balances if fiat are displayed with decimals. We should use the system currency formatter for displaying fiat balances.

**Device & App version**
 - App version: 3.5.0

liliya-soroka:
Verified
app version 3.6.0 (822)

PR #1829, opened 2022-01-19
gh-1749 Display fiat balances with decimals

elgatovital:
Handles #1749

### DIFF
--- a/Multisig/Logic/Models/TokenBalance.swift
+++ b/Multisig/Logic/Models/TokenBalance.swift
@@ -76,7 +76,7 @@ extension TokenBalance {
     static func displayCurrency(from serverValue: String, code: String) -> String {
         let number = serverCurrencyFormatter.number(from: serverValue) ?? 0
         let formatter = TokenFormatter()
-        let amount = Int256(number.doubleValue) * 100
+        let amount = Int256(number.doubleValue * 100)
         let precision = 2
         let currencyCode = code
         let value = formatter.string(from: BigDecimal(amount, precision),


Multisig/Logic/Models/TokenBalance.swift
```diff
@@ -76,7 +76,7 @@ extension TokenBalance {
     static func displayCurrency(from serverValue: String, code: String) -> String {
         let number = serverCurrencyFormatter.number(from: serverValue) ?? 0
         let formatter = TokenFormatter()
-        let amount = Int256(number.doubleValue) * 100
+        let amount = Int256(number.doubleValue * 100)
```
biafra23, line 79: Does this work for more or less than 2 minor units? 
There are currencies that have 0, 1 or 3 minor units of currency.

https://en.wikipedia.org/wiki/ISO_4217#Minor_units_of_currency

biafra23: BTW: We should never use floating point types for Fiat values. Always integer including minor units. But that is something for another PR :-)

DmitryBespalov: some currencies have 3 units

DmitryBespalov (approved):
I think it mergeable

Multisig/Logic/Models/TokenBalance.swift
elgatovital: apparently there is no out of the box support for ISO4217 in ios. NumberFormatter returns 2 decimal digits even for codes that should have 3 or 4. In fiat list that we get from cgw that would be CLF and IQD
Issue opened: https://app.zenhub.com/workspaces/mobile-development-board-60cc78740f488d000e996335/issues/gnosis/safe-ios/1837

Issue #1697, opened 2021-12-14
[iOS] Conversations are not displayed after calling the app from background


liliya-soroka:
**Bug description**
Conversations are not displayed after calling the app from the background.
The user has to restart app to see conversations

**Steps To Reproduce**
1. the user should have conversations in the intercom
2. run app and send it to background ( without opened intercom)
3. call app and open Intercom
Current result: No conversations

4. close the app fully and restart it ( cold start)
Current result: Conversations are displayed

**Expected Result**
The conversations should not disappear if the app was in background

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.5.0 (741)
 
**Environment**
prod
**Additional info**
(Optional)

sche:
I assume this is an Intercom issue as well.

biafra23:
Apparently the intercom messenger is dismissed when the app goes into background. But somehow the intercom sdk assumes it is still shown, so another call to presentMessenger() is ignored.

biafra23:
I contacted Intercom support regarding this issue.

biafra23:
Intercom support asked me to upgrade to the newest SDK. Upgrading to 11.0.1 did not solve the issues.

biafra23:
Is  #1696 a duplicate of this?

liliya-soroka:
Verified
app version 3.7.0 (835)

Issue #1696, opened 2021-12-14
[iOS] App restart is required to open intercom if the app was sent to background with opened Intercom


liliya-soroka:
**Bug description**
 App restart is required to open intercome if the app was sent to background with opened Intercome
**Steps To Reproduce**
1. Run app and open intercom using"Chat with us" in the settings
2. send app to the background
3. call the app from background and try to open Intercom again
Current result: Nothing happens

4. send message from intercomn to mobile app user
Current result: the badge appears but the user can not open Intercom

5. restrat app
Current result: the intercom is opened

**Expected Result**
The user should be able to open the Intercom without app restart

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.5.0 (741)
 
**Environment**
prod
**Additional info**
(Optional)

sche:
It looks like an Intercom issue. We need to provide feedback to their team. Unfortunately, they don't have the "Issues" section here https://github.com/intercom/intercom-ios 

What we can try to do to fix it:
- detect if there is an open Intercom window before going to background, and close it manually

liliya-soroka:
Verified
app version 3.7.0 (835)

PR #1841, opened 2022-01-19
GH-1696: Intercom can't be opened after calling app from background -> main

biafra23:
Handles #1696 #1697

Hide Intercom when ging to background. There is no way to check whether it is open and that check also seems unnecessary. hide() works either way.

- hide() Intercom when going to background. Whether it is open or not.

### DIFF
--- a/Multisig/UI/App/MainTabBarViewController.swift
+++ b/Multisig/UI/App/MainTabBarViewController.swift
@@ -68,6 +68,16 @@ class MainTabBarViewController: UITabBarController {
             selector: #selector(updateTabs),
             name: .updatedExperemental,
             object: nil)
+
+        NotificationCenter.default.addObserver(
+                self,
+                selector: #selector(appMovedToBackground),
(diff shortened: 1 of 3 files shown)


Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -68,6 +68,16 @@ class MainTabBarViewController: UITabBarController {
...
+
+        NotificationCenter.default.addObserver(
+                self,
+                selector: #selector(appMovedToBackground),
+                name: UIApplication.willResignActiveNotification,
+                object: nil)
+    }
+
+    @objc func appMovedToBackground() {
+        App.shared.intercomConfig.hide()
```
DmitryBespalov: This seems a bit more complicated than needed: will it work if we put this line into SceneDelegate.swift sceneWillResignActive() method? Then no need to use notifications

DmitryBespalov (approved):
Looks good! 🚀

sche (approved):

Issue #1813, opened 2022-01-18
[iOS]Execution with connected WC key doesn't work ( MM and rainbow mobile apps)


liliya-soroka:
**Bug description**
Execution with connected WC key doesn't work ( MM and rainbow mobile apps)
**Steps To Reproduce**
1. Owner key is connected via WC with mobile MM or mobile rainbow app
2. try to execute the tx

**Current result**
Sign execution message doesn't appear in connected mobile app.
At the same time sign confirmation message appears without the problem

**Expected Result**
The execution with key connected via WC should work

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.6.0 (812)
 
**Environment**
staging
**Additional info**
(Optional)

liliya-soroka:
Verified for MM
The wrong URL is shown for the rainbow - a separate issue

app version 3.7.0 (835)

PR #1848, opened 2022-01-20
Gh 1813/hack wc send tx

DmitryBespalov:
Handles #1813

- Using the 'sendTransaction' method to execute with WalletConnect instead of 'signTransaction'


Multisig/UI/WalletConnect/Client/WalletConnectClientController.swift
```diff
@@ -135,27 +135,27 @@ class WalletConnectClientController {
         }
     }
 
-    func sign(transaction: Client.Transaction, completion: @escaping (String?) -> Void) {
+    func send(transaction: Client.Transaction, completion: @escaping (String?) -> Void) {
```
sche, line 138: I think it will break signing with WC keys. 
Execution is in the `execute` method.

Multisig/UI/WalletConnect/Client/WCPendingConfirmationViewController.swift
```diff
@@ -102,6 +98,23 @@ class WCPendingConfirmationViewController: UIViewController {
         }
     }
 
+    func send(completion: @escaping (String?) -> Void) {
```
sche, line 101: when do we call this method?

sche (changes requested):
need to clarify some moment

Multisig/UI/WalletConnect/Client/WalletConnectClientController.swift
DmitryBespalov: this was an additional custom method that I added while working on the release, it will not touch anything else

Multisig/UI/WalletConnect/Client/WCPendingConfirmationViewController.swift
DmitryBespalov: please use the search

DmitryBespalov: Check the ReviewExecutionViewController file in this pr

Multisig/UI/WalletConnect/Client/WalletConnectClientController.swift
sche: Should we remove 'execute' method?

DmitryBespalov: not at the moment, I don't expect this to be a WC refactoring, I think this should be done in a separate task

sche (approved):

Issue #1862, opened 2022-01-20
[iOS] Crash when all data is removed and the app is still waiting for network response.


DmitryBespalov:
**Bug description**
App crashes after clearing all data. This crash also appeared during regression testing.

**Steps To Reproduce**
0. There must be passcode set up.
1. Enable 'Settings --> Developer --> Network Link Conditioner' in phone settings with very slow network response (100% loss or create a custom with big latency/delay)
2. Have a safe in the app. Reload safe settings screen or dapps screen.
3. While it is reloading, minimize the app, wait a bit, then open to see passcode prompt.
4. Tap on "Forgot passcode" and confirm to remove all data
---
The app crashes

**Expected Result**
The app does not crash and removes all data.

**Device & App version**
 - Device/OS: [e.g. iPhone8/iOS13.4.1] iPhone 13 Pro Max / iOS 15.2.1
 - App version: release 3.6.0 824
 
**Environment**
staging

**Additional info**
(Optional)

PR #1863, opened 2022-01-20
gh-1862 Fixing the crash introduced earlier

DmitryBespalov:
Handles #1862

- Fixing the crash introduced earlier during the fix of another crash.

### DIFF
--- a/Multisig/UI/Dapps/DappsDataSource.swift
+++ b/Multisig/UI/Dapps/DappsDataSource.swift
@@ -20,7 +20,7 @@ class DappsDataSource {
 
     init() {
         guard let chain = try? Safe.getSelected()?.chain else {
-            assertionFailure("Developer error: expect to have selected safe with network")
+            // it may happen that the data was deleted and we get updated from a notifictaion.
             dapps = []
             return
         }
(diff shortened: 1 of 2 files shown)


Multisig/UI/Dapps/DappsDataSource.swift
```diff
@@ -20,7 +20,7 @@ class DappsDataSource {
 
     init() {
         guard let chain = try? Safe.getSelected()?.chain else {
```
DmitryBespalov, line 22: @sche @elgatovital please bear in mind that we're not always will get a safe if the safe selection changes. If the data is cleared, then selection is changed to nil. This is an expected case and it should be handled.

This situation arised when this screen was still refreshing due to slow network. Edge case, but still 🤷

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -102,7 +102,11 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
 
     @objc func update() {
         do {
-            safe = try Safe.getSelected()!
+            // it may happen that data is deleted, so we should just finish gracefully.
```
DmitryBespalov, line 105: same here

DmitryBespalov:
Please merge when approved 🙇

Issue #1886, opened 2022-02-03
[iOS]Execution - Tx will fail is displayed for any tx on the network without EIP-1559 ( optimism, arbitrum)


liliya-soroka:
**Bug description**
[iOS]Execution - Tx will fail is displayed for any tx on the network without EIP-1559 ( optimism, arbitrum)
**Steps To Reproduce**
1. try to execute tx on the network without EIP-1559 ( safe =oeth:0xf9D445a46D65427550174ec2A49FceFA3dae7f8D)
Current result: the error is displayed and Submit button is blocked
**Expected Result**
It should be possible to execute valid txs on the networks without EIP-1559
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.6.1(3)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
@DmitryBespalov , could you check tx execution in the safe=ewt:0x855cB0Bb58a5413969BA455f860FAFd7FA66F883 . I am getting tx will fail only on mobile 
3.6.1(6)

liliya-soroka:
Verified
app version 3.6.1(7)

PR #1888, opened 2022-02-04
gh-1886 using legacy tx data format

DmitryBespalov:
Handles #1886 

- Revert previous fix
- Use 'legacy' tx format. The eth_call was failing because the API did not expect to see 'chainId' and other parameters. When executed on EWT this received proper error response. I'm reusing the same format that worked for estimating gas.

### DIFF
--- a/Multisig/UI/Transaction/Execution/TransactionExecutionController.swift
+++ b/Multisig/UI/Transaction/Execution/TransactionExecutionController.swift
@@ -214,11 +214,9 @@ class TransactionExecutionController {
                     gas = try partialResults.gas.get()
 
                     let execTransactionSuccess = try partialResults.ethCall.get()
-                    if !execTransactionSuccess.isEmpty {
-                        let success = try Sol.Bool(execTransactionSuccess).storage
-                        guard success else {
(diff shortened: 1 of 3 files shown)


DmitryBespalov:
@MouazAlzahabi you were right that the fix was not a proper fix... I didn't know why the returned data was empty bytes. Turned out it is because the API didn't expect some parameters.

Multisig/UI/Transaction/Execution/TransactionExecutionController.swift
```diff
@@ -214,11 +214,9 @@ class TransactionExecutionController {
                     gas = try partialResults.gas.get()
 
                     let execTransactionSuccess = try partialResults.ethCall.get()
-                    if !execTransactionSuccess.isEmpty {
-                        let success = try Sol.Bool(execTransactionSuccess).storage
-                        guard success else {
-                            throw TransactionExecutionError(code: -7, message: "Internal Gnosis Safe transaction fails. Please double check whether this transaction is valid with the current state of the Safe.")
-                        }
+                    let success = try Sol.Bool(execTransactionSuccess).storage
```
DmitryBespalov, line 217: reverted because now we get the result

Issue #1901, opened 2022-02-07
[WebConnect] Open Connection Request when scanning from the Dapps screen.


DmitryBespalov:
We already have the "WC" button in the Dapps screen. We want to make possible to scan the QR codes for the Connect To Web feature.

This would navigate to the Web Connections list screen and open the Connection Request screen.

DmitryBespalov:
@TanyaEfremova I think we could change the icon of the scanner button on the dapps tab.

TanyaEfremova:
@DmitryBespalov The icon has been replaced here:  https://www.figma.com/file/61K9gzH5518LuSBpHLIfki/DApps-improvements?node-id=104%3A646

biafra23:
@TanyaEfremova Looks like this button is smaller than the current WC-Button. Is that intentional?

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/adaf815c-5dd2-4830-b6ef-3655d4c67cbe)
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/b1ac6c04-da31-4594-9e82-e7e44a414e9f)

TanyaEfremova:
@biafra23 I am not sure where the second screenshot comes from, but when I compare the current state (left) and design (right) they look similar (in design it is even a couple pixels bigger): 
<img width="490" alt="Screenshot 2022-02-09 at 17 32 00" src="https://user-images.githubusercontent.com/93707525/153245770-9d5d6246-cd12-4737-91aa-21da0c67ec61.png">


In any case - there is no need to change the icon size.

biafra23:
@TanyaEfremova Looks like the icon inside the PDF is smaller: 

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/6e54be3a-abed-4b8e-af0b-77a7ba66d9e7)
![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/eea618a2-27b3-4ca0-9fee-5ac85a496c78)

Not sure I can or should fix this.

liliya-soroka:
Verified
app version 3.7.0 (865)

PR #1908, opened 2022-02-09
Gh 1901 - WC FAB scanner

biafra23:
Handles #1901

- Change FAB icon in dApps tab from WC to more generic scan icon
- Scan safe-wc (web connect) qr codes from Dapp-Tab
- Redirect users to `DesktopPairingViewController()` if scanned url has a `safe-wc` scheme

image


Multisig/UI/Dapps/DappsViewController.swift
```diff
@@ -257,9 +259,12 @@ class DappsViewController: UIViewController, UITableViewDataSource, UITableViewD
 
 extension DappsViewController: QRCodeScannerViewControllerDelegate {
     func scannerViewControllerDidScan(_ code: String) {
+        let wcServerController = WalletConnectKeysServerController.shared
+        wcServerController.delegate = self
```
biafra23: Set as delegate

Multisig/UI/Dapps/DappsViewController.swift
```diff
@@ -270,3 +275,37 @@ extension DappsViewController: QRCodeScannerViewControllerDelegate {
         dismiss(animated: true, completion: nil)
     }
 }
+
+extension DappsViewController: WalletConnectKeysServerControllerDelegate {
+    func shouldStart(session: Session, completion: @escaping ([KeyInfo]) -> Void) {
```
biafra23: This should be called after the scanner returns a URL

DmitryBespalov: @biafra23 I would suggest to leave this functionality to the "Connect screen" I'm working on right now. For this issue, if you could trigger opening Settings and opening the "Connect To Web" screen after scanning the QR code, this would be good to go!

biafra23: Really to `Connect to web`? Then they have to scan again. I could go to the key selection screen. Or will you change this later?

elgatovital:
this will conflict with https://github.com/gnosis/safe-ios/pull/1909 as DesktopPairingViewController was refactored there and renamed to WebConnectionsViewController. We have to think about the optimal order for merging. @DmitryBespalov what do you think?

elgatovital (approved):

Issue #1896, opened 2022-02-07
[WebConnect] Connection Details


DmitryBespalov:
When tapping on an item in the Web Connections list, the user would see the details of the connection.

- [Figma](https://www.figma.com/file/IjW9CfVZvUqiMgj28euXVX/Desktop-Pairing?node-id=397%3A4384)

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/244cb2de-2e93-40b2-afb9-ed72b4d49dcc)

Sketch of the data model:
![Screenshot 2022-02-07 at 19.02.39.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/ae969a5e-ae9a-4587-aa5b-4985610413f0)

This work item **does not** include actual disconnection, changing network or changing account functionality. It is only about displaying connection details and navigating to other items.

DmitryBespalov:
Note for @liliya-soroka: the "half-size" of this screen is only possible on iOS 15, and then we can't actually control the height to fit the content easily (would require much more effort to implement).

So, on iOS 15 it will open in the "middle" of the screen, but on earlier versions it will open in full screen.

Also note, right now we don't receive proper "browser" and "version" from the web app, at least something is wrong there. When these are not available, we'll show the app's "description" text that we receive.

liliya-soroka:
Verified
app version 3.7.0 (884)

PR #1918, opened 2022-02-16
GH-1896 - Connection Details

biafra23:
Handles #1896

**Missing/Wrong**
- Screen Title ("Conection Details") is not in the design. Easy to remove, but I find it helpful. If we remove it, how can I get rid of the line below it. We need the close button
- Icon is too small. How do I scale it? Od do  I have to add an asset with the right size?
- "Gnosis Safe" is not centered. Why is the icon centered?
- The time since creation is not shown. The other two are created but not detailTextLabel. Why
- Version -> missing in model
- Browser -> missing in model
- Button has wrog style and color. Does it make more sense to adapt ButtonTableViewCell or create a new cell type 

Design | Draft
-------|------
image |


DmitryBespalov:
> Icon too small
Icon size: choose owner details header view
Make an outlet to the swift file for the constraint

constratint.constant = <new size>
setNeedsUpdateConstratints()

> Gnosis Safe centering
Check that label's text alignment is centered

biafra23:
@DmitryBespalov I created that outlet and called it: imageHeight. But I do not know, what to do with it. setContent() already calls setNeedsUpdateConstraints(). See: SHA1: dfef6fc156090edbc52716c3e44a5db151a1520d

DmitryBespalov (approved):

Issue #1960, opened 2022-02-22
[iOS] Desktop pairing - execution on polygon doesn't work 


liliya-soroka:
**Bug description**
Desktop pairing - execution on polygon doesn't work 
**Steps To Reproduce**
1. connect to the web app on polygon
2. try to execute tx on web
3. submit tx on mobile
Current result: the error is displayed and tx is not executed
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/24e921d0-a462-4042-b478-011f32066070)
**Expected Result**
the execution on polygon should work

**Device & App version**
 - Device/OS:iPhone 11/14.8
 - App version: 3.7.0 (876)
 
**Environment**
staging, polygon
**Additional info**
(Optional)

related to :
[WebConnect] Send Transaction Request

liliya-soroka:
Verified
app version 3.7.0 (882)

PR #1967, opened 2022-02-23
gh-1960 Use chain-related execution parameters

DmitryBespalov:
Handles #1960 

- Use only the 'message call' related values received from the web connection request

### DIFF
--- a/Multisig/Features/Connect to Web/Logic/WebConnectionController.swift
+++ b/Multisig/Features/Connect to Web/Logic/WebConnectionController.swift
@@ -5,6 +5,8 @@
 
 import Foundation
 import WalletConnectSwift
+import Ethereum
+import Solidity
 
 protocol WebConnectionObserver: AnyObject {
     func didUpdate(connection: WebConnection)
@@ -644,6 +646,7 @@ class WebConnectionController: ServerDelegateV2, RequestHandler, WebConnectionSu
                 try? server.send(Response(request: wcRequest, error: .invalidParams))
                 return
             }
+            prepareTransactionForExecution(sendTxRequest)
 
         default:
             break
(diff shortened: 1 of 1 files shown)


Multisig/Features/Connect to Web/Logic/WebConnectionController.swift
```diff
@@ -655,6 +658,40 @@ class WebConnectionController: ServerDelegateV2, RequestHandler, WebConnectionSu
...
+            let features = chain.features
+        else {
+            return
+        }
+
+        let minerTip: Sol.UInt256 = 1_500_000_000
+
+        // select tx type based on the chain.
+        let isEIP1559 = features.contains("EIP1559")
+        if isEIP1559 {
```
elgatovital, line 675: what about EIP2930?

Multisig/Features/Connect to Web/Logic/WebConnectionController.swift
```diff
@@ -655,6 +658,40 @@ class WebConnectionController: ServerDelegateV2, RequestHandler, WebConnectionSu
...
+            request.transaction = Eth.TransactionEip1559(
+                chainId: Sol.UInt256(chainId),
+                from: request.transaction.from,
+                to: request.transaction.to,
+                value: request.transaction.value,
+                input: request.transaction.data,
+                fee: .init(maxPriorityFee: minerTip)
+            )
+        } else {
+            request.transaction = Eth.TransactionLegacy(
```
elgatovital, line 685: would EIP-2930 type be handled here correctly?

biafra23: Do we need to handle EIP-2930? We're not using access lists. Do we allow the creation of a tx with EIP-2930 specific featueres? For example access lists?

biafra23: If this handling is a problem, we need to change it in TransactionExecutionController.swift:L324 as well.

Issue #1989, opened 2022-02-28
[CreateSafe] Add animated images for the 'in progress' and 'success' screens


DmitryBespalov:
Add animation

liliya-soroka:
Verified
app version 3.8.0 (912)

PR #1999, opened 2022-02-28
Add animation

MouazAlzahabi:
Handles #1989


elgatovital:
so many files for a little animation. I'm afraid this is not very manageable. It will take unnecessary effort to maintain this. Maybe there is another easier and more compact way.

DmitryBespalov (approved):

Issue #1996, opened 2022-02-28
[CreateSafe] Remove 'Error' title from 'Insufficient balance


TanyaEfremova:
- Whenever there is insufficient balance for executing a tx, we display the text without the label 'Error'. Remove it from the error message.

- Remove the section header from the error messafe.

https://files.slack.com/files-pri/T58NW6E06-F034WPY868J/image_from_ios.png

liliya-soroka:
Reopened:
The current view of the errors during safe creation:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/1d69b62b-669b-4181-ab5c-e84af9c85224)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/4267b479-6084-4af9-b2b1-154b65371956)

Expected result: According to the design - https://www.figma.com/file/dPIs2Whb0BoGeeoMS6gSro/Create-a-New-Safe?node-id=684%3A7349

app version 3.8.0 (912)

elgatovital:
should be fixed with gh-2067

liliya-soroka:
Verified
app version 3.9.0 (960)

PR #2003, opened 2022-03-01
Gh 1996/remove 'error' title from 'insufficient balance

MouazAlzahabi:
Handles #1996

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
@@ -36,7 +36,7 @@ class CreateSafeViewController: UIViewController, UITableViewDelegate, UITableVi
         tableView.registerCell(ActionDetailAddressCell.self)
         tableView.registerCell(StepperTableViewCell.self)
         tableView.registerCell(DisclosureWithContentCell.self)
-        tableView.registerCell(DetailExpandableTextCell.self)
+        tableView.registerCell(BasicCell.self)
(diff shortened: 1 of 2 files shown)


elgatovital (approved):

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
```diff
@@ -664,18 +664,14 @@ class CreateSafeViewController: UIViewController, UITableViewDelegate, UITableVi
...
-        let errorPreview = errorText.count <= 144 ? nil : (String(errorText.prefix(144)) + "…")
-        cell.tableView = tableView
-        cell.titleStyle = .error.weight(.medium)
-        cell.expandableTitleStyle = (collapsed: .error, expanded: .error)
-        cell.contentStyle = (collapsed: .error, expanded: .secondary)
-        cell.setTitle("Error")
-        cell.setText(errorText)
-        cell.setCopyText(errorText)
-        cell.setExpandableTitle(errorPreview)
+        let errorPreview = errorText.count <= 144 ? errorText : (String(errorText.prefix(144)) + "…")
```
DmitryBespalov, line 670: this is not good for the case when the error is bigger than 144 characters, for example when the error contains server response.

We need to find the way to show full content - that's why it was expandable cell before.

MouazAlzahabi: OK I will then introduce a new type of cell

Issue #1984, opened 2022-02-25
[CreateSafe] Allow to add owners from the address book


sche:
If I already have an owner key imported in the app it is helpful to be able to add it easily as an owner for a new safe. Right now I need to leave the Create Safe screen to do so. What is even worse is that if I want to add two owners that I already control with the mobile app, it becomes much harder to do so.

liliya-soroka:
Verified
app version 3.8.0 (911)

PR #2006, opened 2022-03-01
gh-1984 Added address parsing and address book

DmitryBespalov:
Handles #1984 

- Added a controller for selecting address

### DIFF
--- a/Multisig/Logic/Models/Chain.swift
+++ b/Multisig/Logic/Models/Chain.swift
@@ -45,6 +45,14 @@ extension Chain {
         return chain
     }
 
+    static func by(shortName: String) -> Chain? {
+        dispatchPrecondition(condition: .onQueue(.main))
+        let context = App.shared.coreDataStack.viewContext
+        let fr = Chain.fetchRequest().by(shortName: shortName)
+        guard let chain = try? context.fetch(fr).first else { return nil }
+        return chain
+    }
+
     @discardableResult
     static func createOrUpdate(_ chainInfo: SCGModels.Chain) -> Chain {
         guard let chain = Chain.by(chainInfo.id) else {
@@ -249,6 +257,13 @@ extension Chain {
 extension NSFetchRequest where ResultType == Chain {
     func all() -> Self {
         sortDescriptors = [NSSortDescriptor(keyPath: \Chain.id, ascending: true)]
+        return self
+    }
+
+    func by(shortName: String) -> Self {
+        sortDescriptors = []
+        predicate = NSPredicate(format: "shortName == %@", shortName)
+        fetchLimit = 1
         return self
     }
 }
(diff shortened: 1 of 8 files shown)


Multisig/Logic/Models/Chain.swift
```diff
@@ -45,6 +45,14 @@ extension Chain {
         return chain
     }
 
+    static func by(shortName: String) -> Chain? {
+        dispatchPrecondition(condition: .onQueue(.main))
```
elgatovital, line 49: why is this precondition necessary?

DmitryBespalov: Core Data is not thread safe. We're using the context for the main thread, and so every call to core data or working with any core data object from that context must happen on the same thread.

Multisig/UI/Safe Management/Create Safe/Setup Safe/SelectAddressViewController.swift
```diff
@@ -0,0 +1,192 @@
...
+        addressBookVC.onSelect = { [unowned addressBookVC, unowned self] address in
+            addressBookVC.dismiss(animated: true) {
+                completion(address)
+            }
+        }
+        let vc = ViewControllerFactory.modal(viewController: addressBookVC)
+        presenter?.present(vc, animated: true)
+    }
+
+    func validate(text: String?) throws -> Address {
```
MouazAlzahabi, line 63: In Address.swift, we have method called addressWithPrefix which do part of the logic here, can we adapt it so we have the full logic all over the app ?

Multisig/UI/Safe Management/Create Safe/Setup Safe/SelectAddressViewController.swift
```diff
@@ -0,0 +1,192 @@
...
+    convenience init(chain: Chain?, presenter: UIViewController, completion: @escaping (Address) -> Void) {
+        self.init(title: nil, message: nil, preferredStyle: .actionSheet)
+        self.presenter = presenter
+        self.chain = chain
+        self.completion = completion
+
+        addAction(UIAlertAction(title: "Paste from Clipboard", style: .default, handler: handlePaste(_:)))
+        addAction(UIAlertAction(title: "Scan QR Code", style: .default, handler: handleScan(_:)))
+        addAction(UIAlertAction(title: "Address Book", style: .default, handler: handleAddressBook(_:)))
+        addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: nil))
```
MouazAlzahabi, line 27: I would create another ticket to support ENS & unstoppable domains, this way we can reuse this component in different places in the app WDYT?

MouazAlzahabi (approved):

Multisig/UI/Safe Management/Create Safe/Setup Safe/SelectAddressViewController.swift
DmitryBespalov: yes it would evolve into a thing that can be reused in multiple places, but not in this pr

Multisig/UI/Safe Management/Create Safe/Setup Safe/SelectAddressViewController.swift
DmitryBespalov: I think this would be good, but I don't want to spend more time on that right now. This would require refactoring/redoing other code in other places.

Issue #1994, opened 2022-02-28
[CreateSafe] Implement "show assets" navigation programmatically 


DmitryBespalov:
Currently "showAssets" route opens only selected safe's balances segment.
Let's make it so that the app can select a safe and display appropriate screen.

Route: "/assets/"
Parameters:
- "address": safe address, string
- "chainId": chain id, string
Either both parameters present or nothing.

When address and chainId are present in the route, then the app should select such a safe. If there is no such safe, then ignore the route and do nothing.

If there are no parameters, then switch to the assets/coins for the selected safe.

Not deployed safe:
- [Select safe if it's not selected]
- Just switch to assets tab

Deployed safe:
- ]Select safe if it's not selected]
- Switch to assets tab
- Switch to coins

liliya-soroka:
Verified
app version 3.8.0 (910)

PR #2008, opened 2022-03-01
gh-1994 Routing in the tab bar controller

DmitryBespalov:
Handles #1994

- Implemented switching to any tab and segment inside tab bar using IndexPath
- Selecting safe and opening balances tab from navigation route


Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -318,21 +346,77 @@ extension MainTabBarViewController: NavigationRouter {
...
+                if !safe.isSelected {
+                    safe.select()
+                }
+            }
+            switchTo(indexPath: Path.balances)
+        }
+    }
+
+    func switchTo(indexPath: IndexPath) {
+        var indexPath = indexPath
```
MouazAlzahabi, line 373: Why we define this parameter ?

Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -318,21 +346,77 @@ extension MainTabBarViewController: NavigationRouter {
...
+
+        guard !indexPath.isEmpty else { return }
+        let index = indexPath.removeFirst()
```
MouazAlzahabi, line 376: Why not just use indexPath.first/last ?

MouazAlzahabi (approved):

Multisig/UI/App/MainTabBarViewController.swift
DmitryBespalov: because it is not always 1 index in the index path, and last/first would return the same index if the number of indexes is 1

Issue #1579, opened 2021-11-10
[iOS] Txn details is not opened from push notification if the app was closed


liliya-soroka:
**Bug description**
[iOS] Txn details is not opened from push notification if the app was closed
**Steps To Reproduce**
1. Close safe app
2. execute tx or create tx to receive notification on ios device
3. click push notification
4. enter passcode

Current result: The app is opened on the balance page

**Expected Result**
Txn details should be opened if the push notification was pressed even the app was closed

**Device & App version**
 - Device/OS: iphone 11/14.8
 - App version: 3.3.0 ( 672)
 
**Environment**
prod, rinkeby
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.8.0 (928)

PR #2036, opened 2022-03-03
gh-1579 Delayed opening tx details

DmitryBespalov:
Handles #1579 

- Patched it so that tx details would open

### DIFF
--- a/Multisig/UI/App/MainTabBarViewController.swift
+++ b/Multisig/UI/App/MainTabBarViewController.swift
@@ -339,7 +339,7 @@ class MainTabBarViewController: UITabBarController {
     
     private func showTransactionDetails(transactionDetails: SCGModels.TransactionDetails) {
         let vc = ViewControllerFactory.transactionDetailsViewController(transaction: transactionDetails)
-        present(vc, animated: true, completion: nil)
+        presentWhenReady(vc)
     }
 
     private var delayedPresentationViewController: UIViewController?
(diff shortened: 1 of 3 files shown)


Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -332,12 +334,29 @@ class MainTabBarViewController: UITabBarController {
...
         let vc = ViewControllerFactory.transactionDetailsViewController(transaction: transactionDetails)
-        present(vc, animated: true, completion: nil)
+        presentWhenReady(vc)
+    }
+
+    private var delayedPresentationViewController: UIViewController?
+
+    private func presentWhenReady(_ vc: UIViewController) {
+        if view.window == nil {
+            delayedPresentationViewController = vc
```
elgatovital, line 349: how is it possible? If app is closed view.widow == nil, but thew view controller is not nil?

elgatovital (approved):

Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -332,12 +334,29 @@ class MainTabBarViewController: UITabBarController {
...
     private func showTransactionDetails(transactionDetails: SCGModels.TransactionDetails) {
         let vc = ViewControllerFactory.transactionDetailsViewController(transaction: transactionDetails)
-        present(vc, animated: true, completion: nil)
+        presentWhenReady(vc)
+    }
+
+    private var delayedPresentationViewController: UIViewController?
+
+    private func presentWhenReady(_ vc: UIViewController) {
+        if view.window == nil {
```
elgatovital, line 348: if app is closed view.window == nil but MainTabBarViewController still receives notificaiton callbacks?

DmitryBespalov: this condition means that the controller is not in the window hierarchy, i.e. it is not on screen yet. This happens when the animation is not started. When notification is received, the app is not fully functional, that's why it needs delay until the tab bar is shown

Issue #2028, opened 2022-03-02
[iOS][safe creation]Notification text is not seen fully if the safe name is long


liliya-soroka:
**Bug description**
[iOS][safe creation]Notification text is not seen fully if the safe name is long
**Steps To Reproduce**
1. create a safe with a long name on the mobile
2. check notification after safe creation 

**Expected Result**
The full notification text should be seen 
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/f7a8c5eb-bab2-47ed-83f0-08f71a1bc01b)
**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.8.0 (911)
 
**Environment**
prod , rinkeby 
**Additional info**
(Optional)

Relates to :
https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/gnosis/safe-ios/1951

liliya-soroka:
Verified
app version 3.8.0 (933)

PR #2037, opened 2022-03-03
GH-2028 Ellipsize name if too long

biafra23:
Handles #2028

- If the safe name is longer than 14 characters, take the first 14 character trim any whitespaces and append "…"

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/SafeDeploymentNotificationController.swift
+++ b/Multisig/UI/Safe Management/Create Safe/SafeDeploymentNotificationController.swift
@@ -18,12 +18,13 @@ class SafeDeploymentNotificationController {
         let adressString = "\(shortNamePrefix)\(safeAddress)"
         let chainName = safe.chain!.name ?? ""
         let safeName = safe.name ?? safeAddress
+        let safeNameForTitle = safeName.ellipsize()
 
         let notificationId: String
         let content = UNMutableNotificationContent()
         switch safe.safeStatus {
         case .deployed:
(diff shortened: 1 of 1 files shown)


Multisig/UI/Safe Management/Create Safe/SafeDeploymentNotificationController.swift
```diff
@@ -57,11 +58,22 @@ class SafeDeploymentNotificationController {
...
         let chainId: String = userInfo["chainId"] as! String
         let route = NavigationRoute.showAssets(address, chainId: chainId)
         DefaultNavigationRouter.shared.navigate(to: route)
     }
 }
+
+extension String {
+    func ellipsize(maxLength: Int = 14) -> String {
+        if self.count > maxLength {
+            if var shortString = self.substr(0, maxLength) {
```
DmitryBespalov: Do you think you could use the method that already exists in the String - self.prefix(maxLength)?

Multisig/UI/Safe Management/Create Safe/SafeDeploymentNotificationController.swift
```diff
@@ -57,11 +58,22 @@ class SafeDeploymentNotificationController {
...
+                return shortString.trimmingCharacters(in: .whitespaces) + "..."
```
DmitryBespalov: I would replace 3 dots with unicode for ellipsis: `…`

DmitryBespalov (approved):

Issue #2116, opened 2022-03-24
[iOS] [WC] Wallet description is empty or unknown for some wallets


liliya-soroka:
**Bug description**
Wallet description is empty or unknown for some wallets
**Steps To Reproduce**
1. connect to Trust or Argent wallet key
2. open connection details to check the description
**Current result:**
The description is empty or unknown 
**Expected Result**
Show wallet name instead of a description if it's not available

**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.9.0 (967)

![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/b1ddcddb-8bbd-492a-81b4-9be58dfeea2f)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/2874fb68-8339-4d50-80d3-3da49a52776b)
 
**Environment**
staging
**Additional info**
(Optional)

Relates to :
https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/gnosis/safe-ios/2053

liliya-soroka:
Verified
app version 3.9.0 (970)

PR #2118, opened 2022-03-24
gh-2116 Show wallet name in description if description is missing

elgatovital:
Handles #2116


Multisig/Features/Connect to Web/UI/Details/WebConnectionDetailsViewController.swift
```diff
@@ -167,7 +167,7 @@ class WebConnectionDetailsViewController: UITableViewController, WebConnectionOb
         case .description:
             let cell = contentCell()
             cell.setText("Description")
-            let text = peer.description ?? "Unknown"
+            let text = peer.description ?? peer.name
```
DmitryBespalov: is the reason that the description shows empty the null text or is that the text string itself is empty?

elgatovital: right, the second one

DmitryBespalov (approved):

Issue #2107, opened 2022-03-23
[iOS] Add more tracking parameters for WalletConnect events


sche:
<table>
  <tr>
    <th>event</th>
    <th>params</th>
  </tr>
  <tr>
    <td>user_owner_connected_external_wallet</td>
    <td>[new] wallet (String): name of the wallet (first 100 chars)</td>
  </tr>
  <tr>
    <td>user_transaction_confirmed_walletconnect</td>
    <td>
        [already implemented] chain_id (String): Chain id<br>
        [new] wallet (String): name of the wallet (first 100 chars)
    </td>
  </tr>
  <tr>
    <td>screen_exec_tx_submitted</td>
    <td>
        [already implemented] chain_id (String): Chain id<br>
        [already implemented] keyType: one of [imported, generated, wallet_connect, ledger_nano_x]<br>
        [new] wallet (String?): name of the wallet (first 100 chars). Present only for wallet_connect key type
    </td>
  </tr>
  <tr>
    <td>incoming_transaction_confirmed_wc</td>
    <td>
        [already implemented] chain_id (String): Chain id<br>
        [new] wallet (String): name of the wallet (first 100 chars)
    </td>
  </tr> 
</table>

liliya-soroka:
1. user_owner_connected_installed_wallet is used instead of user_owner_connected_external_wallet

2. No connected wallet name - incoming_transaction_confirmed_wc ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/6fd9a6bd-aaa6-4bde-9c7f-f25a7510680b)

3. Metamask was not recognized as a connected wallet ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/9465333d-47db-4f4c-b013-1349949a16ce)
3.11.0

PR #2127, opened 2022-03-28
gh-2107 WC Tracking: track wallet name

elgatovital:
Handles #2107


Multisig/UI/Transaction/Execution/ReviewExecutionViewController.swift
```diff
@@ -583,10 +583,17 @@ class ReviewExecutionViewController: ContainerViewController {
...
             }
 
-            successVC.trackingParams = [
+            let trackingParams = [
                 "keyType": trackedKeyType,
                 "chainId": self.controller.chainId
             ]
+            
+            if key.keyType == .walletConnect {
+                let walletName = key.wallet?.name ?? "Unknown"
```
DmitryBespalov: this will fail for the connections via qr code, please see my other pr - we would need to use the connection's remote peer name to get the name in case we don't have 'wallet'

DmitryBespalov (approved):

Issue #1505, opened 2021-10-06
[iOS] Execution with key connected via WC doesn't work


liliya-soroka:
**Bug description**
[iOS] Execution with key connected via WC doesn't work. The error is displayed 
**Steps To Reproduce**
1. install Rainbow mobile app on the same device with the safe mobile app
2. install mobile MM on another device
3. connect Rainbow key via WC in the mobile app using the Rainbow mobile app on the same device
4. connect MM key using QR during connection via WC
5. sign tx using added key
6. try to Execute tx using rainbow key and MM key

Current result: The errors appear and the tx is not executed

**Expected Result** The execution should work or should be hidden 

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/1c601de4-c25a-4fa8-b699-f3a0687284d3)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/298defa7-e97f-475b-832e-bca8c1f930c9)
**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.2.0 (617)
 
**Environment**
prod , rinkeby , BSC
**Additional info**
(Optional)

liliya-soroka:
@sche , could you check if it works for you in (617) app version

liliya-soroka:
Non-fatal errors from the firebase - 
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/ios:io.gnosis.multisig.prod.mainnet/issues/070d43e5607a6db5a8001443219ab2ab?time=last-twenty-four-hours&sessionEventKey=9512a06542994d9984b4710495bc2ece_1594450830439362389

https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/ios:io.gnosis.multisig.prod.mainnet/issues/070d43e5607a6db5a8001443219ab2ab?time=last-twenty-four-hours&sessionEventKey=12c31084444545c19c17c52d298aee55_1594449304697186041  - this non-fatal error appears only when key connected via WC is used (Domain: iOSError Code: 1300 NSLocalizedDescription: Failed to load transaction details: Could not connect to the server. (Error 1300) ) I don't think that it's Wifi or server problem

sche:
I also met this issue (but can't reproduce it)
Mostly execution works well for me.

liliya-soroka:
Verified
app version 3.9.0 (981)

PR #2134, opened 2022-03-28
gh-1505 Fixing the error and crashes

DmitryBespalov:
- Wallet is optional
- Wallet was lost when entry registries recreated

Handles #1505 

- Wallet made optional in sign/sendtx requests
- Restore the key-entry relationship when entries are recreated during connecting multiple wallets

### DIFF
--- a/Multisig/Features/Connect to Wallet/UI/PendingWalletActionViewController.swift
+++ b/Multisig/Features/Connect to Wallet/UI/PendingWalletActionViewController.swift
@@ -26,11 +26,15 @@ class PendingWalletActionViewController: ContainerViewController, UIAdaptivePres
     var requestTimeout: TimeInterval = 120
 
     var walletName: String {
-        wallet?.name ?? connection.remotePeer?.name ?? "wallet"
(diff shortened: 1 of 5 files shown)


Multisig/Features/Connect to Wallet/UI/PendingWalletActionViewController.swift
```diff
@@ -26,11 +26,15 @@ class PendingWalletActionViewController: ContainerViewController, UIAdaptivePres
...
     var walletName: String {
-        wallet?.name ?? connection.remotePeer?.name ?? "wallet"
+        wallet?.name ?? connection?.remotePeer?.name ?? "wallet"
     }
 
     override func viewDidLoad() {
         super.viewDidLoad()
+
+        if let keyInfo = keyInfo, connection == nil {
+            connection = WebConnectionController.shared.walletConnection(keyInfo: keyInfo).first
```
biafra23, line 36: connection will not be overridden if it wasn't nil.  What if keyInfo.connections changes? Or is that impossible?

biafra23 (approved):
🚀

Issue #1986, opened 2022-02-25
[iOS] Your tx is submitted with broken view details is displayed twice if tx details opened from push notification above "Your tx is submitted"


liliya-soroka:
**Bug description**
[iOS] Your tx is submitted with broken view details is displayed twice if tx details opened from push notification above "Your tx is submitted"
**Steps To Reproduce**
1. Create tx
2. Confirm ,execute and leave "Your tx is submitted" opened
3. send the app to the background and wait for the push notification
4. click push notification
Current result: Your tx is submitted is displayed instead of tx details

5. click "View details"
6. click "View details" on the next "Your tx is submitted" screen
Current result: The view button doesn't work but the back link appears in the header

7. click back
Current result: Tx details with old data is displayed so the user can submit the same tx again  and the user will pay for it 

**Expected Result**
Close the old "Your tx is submitted" and open tx details with up to date info
**Screenshots**
https://drive.google.com/file/d/1KS2HBGBN7M9tl2AwW4zzEHlpXBtqeUDm/view?usp=sharing
**Device & App version**
 - Device/OS: iPhone 11/14.8
 - App version: 3.7.0 (884)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.9.0 (1006)

PR #2151, opened 2022-03-31
gh-1986 Changed presentation sequence

DmitryBespalov:
Handles #1986

- Dismiss any controller that is already presented

### DIFF
--- a/Multisig/UI/App/MainTabBarViewController.swift
+++ b/Multisig/UI/App/MainTabBarViewController.swift
@@ -347,8 +347,12 @@ class MainTabBarViewController: UITabBarController {
     private func presentWhenReady(_ vc: UIViewController) {
         if view.window == nil {
             delayedPresentationViewController = vc
-        } else {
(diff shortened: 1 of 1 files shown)


elgatovital (approved):

Multisig/UI/App/MainTabBarViewController.swift
```diff
@@ -347,8 +347,12 @@ class MainTabBarViewController: UITabBarController {
     private func presentWhenReady(_ vc: UIViewController) {
         if view.window == nil {
             delayedPresentationViewController = vc
-        } else {
+        } else if presentedViewController == nil {
             present(vc, animated: true)
+        } else {
+            dismiss(animated: true) { [unowned self] in
```
elgatovital, line 353: would it be better to set animated to false here, so that already presented vc is dismissed without drawing too much attention?

DmitryBespalov: it might be too abrupt, tbh

Issue #2145, opened 2022-03-30
[iOS] [wc]Scan QR code is displayed on reconnect the key if a few accounts from the same wallet is added 


liliya-soroka:
**Bug description**
Scan QR code is displayed on reconnecting the key if a few accounts from the same wallet is added 
**Steps To Reproduce**
1. Use MM mobile app with a few accounts on the same device 
2. add the accoun1 from MM key to the safe mobile app
3. switch account in the MM mobile app to account 2
4. go to the mobile app and add a new owner key from MM ( account 2)
5. Open account 1 key details in the safe mobile app
6. try to connect from the owner key details

**Current result:** 
Scan QR code screen is opened instead of connection dialog to the MM mobile app

**Expected Result**
Connection dialog to the MM mobile app should be used instead of scan QR code
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/15.4
 - App version: 3.9.0 (985)
 
**Environment**
staging
**Additional info**
(Optional)

Relates to :
https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/gnosis/safe-ios/2108

sche:
@liliya-soroka does it work in the previous app version, or do we have the same issue there?

liliya-soroka:
It was not possible to have a few connections to the same wallet in the WC v1

liliya-soroka:
Verified
app version 3.9.0 (1006)

PR #2153, opened 2022-03-31
gh-2145 Fix scheme issue

DmitryBespalov:
Handles #2145

- Changed 1 to 1 relationship to 1 to Many (1 wallet can have many keys connected)

### DIFF
--- a/Multisig/Features/Connect to Wallet/Logic/WCAppRegistryRepository.swift
+++ b/Multisig/Features/Connect to Wallet/Logic/WCAppRegistryRepository.swift
@@ -46,11 +46,11 @@ class WCAppRegistryRepository {
         guard let cdEntries = (try? CDWCAppRegistryEntry.getAll()) else {
             return
         }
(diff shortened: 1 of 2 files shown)


elgatovital (approved):

Multisig/Features/Connect to Wallet/Logic/WCAppRegistryRepository.swift
```diff
@@ -46,11 +46,11 @@ class WCAppRegistryRepository {
         guard let cdEntries = (try? CDWCAppRegistryEntry.getAll()) else {
             return
         }
-        var keysByEntry: [String: KeyInfo] = [:]
+        var keysByEntry: [String: NSSet] = [:]
```
elgatovital, line 49: can we be more specific here and restrict type of collection?

DmitryBespalov: it's an objective-c NSSet from the core data, I think it's OK

Issue #1576, opened 2021-11-10
[iOS]Txn details page blinks if the txn signing was canceled  or accepted on Ledger


liliya-soroka:
**Bug description**
Txn details page blinks if the txn signing was canceled  or accepted on Ledger
**Steps To Reproduce**
1. Click confirm tx with Ledger key
2. Select "Cancel signing" on Ledger/ (accept signing)
3. Close select device screen
Current result: Txn details page blinks 
https://drive.google.com/file/d/1Gavk4QzIMhpccosHz0lFh5Osq6BZEKGF/view?usp=sharing
**Expected Result**
In my opinion, txn details page should be opened if the signing message was canceled on ledger or at least remove blinking of txn details 

**Device & App version**
 - Device/OS: Iphone X/ 14.8
 - App version: 3.3.0 (669)
 
**Environment**
prod, rinkeby
**Additional info**
(Optional)

sche:
Does it blink or does it reload data, but very fast?

liliya-soroka:
Verified
app version 3.9.0 (1006)

PR #2165, opened 2022-04-01
Gh 1576/ledger cancel sign

DmitryBespalov:
Handles #1576

- Don't reload transaction details when cancelled signing on ledger or in wallet connect because data is not changed.


Multisig/UI/Transaction/TransactionDetailsViewController/TransactionDetailsViewController.swift
```diff
@@ -347,12 +344,20 @@ class TransactionDetailsViewController: LoadableViewController, UITableViewDataS
                 Tracker.trackEvent(.reviewExecutionLedger)
             })
 
+            // needed to fix 'blinking' issue (screen reloads) when
+            // cancelling ledger signing on the device.
+            // Doing this via this variable 'patch' so that we don't rewrite the ledger implementation just yet.
+            var didSign = false
```
MouazAlzahabi, line 350: I would make this as return parameter from the onClose closure instead of this

DmitryBespalov: my point in the comment was that I don't want to touch the ledger implementation right now but rather restructure it later. That's why I'm using this variable here now.

biafra23 (approved):

Issue #2168, opened 2022-04-01
Make it harder to remove all app data when clicking on forgot passcode


DmitryBespalov:
Two times in the past we had an issue when a phone accidentally deleted all app content when it was in an internal pocket during a sports activity.

To make this less probable, we should request the user to type something in the form field so that we have active confirmation from a person.

Proposal is to change the confirmation to reset all data to the following:

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/c0b2847a-9ae3-4b82-bc74-a51c8ccf59e2" width=400/>

In addition to that, after 5 failed password attemtps, display error message:
`X failed password attempts. You can reset password via "Forgot passcode?" button below.`

liliya-soroka:
Verified
app version 3.9.0 (1011)

Issue #1860, opened 2022-01-20
All data is removed is displayed above tx details


liliya-soroka:
T48507: Forgot passcode during execution
https://gno.testrail.io/index.php?/tests/view/48507

Description:
When the forgot passcode is used during execution or tx creation the all data are removed but the tx details are not closed 

Steps:
1. create or execute tx using imported/generated key with passcode protection
2. Click forgot passcode and confirm data deleting

Current result: Error message about removed data appears above tx creation details page or tx execution page

Expected result: The pop-up window with tx details should be closed before the message as was done for tx details opened from push notifications

app version 3.6.0 (822)
ios : iPhone 11/14.8

screenshot:
![image](https://user-images.githubusercontent.com/338622/150390120-363bf4d4-7d88-4fdd-ac19-589ec97be5e9.png)

liliya-soroka:
Verified
app version 3.9.0 (1013)

PR #2169, opened 2022-04-01
gh-1860 Fixed opened screen and changed forgot passcode confirmation

DmitryBespalov:
GH-1860 close any open screen after data removal
GH-2168 change passcode reset confirmation prompt

Handles #1860
Handles #2168

### DIFF
--- a/Multisig/UI/Settings/Passcode/PasscodeViewController.swift
+++ b/Multisig/UI/Settings/Passcode/PasscodeViewController.swift
@@ -323,10 +323,16 @@ class EnterPasscodeViewController: PasscodeViewController {
 
     override func didTapButton(_ sender: Any) {
         let alertController = UIAlertController(
-            title: nil,
-            message: "You can disable your passcode. This will remove all data from the app.",
-            preferredStyle: .actionSheet)
+            title: "Remove all content",
+            message: "Disabling the passcode will remove all app content. This cannot be undone. Please type in \"Remove\" to continue.",
+            preferredStyle: .alert)
+
+        alertController.addTextField()
+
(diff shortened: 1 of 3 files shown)


Multisig/UI/Settings/Passcode/PasscodeViewController.swift
```diff
@@ -323,10 +323,16 @@ class EnterPasscodeViewController: PasscodeViewController {
...
-            message: "You can disable your passcode. This will remove all data from the app.",
-            preferredStyle: .actionSheet)
+            title: "Remove all content",
+            message: "Disabling the passcode will remove all app content. This cannot be undone. Please type in \"Remove\" to continue.",
+            preferredStyle: .alert)
+
+        alertController.addTextField()
+
         let remove = UIAlertAction(title: "Disable Passcode", style: .destructive) { [unowned self] _ in
+            guard alertController.textFields?.first?.text?.trimmingCharacters(in: .whitespacesAndNewlines) == "Remove" else {
```
MouazAlzahabi, line 340: I think we should use .lowerCase before comparing

MouazAlzahabi (changes requested):

Multisig/UI/Settings/Passcode/PasscodeViewController.swift
DmitryBespalov: I thought about this and thought that the purpose of this prompt is to prevent accidental deletion of the content. And I think asking to enter "Remove" exactly is good enough, we should not make it even more easier

biafra23 (approved):

elgatovital (approved):

Issue #2149, opened 2022-03-30
[iPad] When using app in Slide over mode, qr code scanner stays black


biafra23:
It is not possible to scan a QR code (to add a safe, use connected to web or WalletConnect), when app is used in slide over or any of the other split screen modes. 

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/9ce44df7-69c6-4ff8-9c00-69dc9cf2af81)

sche:
I put `minor` because there is a way to fulfill the use case, and the iPad version is not a priority at the moment. But will be good to fix it :)

DmitryBespalov:
https://developer.apple.com/documentation/avkit/accessing_the_camera_while_multitasking

liliya-soroka:
Verified by Dirk in 3.9.0 rc

Issue #2148, opened 2022-03-30
[iPad] QR code scanner picture rotated by 90° in landscape mode


biafra23:
**Bug description**
When using the qr code scanner in landscape mode, the X and Y axis are confused. The picture shown on the screen is rotated 90°. This makes it really tough to scan the qr code.

I don't know how to describe this better. Please try to scan a QR code in landscape mode. 
 
**Device & App version**
 - Device/OS: [e.g. iPhone8/iOS13.4.1]
 - App version: 3.9.0 (990)

DmitryBespalov:
A picture is worth a thousand words :wink:

liliya-soroka:
Verified by Dirk in 3.9.0 rc

PR #2175, opened 2022-04-04
Gh 2148/ipad camera rotation

DmitryBespalov:
Handles #2148
Handles #2149

- Apply correct transformation matrix when view is on screen and after interface rotation
- Tell that camera is unavailable if the camera stream is interrupted


biafra23 (approved):
Works as described

Issue #2177, opened 2022-04-04
Change 'add owner key' banner text on balances screen


DmitryBespalov:
Change text to:

"Did you know that you can import your owner key to sign and execute transactions on the go?"

JagoFigueroa:
All bueno, thanks! Tested and verified on v 3.10.0 (1036)

PR #2185, opened 2022-04-05
Gh 2177/change 'add owner key' banner text on balances screen

MouazAlzahabi:
Handles #2177 

And removed BalancesViewControllerV2


DmitryBespalov (approved):
Can you also find where the deleted classes was used? I think they're used in the snapshot tests

elgatovital (approved):

Issue #2154, opened 2022-03-31
Polygon gas estimations is wrong for executing transactions


DmitryBespalov:
At least when creating a safe, the gas estimation in the app is too low. User can submit if manually set the gas parameters.

liliya-soroka:
@DmitryBespalov , there is a new error for the tx execution on Polygon 
<img width="394" alt="image" src="https://user-images.githubusercontent.com/338622/176874135-411f5cf4-6685-4879-b523-ceda40ebfa72.png">

biafra23:
https://safe-client.staging.gnosisdev.com/v1/chains/137
https://gasstation-mainnet.matic.network/

biafra23:
Unfortunately Infura does not provide an error message, when using `eth_call` with "`pending`" as the block specifier. To debug this, I will use `latest` from now on. 
Why are we using pending? Do we have to use pending?

liliya-soroka:
Verified
Tx execution and safe creation both work on polygon in the latest version without the error 

app version 3.17.0 (1346)

PR #2200, opened 2022-04-07
Rename input to data

MouazAlzahabi:
Handles #2154


elgatovital:
how does this fix the issue?

biafra23:
> how does this fix the issue?

Same question from me.

DmitryBespalov:
Mouaz, did you test on all networks?

DmitryBespalov:
This changes fixes it because the API gives incorrect estimation when we send the "input" parameter name instead of "data" parameter name

DmitryBespalov (approved):

DmitryBespalov:
Fails on Polygon: "transaction underpriced".

This is most probably because our miner fee is too low.

DmitryBespalov:
I would close this PR for now and use it later to revisit our transaction estimation completely at another time.

Issue #2180, opened 2022-04-05
[Backup] Intro Screen


DmitryBespalov:
## Description
A screen that explains to the user the need to backup the imported key.

## What to do
* After a user enters the generated key name, the App suggests to set up a passcode if it is not set up yet.
* After setting up a passcode or skipping it, the app shows the Backup tips screen
* The screen can not navigate back. Users can only close the modal.
* Not now leads to generated key details screen in the same modal window. After closing the modal user stays on a screen where the flow was initiated.

## Acceptance Criteria
1. Backup intro screen opens
1. Possible to skip creating backup
1. Next/start button opens the 'Show Secret Phrase' screen.

## Links
[Requirements](https://docs.google.com/document/d/1v3v_XQVu4JGMSx0vMnXLISd_ZJnhEYOh0I3k7uY7MB8/edit#)
[Design](https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=745%3A10442)

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/aa4404f0-a68d-42bf-b3a8-119356101154)

JagoFigueroa:
All good! tested and verified on v 3.10.0 (1040) :)

PR #2212, opened 2022-04-12
gh-2180 Backup intro screen

elgatovital:
Handles #2180

- Custom view for the list of tips
- Backup intro screen
- Adjust key generation flow

simulator_screenshot_3D5DA33C-F7AF-43E8-AFB1-45A022C459CB


Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/OnboardingGenerateKeyViewController.swift
```diff
@@ -57,11 +57,21 @@ class OnboardingGenerateKeyViewController: AddKeyOnboardingViewController {
...
+            
             let passcodeCompletion = { [unowned self, unowned vc] in
                 App.shared.snackbar.show(message: message)
-                let detailsVC = OwnerKeyDetailsViewController(keyInfo: keyInfo, completion: self.completion)
-                vc.show(detailsVC, sender: vc)
+                let backupVC = BackupIntroViewController()
+                backupVC.backupCompletion = { [unowned self, unowned vc] backup in
+                    if backup {
+                        //TODO: proceed to backup screen instead
+                        let detailsVC = OwnerKeyDetailsViewController(keyInfo: keyInfo, completion: self.completion)
```
biafra23: Why not proceed to the backup screen (ExportViewController) right now?

biafra23 (approved):

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/TipsView.swift
```diff
@@ -0,0 +1,51 @@
...
+//  TipsView.swift
+//  Multisig
+//
+//  Created by Vitaly on 11.04.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class TipsView: UINibView {
```
DmitryBespalov, line 11: Like!

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/BackupIntroViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+        
+        backupButton.setText("Back up manually", .filled)
+        cancelButton.setText("Not now", .plain)
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        if self.isMovingFromParent {
+            backupCompletion(false)
```
DmitryBespalov: Can you elaborate more on what is the intent of these lines?

elgatovital: If user goes back from backup intro screen, it would lead back to enter key name screen. Key is already added and attempt to enter something and save it again would lead to exception. This prevents navigation back to enter key name screen and treats back navigation as dismissal.

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/OnboardingGenerateKeyViewController.swift
elgatovital: it can be done

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/BackupIntroViewController.swift
DmitryBespalov: A, ok. I think it was mentioned but never written down or changed the UI: we should disable the back button in this screen because user should not be able to go back to editing name after that is done. You can double-check with @sche .

elgatovital: user can still swipe back even without back button. This code will handle it.

DmitryBespalov: the swipe gesture can be disabled. If it's not on the 'create passcode' screen as well, then we should change it.

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/TipsView.swift
```diff
@@ -0,0 +1,51 @@
...
+        titleLabel.setStyle(.primary)
+        backgroundView.backgroundColor = .backgroundPrimary
+        backgroundView.layer.cornerRadius = 8
+    }
+    
+    func setContent(title: String, tips: [String]) {
+        
+        titleLabel.text = title
+        tipsContainer.arrangedSubviews.forEach { $0.removeFromSuperview() }
+        for tip in tips {
```
MouazAlzahabi: Can we make this for as the pervious line? (forEach)

DmitryBespalov (approved):
Thanks! Looks good! 🚢

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/TipsView.swift
elgatovital: Will it make this block more readable or concise? What makes you like forEach better?

Issue #2181, opened 2022-04-05
[Backup] Reminder to backup


DmitryBespalov:
## Description
When a key is not backed up, the application shows warnings to remind and encourage user to backup the key.

## What to do
* App settings
  * If there is at least one key that requires a backup, App displays a lock indicator

* Keys list
  * Tapping on the “Not backed up” banner initiates a backup process in a modal.
  * After finishing the backup process the modal is closed and users don’t see a warning anymore on the keys list screen.

* Key details
  * Tapping on the “Back up now” button initiates a backup process in a modal.
  * After finishing the backup process the modal is closed and users don’t see a warning anymore on the key details screen.

* Existing users with generated keys should not see a suggestion to back up

## Acceptance Criteria
1. App Settings shows a warning if there is at least one not backed up key.
1. Owner Key list shows a warning for those keys that are not backed up.
1. Key details show a warning if this key was not backed up.
1. Key details have a way to create backup.
1. Tap on the Owner Key list item's warning starts the backup flow.
1. An owner key needs backup if:
  a. It is a generated key AND
  b. The backup process was not completed before.

## Links
[Requirements](https://docs.google.com/document/d/1v3v_XQVu4JGMSx0vMnXLISd_ZJnhEYOh0I3k7uY7MB8/edit#)
[Design](https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=745%3A10442)

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/a28c5a7a-87fe-48d0-a030-de36a97b160d)

DmitryBespalov:
@sche @TanyaEfremova can we measure how annoying this reminder is for users? None of the developers like it, really :-) Maybe we rather make the backup step required instead? All older keys we can mark as backed up, but for the new keys we would require backup as mandatory.

sche:
@DmitryBespalov  some popular wallets use even more annoying reminders. Security is a top priority for us, so IMO we should at least keep the banner.
Not sure how we can get more qualitative results. Do you have good ideas @TanyaEfremova ?

![IMG_0858](https://user-images.githubusercontent.com/2284385/162422525-6a3d88f9-4492-48e4-9588-a9459614724c.PNG)
![IMG_0860](https://user-images.githubusercontent.com/2284385/162422594-b0c2393b-efc7-4b50-a8b0-f5b2e538de56.PNG)

sche:
[...]

PR #2215, opened 2022-04-12
Gh 2181/reminder to backup

MouazAlzahabi:
Handles #2181


DmitryBespalov (approved):

Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -87,6 +91,10 @@ extension KeyInfo {
         }
     }
 
+    static func keysNeedBackup() -> [KeyInfo] {
```
elgatovital, line 94: not sure if

elgatovital: not sure if KeyInfo model should have such logic here. Maybe it could be placed in some repository class. 
Nevertheless I would suggest to rename keysNeedsBackup -> keysWithoutBackup

elgatovital (approved):

Issue #2183, opened 2022-04-05
[Backup] Confirm Secret Phrase screen and Success screen


DmitryBespalov:
# Description
This screen is supposed to ensure that the user correctly backed up the secret phrase.

# What to do
1. A component for a word with multiple appearances for different states (empty, normal, error, do we want 'succes'?)
2. Probably a page view controller to animate between word changes
3. Play with animation of moving words back and forth 
4. Update the content depending on the challenge status
5. Model/implement the screen logic as a "challenge" model
6. When successful, remember that the backup was successful for that key.
7. Database changes: store the date of the key backup (or the flag whether key has a backup, but date is preferable because it is more flexible for the future changes)
8. Success screen

# Acceptance criteria
* Tapping on a word places it on the box
* Tappin on a word in the box places it back 
* Tappin on the Continue button makes a check
* An error is displayed if the word was wrong
  * A user can restart the challenge
* The challenge should be new when restarting
* A user needs to check three random words
* Once the last word is checked, the backup process automatically finishes, and a user appears on a success screen.
## Links
[Requirements](https://docs.google.com/document/d/1v3v_XQVu4JGMSx0vMnXLISd_ZJnhEYOh0I3k7uY7MB8/edit#)
[Design](https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=745%3A10442)

![image.png](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/30b72948-1f24-4e56-8840-10f9110aed99)

JagoFigueroa:
Morning guys, I see that some words that are on the longer end don't fit properly on the boxes:

Example 1 | Example 2
----- | ------ 
![IMG_0529.PNG](https://images.zenhubusercontent.com/625e824cecd9c30f320dd106/e8b2218a-af5b-4200-add7-0376977f8701) | ![IMG_0533.PNG](https://images.zenhubusercontent.com/625e824cecd9c30f320dd106/fd38ca1a-a534-4f41-a2be-a3bb8f45a100)

What do you think if we make them a tiny bit bigger? I think that should do the trick.

Cheers!

DmitryBespalov:
@TanyaEfremova can we live with this for the initial release and address it afterwards? I don't know how the layout should change in case there are multiple or even one word that does not fit the specified width.

sche:
@JagoFigueroa could you please create a major bug?
[...]

PR #2218, opened 2022-04-13
Gh 2183/confirm seed

DmitryBespalov:
Handles #2183

- controller and word view implementing the verification

### DIFF
--- a/Multisig/UI/Transaction/Execution/SuccessViewController.swift
+++ b/Multisig/UI/Transaction/Execution/SuccessViewController.swift
@@ -1,5 +1,5 @@
 //
-//  TransactionSuccessViewController.swift
+//  SuccessViewController.swift
 //  Multisig
 //
 //  Created by Vitaly Katz on 14.12.21.
@@ -8,7 +8,7 @@
 
 import UIKit
 
-class TransactionSuccessViewController: UIViewController {
+class SuccessViewController: UIViewController {
 
     @IBOutlet weak var titleLabel: UILabel!
     @IBOutlet weak var bodyLabel: UILabel!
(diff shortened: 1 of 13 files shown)


biafra23:
When I tap the "Back up manually" button, the seed phrase screen is shown only a very short time. So I have to go back to it. And then there is no "backup button". Apparently I have to go back one more screen to the "back up key" screen. and the hit back up manually again. I guess you automatically advance because the button is missing? And will be added in future PR?


https://user-images.githubusercontent.com/246473/163205531-7b03248d-364d-4e6c-9d41-b78245b6fe36.mov

DmitryBespalov:
@dirk This is only this screen so far. The screen is not integrated - we don't have the "show seed" screen ready yet.

biafra23:
The success screen is also missing. Also part of a future PR?

DmitryBespalov:
No, the success screen is indeed missing, thank you! Good catch!

DmitryBespalov:
@biafra23 updated

biafra23 (approved):
LGTM. Pending Tanyas ok regarding the success screen

Multisig/UI/Transaction/Execution/SuccessViewController.swift
```diff
@@ -52,6 +52,10 @@ class TransactionSuccessViewController: UIViewController {
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        guard let trackingEvent = trackingEvent else {
```
MouazAlzahabi, line 55: why we need this ?

MouazAlzahabi (approved):

Multisig/UI/Transaction/Execution/SuccessViewController.swift
DmitryBespalov: in several places the success tracking event is different. I don't have a tracking event yet, so I put this to nil. Otherwise it would crash right?

DmitryBespalov:
I have added a task for lottie animation for success screen - this was from Tanya - #2221

Issue #2202, opened 2022-04-08
[Backup] Setup passcode adjustments


sche:
## Description
Setup passcode should be requested when adding Ledger Nano X and WalletConnect keys.

## What to do
* Request to setup passcode before displaying key details when adding Ledger Nano X key
* Request to setup passcode before displaying key details when adding WalletConnect key

## Acceptance Criteria
1. Add a new Ledger Nano X or WalletConnect key
=> After specifying the key name, the app should request to set up a passcode if it does not exist

## Links
[Requirements](https://docs.google.com/document/d/1v3v_XQVu4JGMSx0vMnXLISd_ZJnhEYOh0I3k7uY7MB8/edit#)

liliya-soroka:
Verified
app version 3.10.0 (1047)

PR #2219, opened 2022-04-13
Add createPasscode

MouazAlzahabi:
Handles #2202

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/OnboardingConnectOwnerKeyViewController.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/OnboardingConnectOwnerKeyViewController.swift
@@ -48,28 +48,43 @@ class OnboardingConnectOwnerKeyViewController: AddKeyOnboardingViewController {
             enterNameVC.address = address
             enterNameVC.badgeName = KeyType.walletConnect.imageName
 
-            enterNameVC.completion = { [unowned self] name in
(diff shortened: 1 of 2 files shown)


DmitryBespalov (approved):

Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/OnboardingConnectOwnerKeyViewController.swift
```diff
@@ -48,28 +48,43 @@ class OnboardingConnectOwnerKeyViewController: AddKeyOnboardingViewController {
             enterNameVC.address = address
             enterNameVC.badgeName = KeyType.walletConnect.imageName
 
-            enterNameVC.completion = { [unowned self] name in
+            enterNameVC.completion = { [unowned self, unowned enterNameVC] name in
                 let success = OwnerKeyController.importKey(connection: connection, wallet: wallet, name: name)
 
-                if !success {
+                if success {
```
elgatovital, line 54: to prevent code duplication this part could be extracted and placed in a passcode related handler/controller

DmitryBespalov: I think we should have a refactoring all related onboarding stuff in a separate issue. In every such screen is completion blocks and repetition. I have untangled some in my PR #2218 but I would do a separate issue to do this for all other places.

Issue #2203, opened 2022-04-08
[Backup] Seed phrase security adjustment


sche:
## Description
Adjust security for revealing seed phrase

## What to do
* It is possible to reveal a seed phrase if a passcode is not set. (current status quo)
* Biometry should not be accepted to reveal the seed phrase.
* The option “Require for exporting key” of passcode settings should be removed.

## Links
[Requirements](https://docs.google.com/document/d/1v3v_XQVu4JGMSx0vMnXLISd_ZJnhEYOh0I3k7uY7MB8/edit#)

JagoFigueroa:
Tested and verified in v 3.10.0 (1044). Thanks!

PR #2225, opened 2022-04-14
Disable biometry when exporting key

MouazAlzahabi:
Handles #2203

### DIFF
--- a/Multisig/Logic/AuthenticationController.swift
+++ b/Multisig/Logic/AuthenticationController.swift
@@ -42,7 +42,7 @@ class AuthenticationController {
         try accessService.registerUser(password: password)
         AppSettings.passcodeWasSetAtLeastOnce = true
 
-        AppSettings.passcodeOptions = [.useForLogin, .useForConfirmation, .useForExportingKeys]
+        AppSettings.passcodeOptions = [.useForLogin, .useForConfirmation]
 
         NotificationCenter.default.post(name: .passcodeCreated, object: nil)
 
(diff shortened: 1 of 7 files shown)


biafra23 (approved):

Multisig/Logic/AuthenticationController.swift
```diff
@@ -42,7 +42,7 @@ class AuthenticationController {
         try accessService.registerUser(password: password)
         AppSettings.passcodeWasSetAtLeastOnce = true
 
-        AppSettings.passcodeOptions = [.useForLogin, .useForConfirmation, .useForExportingKeys]
+        AppSettings.passcodeOptions = [.useForLogin, .useForConfirmation]
```
elgatovital, line 45: don't we want to protect key export with passcode?

Multisig/Logic/Models/AppSettings.swift
```diff
@@ -151,7 +148,6 @@ struct PasscodeOptions: OptionSet {
     static let useBiometry = PasscodeOptions(rawValue: 1 << 0)
     static let useForLogin = PasscodeOptions(rawValue: 1 << 1)
     static let useForConfirmation = PasscodeOptions(rawValue: 1 << 2)
-    static let useForExportingKeys = PasscodeOptions(rawValue: 1 << 3)
 
-    static let all: PasscodeOptions = [.useBiometry, .useForLogin, .useForConfirmation, .useForExportingKeys]
```
elgatovital, line 156: passcode option values are confusing here. It makes impression that key export is not protected by passcode anymore.

Multisig/Logic/Models/AppSettings.swift
```diff
@@ -151,7 +148,6 @@ struct PasscodeOptions: OptionSet {
     static let useBiometry = PasscodeOptions(rawValue: 1 << 0)
     static let useForLogin = PasscodeOptions(rawValue: 1 << 1)
     static let useForConfirmation = PasscodeOptions(rawValue: 1 << 2)
-    static let useForExportingKeys = PasscodeOptions(rawValue: 1 << 3)
```
DmitryBespalov, line 154: Can you add a comment that we had this flag as (1 << 3) value so that if we extend this parameter, existing app installation don't conflict with the deprecated value?

DmitryBespalov (approved):
Looks good! 🚀

elgatovital (approved):

Issue #2249, opened 2022-04-20
[AddOwner] Busines logic to add owner


DmitryBespalov:
# Overview
This is about implementing the required business logic to make the "Add Owner" transaction.

# What To Do
1. Create a SafeOwnerEditingController
2. Implement creating the 'safe' transaction for the AddOwnerWithThreshold call to the safe contract
3. Perhaps, also move the safe transaction estimation (using client gateway) and proposing transaction to that class.
4. Remember to handle errors with explanatory user-facing messages.

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/c1e5cff9-04c9-4eee-a40f-09efa0cd77b9)

# Acceptance criteria
* Integration tests created that would propose "add owner" transaction

liliya-soroka:
Verified
app version 3.11.0 (1081)

Issue #2248, opened 2022-04-20
[AddOwner] Review Add Owner Tx screen


DmitryBespalov:
# Overview
As a safe owner I want to review the change to the safe so I can double-check for the errors and go back to change something if necessary, and submit transaction when I'm ready.

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/a2b4fc7c-f1bb-40ce-8125-1f9ac1ffdae7" width=300 />

Figma: https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1099%3A16226

# What To Do
1. Build UI based on the existing 'change safe review tx' class
2. Connect to the business logic 
3. On completion call the navigation controller to proceed to the next screen.

# Acceptance Criteria
* Shows safe's address
* Has safe's network indicator (ribbon)
* Has flow screen number (2 of 2)
* Can edit transaction parameters (Safe Nonce) **Question: can we edit safe tx gas? was it missed last time?**
* Shows to-be-added owner key address
* Shows change in the confirmations if there was a change
* Shows change in number of safe owners (must be 1 more owner after transaction)
* Screen creates the transaction that can then be viewed in web interface, too

liliya-soroka:
Verified
app version 3.11.0(1081)

PR #2314, opened 2022-04-28
Gh 2258/views for review change safe

MouazAlzahabi:
Handles  #2248 #2249 
We need to refactor to split Review settings change into two view controllers (AddOwner, SwapOwner)


Multisig/Logic/Models/Transaction.swift
```diff
@@ -218,6 +218,23 @@ extension Transaction {
         return transaction
     }
 
+    static func settingsChangeTransaction(safeAddress: Address,
```
DmitryBespalov: we should inline this

DmitryBespalov: or delete - I think it's not used anymore

DmitryBespalov (approved):
minor change might be needed, otherwise we should merge and continue to tie all pieces together

Issue #2246, opened 2022-04-20
[AddOwner] Navigation controller for adding owner flow


DmitryBespalov:
# Overview
This task is about adding a UI navigation controller that would tie together the screens for adding an owner.

# What To Do
1. Class for a navigation controller
2. Opening/closing of screens for every step in the process, collecting and passing data from one screen to another.

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/415429ae-ffff-4714-a891-5f22c95bab1f" />

# Acceptance Criteria
* The whole flow to add owner works correctly, i.e. the screen sequence as in the activity diagram / figma

liliya-soroka:
Verified
app version 3.11.0 (1081)

PR #2317, opened 2022-05-02
Create navigation controller

MouazAlzahabi:
Handles #2246


DmitryBespalov (approved):

DmitryBespalov (changes requested):
Ok, I have checked and found one issue:

- Review screen has wrong number of confirmations

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/AddOwnerController.swift
```diff
@@ -65,7 +65,7 @@ class AddOwnerController: UINavigationController, UIAdaptivePresentationControll
                                                                 owner: key,
                                                                 oldOwnersCount: safe.ownersInfo?.count ?? 0,
                                                                 oldThreshold: Int(safe.threshold ?? 0),
-                                                                newThreshold: 5)
+                                                                newThreshold: Int(safe.threshold ?? 0))
```
DmitryBespalov: how is the old and new threshold different? We should get the new threshold from the previous screen and use it here

MouazAlzahabi: My bad, new it's solved

DmitryBespalov (approved):

PR #2325, opened 2022-05-06
refs/heads/use-password-text-field Use text field

DmitryBespalov:
Handles #

- Use text field instead of text area because it has the 'secure text entry' and I don't have to implement it as a custom thing for a text view. Before was also using a hack to support a placeholder text for a text view

no error | error hidden | error visible
-------- | ------------ | ------------
passphrase-entry | error-invisible | large-error


biafra23:
This is now single line and no way for the user to show, what they typed. Is this what we want?
As a user who sometimes types the seed phrase I see this as a step back.

DmitryBespalov:
I thought it would be ok, but ok 😄 I can make a button hiding/showing the text, but this can only work with the standard single line component - the previous text area doesn't have an option to secure the entered text (iOS limitation). I will also search for a library that might do it.

biafra23:
@DmitryBespalov Can't we use `GMTextFieldMultipleLines` ? 

Isn't this the component youre using? https://cocoapods.org/pods/GMTextField

DmitryBespalov:
@dirk I changed the screen to make it closer to the new designs https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1551%3A16671

biafra23:
@DmitryBespalov And there is no way to make this multiline?

biafra23:
Also on my iPhone SE the last sentence is cut off. See screenshot.

DmitryBespalov:
no, no way for multiline with the standard iOS controls

DmitryBespalov:
it is cut off because there's no space for the whole content. If you click anywhere except the text field, then it will hide keyboard and you'll see whole content

DmitryBespalov:
@biafra23 I have put a workaround to hide the keyboard when there's an error

DmitryBespalov:
I think I found a way to have a 'secure' text with multiline

Issue #2289, opened 2022-04-25
[ChangeConfirmations] Tracking


DmitryBespalov:
**Change Confirmations Tracking**

<table>
  <tr>
   <td><strong>Description</strong>
   </td>
   <td><strong>Event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
  </tr>
  <tr>
   <td>Change confirmation count screen
   </td>
   <td>screen_change_confirmations
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Review change confirmation transaction
   </td>
   <td>screen_review_change_confirmations
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Success of change confirmation transaction submission
   </td>
   <td>screen_change_confirmations_success
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>  
</table>

biafra23:
Let's split this into creating the events and emitting the events.

liliya-soroka:
Verified
app version 3.12.0

Issue #2286, opened 2022-04-25
[ReplaceOwnerSettings] Tracking


DmitryBespalov:
**Replace Owner Tracking**

<table>
  <tr>
   <td><strong>Description</strong>
   </td>
   <td><strong>Event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
  </tr>
  <tr>
   <td>Select owner to replace
   </td>
   <td>screen_replace_owner_select
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Review replace owner transaction
   </td>
   <td>screen_replace_owner_review
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Success of replace owner transaction submission
   </td>
   <td>screen_replace_owner_success
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>User initiated replace owner flow from safe settings screen
   </td>
   <td>user_replace_owner_from_settings
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Select new owner address to replace
   </td>
   <td>screen_replace_owner_select_new
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Specify new owner name when replacing from safe settings
   </td>
   <td>screen_replace_owner_new_owner_name
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
</table>

biafra23:
Let's split this into creating the events and emitting the events.

liliya-soroka:
Verified
app version 3.12.0

Issue #2279, opened 2022-04-25
[RemoveOwner] Tracking


DmitryBespalov:
**Remove Owner Tracking**

<table>
  <tr>
   <td><strong>Description</strong>
   </td>
   <td><strong>Event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
  </tr>
  <tr>
   <td>User initiated remove owner flow from safe settings screen
   </td>
   <td>user_remove_owner_from_settings
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Select owner to remove
   </td>
   <td>screen_remove_owner_select
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Change policy when removing owner
   </td>
   <td>screen_remove_owner_change_policy
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Success of remove owner transaction submission
   </td>
   <td>screen_remove_owner_success
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Review screen to remove owner
   </td>
   <td>screen_remove_owner_review
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
</table>

DmitryBespalov:
Should we track the threshold? (before/after) as a parameter?

biafra23:
Let's split this into creating the events and emitting the events.

liliya-soroka:
Verified
app version 3.12.0

Issue #2277, opened 2022-04-25
[AddOwnerSettigns] Tracking


DmitryBespalov:
**Add Owner Tracking**

<table>
  <tr>
   <td><strong>Description</strong>
   </td>
   <td><strong>Event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
  </tr>
  <tr>
   <td>Suggestion to add generated key as a safe owner
   </td>
   <td><strike>screen_add_gen_key_as_owner</strike> screen_add_key_as_owner_info
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Skip adding generated key as an owner
   </td>
   <td><strike>user_skip_add_gen_key_as_owner</strike> user_skip_add_key_as_owner
   </td>
   <td>NO
   </td>
  </tr>
  <tr>
   <td>Change policy when adding new owner
   </td>
   <td>screen_add_owner_change_policy
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Review add owner transaction
   </td>
   <td>screen_add_owner_review
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Success of add owner transaction submission
   </td>
   <td>screen_add_owner_success
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>User initiated add owner flow from safe settings screen
   </td>
   <td>user_add_owner_from_settings
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Select new owner address when adding from safe settings
   </td>
   <td>screen_add_owner_select_address
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
  <tr>
   <td>Specify new owner name when adding from safe settings
   </td>
   <td>screen_add_owner_specify_name
   </td>
   <td>chain_id (String): Chain id
   </td>
  </tr>
</table>

DmitryBespalov:
please check the tracking event name length max 40 symbols

biafra23:
Let's split this into creating the events and emitting the events.

elgatovital:
Different event names already in use 
`screen_add_key_as_owner_info` instead of `screen_add_gen_key_as_owner`
`user_skip_add_key_as_owner` instead of `user_skip_add_gen_key_as_owner`
 
everything else is implemented as described in this ticket.

[...]

PR #2330, opened 2022-05-09
GH-2277 Add new tracking event types

biafra23:
Partially handles #2277, #2286, #2279, #2289

- Add new tracking event types
- Does **not** call events to be emitted


biafra23:
If we split these tickets in creating the event types and emitting them later. Then we just emit them while developing the feature. And then close the respective ticket.

DmitryBespalov (approved):

Issue #2335, opened 2022-05-10
[iOS][Tracking] use screen_settings_tx_advanced_params instead of screen_asset_transfer_advanced_params for settings txs


liliya-soroka:
**Bug description**
use screen_settings_tx_advanced_params instead of screen_asset_transfer_advanced_params for settings txs
**Steps To Reproduce**
1. call replace or add owner review screen and open advanced 

Current result: advanced parameters opening is tracked as screen_asset_transfer_advanced_params
**Expected Result**
screen_settings_tx_advanced_params should be used if the advanced parameters are opened for settings transactions  with chain_id (String): Chain id

liliya-soroka:
Reopened:
screen_asset_transfer_advanced_params event is missed for the case when advanced parameters is opened during send assets flow or incoming tx from dapps

app version 3.12.0

biafra23:
@liliya-soroka I am unable to reproduce this. When I create a tx and then open "Advanced Parameters" I see the following in the logs: 

```
2022-05-18 02:37:18.512000 [INFO] ConsoleTracker.swift:15 track(event:parameters:): [TRACKING] event: 'screen_asset_transfer_advanced_params', parameters: ["chain_id": "4"] 
```
Can you give exact steps to reproduce this error?


> incoming tx from dapps

Incoming tx should not emit this event

DmitryBespalov:
@biafra23 the event name is wrong, it should be `....settings_advanced_params` but it is `asset_transfer`. Which is wrong in this case

biafra23:
@DmitryBespalov Liliya was talking about a sending funds tx creation. Not a settings changing transaction. 

> screen_asset_transfer_advanced_params event is missed for the case when advanced parameters is opened during **send assets flow** or incoming tx from dapps

DmitryBespalov:
@biafra23 oh I see, sorry. My bad not reading the comment

liliya-soroka:
Verified
app version 3.12.0

PR #2337, opened 2022-05-11
GH-2335 Use screen_settings_tx_advanced_params tracking event

biafra23:
Handles #2335

- Adding settingsTxAdvancedParams and using it it subclasses of ReviewSafeTransactionViewController that deal with settings changes
-

### DIFF
--- a/Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
+++ b/Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
@@ -25,7 +25,8 @@ class ReviewSafeTransactionViewController: UIViewController {
     @IBOutlet weak var ribbonView: RibbonView!
 
     private var currentDataTask: URLSessionTask?
-    
+    var trackingEvent: TrackingEvent = .assetsTransferReview
+
     var safe: Safe!
     var address: Address!
     var data: Data?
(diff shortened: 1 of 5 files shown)


Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
```diff
@@ -324,6 +323,11 @@ class ReviewSafeTransactionViewController: UIViewController {
         presentModal(ViewControllerFactory.modal(viewController: ribbon))
     }
 
+    // Please override in subclasses
+    func getTrackingEvent() -> TrackingEvent {
```
DmitryBespalov: interesting you chose method instead of a var with setting that var in an initializer, but this also works for me.

DmitryBespalov (approved):

Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
elgatovital: it would be better to extract basic functionality to an abstract class so that overriding this method is mandatory. Having .assetsTransferAdvancedParams as default value is not intuitive when reading this code.

biafra23: @elgatovital Swift does not support abstract classes. We could introduce a protocol or make the base implementation crash. Is it worth it?

DmitryBespalov: no it's not worth it for the protocol. I was saying that instead of a method we could use a property

biafra23: Would making it a property increase the visibility (Make it more intuitive) ?

biafra23: See: 0dd2b77ee0dfe6b65578b47272a5da8cd4a66eeb

DmitryBespalov (approved):

Issue #2266, opened 2022-04-22
[iOS] Confirm seed phrase - some words not fitting when backing up an owner key


JagoFigueroa:
**Bug description**
Some words do not fit correctly on the boxes when backing up an owner key

**Steps To Reproduce**
A bit tricky. Create a new owner key that when backing it up and asked to add it in the correct position, will be one of this larger words.

**Expected Result**
Every seed word is fully visible.

**Screenshots**
Example 1 | Example 2
----- | ------ 
![IMG_0529.PNG](https://images.zenhubusercontent.com/625e824cecd9c30f320dd106/e8b2218a-af5b-4200-add7-0376977f8701) | ![IMG_0533.PNG](https://images.zenhubusercontent.com/625e824cecd9c30f320dd106/fd38ca1a-a534-4f41-a2be-a3bb8f45a100)

**Device & App version**
 - Device/OS: Iphone 13 Pro / iOS 15.4.1
 - App version: v 3.10.0 (10444)
 
**Environment**
Prod

biafra23:
The easiest way to fix this is to scale the font. Is that ok? Will provide screenshots with examples.

liliya-soroka:
Verified
app version 3.12.0 (1110)

PR #2340, opened 2022-05-11
GH-2266 Scale font

biafra23:
Handles #2266

- Scale font

Before | After
-------|------
Image from iOS (3) | Image from iOS (4)


DmitryBespalov:
em... this works but the layout is kinda wrong, the 'proper' way would be to implement a collection view that would layout the items automatically when they don't fit the width

biafra23:
And the selected WordView just grows, then?

DmitryBespalov:
@biafra23 I think it's better to target this fix to  the main branch and not include it in the release already

elgatovital (approved):

biafra23:
@DmitryBespalov 
Let's keep this as a draft. And if we don't implement the new design, then we can take this as a workaround.

biafra23:
This will go to main now. If we merge it, we could keep it until we have the proper solution?

DmitryBespalov:
yes, I agree

Issue #2288, opened 2022-04-25
[ReplaceOwnerSettings] Swipe to replace in Safe Settings


DmitryBespalov:
# Overview
As a safe owner I want to replace one owner with another so I can change owner structure.

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/eacb7561-60d9-46b1-bdba-8aa76e57f838" width=300 />

Figma: https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1194%3A15160

# What To Do
1. Swipe action opens the "replace owner" flow

# Acceptance Criteria

- When selected, replacing the selected owner flow starts
- Works only if safe is not read-only.

liliya-soroka:
Verified
app version 3.12.0 (1123)

Issue #2283, opened 2022-04-25
[RemoveOwner] Business logic for removing an owner


DmitryBespalov:
# Overview
This is about implementing business logic for removing owner flow.

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/d96cc5ad-c630-47aa-8cd4-9f82dcd82ed3)

# What To Do
1. Use the "removeOwnerWithThreshold" Safe contract function
2. The logic is similar to addOwnerWithThreshold.
3. Write an integration test to remove an owner

# Acceptance Criteria

- Integration test is written and works
- Able to remove an owner from the safe

liliya-soroka:
Reopened:
app version 3.12.0(1123)
Integration tests are missed

DmitryBespalov:
@liliya-soroka integration tests are implemented https://github.com/safe-global/safe-ios/blob/44e83e3159da0e34948e09cce774d983b2b59ef1/MultisigIntegrationTests/SafeTransactionControllerIntegrationTests.swift#L130

liliya-soroka:
Verified
app version 3.12.0(1123)

Issue #2282, opened 2022-04-25
[RemoveOwner] Swipe to remove owner in Settings


DmitryBespalov:
# Overview
As a safe owner I want to remove owner so I can change owner structure.

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/78e7c5f4-9bba-4734-a7fb-3e3c6a72b9d0" width=300 />

Figma: https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1194%3A15160

# What To Do
1. Add 'swipe' actions for owner of the safe
2. Opens the RemoveOwnerController

# Acceptance Criteria

- Swipe to remove opens the Remove Owner flow
- "Remove" is only available for owner count more than 1. I.e. if only 1 owner, it should not be possible to remove it.

liliya-soroka:
Verified
app version 3.12.0 (1123)

Issue #2281, opened 2022-04-25
[RemoveOwner] Review Remove Owner Transaction screen


DmitryBespalov:
# Overview
As a safe owner I want to review the removing owner transaction so I can double check correctness before I submit transaction.

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/6f936f2c-b83f-4755-8a50-4a1d834d16c7" width=300 />

Figma: https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1099%3A16068

# What To Do
1. Subclass the `ReviewChangeSafeTxViewController` to build the review screen UI.

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/2ad2331c-277d-4fc2-a728-5a42077ae9ec)

# Acceptance Criteria

- Shows address of a removed owner
- Number of owners decreases
- If confirmations changed, shows the change
- Able to edit transaction parameters (safe nonce)
- Submitting creates a transaction in transaction queue.
- "2 of 2" shows it is a last step in the process.

liliya-soroka:
Verified
app version 3.12.0 (1123)

PR #2341, opened 2022-05-12
Gh 2283/business logic for removing an owner

MouazAlzahabi:
Handles #2283 #2280 #2281 #2282 #2288

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
@@ -79,6 +79,8 @@ class RemoveOwnerFlowFactory {
         confirmationsVC.stepNumber = step
         confirmationsVC.maxSteps = maxSteps
         confirmationsVC.trackingEvent = .addAsOwnerChangeConfirmations
+        confirmationsVC.promptText = "You’re about to remove an owner. Would you like to change the required confirmations?"
+        confirmationsVC.titleText = "Remove owner"
         confirmationsVC.completion = completion
         return confirmationsVC
     }
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
@@ -74,12 +74,12 @@ class RemoveOwnerFlow: UIFlow {
 class RemoveOwnerFlowFactory {
     func confirmations(step: Int, maxSteps: Int, safe: Safe, completion: @escaping (_ newConfirmations: Int) -> Void) -> EditConfirmationsViewController {
         let confirmationsVC = EditConfirmationsViewController()
-        confirmationsVC.confirmations = Int(safe.threshold ?? 0)
+        confirmationsVC.confirmations = Int(safe.threshold ?? 1)
         confirmationsVC.minConfirmations = 1
-        confirmationsVC.maxConfirmations = max(1, (safe.ownersInfo ?? []).count) - 1
+        confirmationsVC.maxConfirmations = max(1, (safe.ownersInfo ?? []).count - 1)
         confirmationsVC.stepNumber = step
         confirmationsVC.maxSteps = maxSteps
-        confirmationsVC.trackingEvent = .addAsOwnerChangeConfirmations
+        confirmationsVC.trackingEvent = .removeOwnerChangePolicy
         confirmationsVC.promptText = "You’re about to remove an owner. Would you like to change the required confirmations?"
         confirmationsVC.titleText = "Remove owner"
         confirmationsVC.completion = completion
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
@@ -16,10 +16,11 @@ class RemoveOwnerFlow: UIFlow {
     var newConfirmations: Int?
     var removeOwnerTransactionDetails: SCGModels.TransactionDetails?
 
-    internal init(owner: Address, safe: Safe, factory: RemoveOwnerFlowFactory = .init(), navigationController: UINavigationController, completion: @escaping (_ success: Bool) -> Void) {
+    internal init(owner: Address, prevOwner: Address?, safe: Safe, factory: RemoveOwnerFlowFactory = .init(), navigationController: UINavigationController, completion: @escaping (_ success: Bool) -> Void) {
         self.factory = factory
         self.safe = safe
         self.ownerToRemove = owner
+        self.prevOwner = prevOwner
         super.init(navigationController: navigationController, completion: completion)
     }
 
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
@@ -1,0 +1,110 @@
+//
+//  RemoveOwnerFlow.swift
+//  Multisig
+//
+//  Created by Moaaz on 5/11/22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class RemoveOwnerFlow: UIFlow {
+    var factory: RemoveOwnerFlowFactory
+    var safe: Safe
+    var ownerToRemove: Address?
+    var prevOwner: Address?
+    var newConfirmations: Int?
+    var removeOwnerTransactionDetails: SCGModels.TransactionDetails?
+
+    internal init(owner: Address, safe: Safe, factory: RemoveOwnerFlowFactory = .init(), navigationController: UINavigationController, completion: @escaping (_ success: Bool) -> Void) {
+        self.factory = factory
+        self.safe = safe
+        self.ownerToRemove = owner
+        super.init(navigationController: navigationController, completion: completion)
+    }
+
+    override func start() {
+        confirmations()
+    }
+
+    func confirmations() {
+        let confirmationsVC = factory.confirmations(
+            step: 1,
+            maxSteps: 2,
+            safe: safe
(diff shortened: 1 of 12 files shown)


Multisig/UI/UI Library/Views/OwnerActionView.swift
```diff
@@ -30,7 +30,7 @@ class OwnerActionView: UINibView {
         case .addingOwner:
             actionTag.set(title: "Adding owner", style: .footnote2, backgroundColor: .backgroundPositive, textColor: .primary)
         case .removingOwner:
-            actionTag.set(title: "Removing owner", style: .footnote2, backgroundColor: .backgroundPrimary, textColor: .labelSecondary)
+            actionTag.set(title: "Removing owner", style: .footnote2, backgroundColor: .backgroundError, textColor: .error)
```
DmitryBespalov, line 35: this is wrong: the designs are different for the 'removing owner' when this part is used in the "replace" and when it is used in the "remove" use cases. This change introduces a mistake for the "replace owner" design UI

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -274,6 +301,47 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
...
+
+    func remove(owner: Address) {
+        guard let navigationController = navigationController else {
+            return
+        }
+
+        removeOwnerFlow = RemoveOwnerFlow(
+            owner: owner,
+            safe: safe!,
+            navigationController: navigationController) { [unowned self] skippedTxDetails in
```
DmitryBespalov: you need to remove the reference to 'removeOwnerFlow' in the completion block because otheriwse it retains reference to a navigation controller

DmitryBespalov: Another comment: the variable in the completion is not used.

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -158,6 +160,30 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
         }
     }
 
+    func reloadSafeOwners() {
+        currentDataTask?.cancel()
```
DmitryBespalov, line 167: do you think it's good to reuse the same data task?

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -158,6 +160,30 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
...
+            updateSections()
+            return
+        }
+
+        currentDataTask = SafeTransactionController.shared.getOwners(safe: safe.addressValue, chain: safe.chain!) { [weak self] result in
+            guard let self = self else { return }
+
+            switch result {
+            case .failure(let error):
+                self.onError(GSError.error(description: "Failed to load safe owners", error: error))
```
DmitryBespalov, line 178: is this user-facing error? Does it trigger a snackbar? (If it does, then it might be confusing or irrelevant)

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -21,9 +21,12 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
     private var currentDataTask: URLSessionTask?
     private var sections = [SectionItems]()
     private var safe: Safe?
+    private var safeOwners: [AddressInfo] = []
```
DmitryBespalov, line 27: comment why we are loading it here instead of relying on safe info - would help for the future code readers

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
```diff
@@ -0,0 +1,112 @@
...
+
+class RemoveOwnerFlowFactory {
+    func confirmations(step: Int, maxSteps: Int, safe: Safe, completion: @escaping (_ newConfirmations: Int) -> Void) -> EditConfirmationsViewController {
+        let confirmationsVC = EditConfirmationsViewController()
+        confirmationsVC.confirmations = Int(safe.threshold ?? 0)
+        confirmationsVC.minConfirmations = 1
+        confirmationsVC.maxConfirmations = max(1, (safe.ownersInfo ?? []).count) - 1
+        confirmationsVC.stepNumber = step
+        confirmationsVC.maxSteps = maxSteps
+        confirmationsVC.trackingEvent = .addAsOwnerChangeConfirmations
```
DmitryBespalov: is tracking right here?

Is this code repeating something else or is it OK? Do we need to generalize "factory" classes?

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
```diff
@@ -0,0 +1,112 @@
...
+        show(successVC)
+    }
+}
+
+class RemoveOwnerFlowFactory {
+    func confirmations(step: Int, maxSteps: Int, safe: Safe, completion: @escaping (_ newConfirmations: Int) -> Void) -> EditConfirmationsViewController {
+        let confirmationsVC = EditConfirmationsViewController()
+        confirmationsVC.confirmations = Int(safe.threshold ?? 0)
+        confirmationsVC.minConfirmations = 1
+        confirmationsVC.maxConfirmations = max(1, (safe.ownersInfo ?? []).count) - 1
```
DmitryBespalov: this can become 0 which is wrong. It always must be at least 1

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
```diff
@@ -0,0 +1,112 @@
...
+            stop(success: !showTxDetails)
+        }
+        show(successVC)
+    }
+}
+
+class RemoveOwnerFlowFactory {
+    func confirmations(step: Int, maxSteps: Int, safe: Safe, completion: @escaping (_ newConfirmations: Int) -> Void) -> EditConfirmationsViewController {
+        let confirmationsVC = EditConfirmationsViewController()
+        confirmationsVC.confirmations = Int(safe.threshold ?? 0)
```
DmitryBespalov: always at least 1 confirmation is required

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
biafra23: IMHO this should be `.removeOwnerChangePolicy`

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
```diff
@@ -0,0 +1,112 @@
...
+            safe: safe,
+            owner: ownerToRemove!,
+            prevOwner: prevOwner, newThreshold: newConfirmations!) { [unowned self] txDetails in
+                removeOwnerTransactionDetails = txDetails
+                success()
+            }
+        show(reviewVC)
+    }
+
+    func success() {
```
DmitryBespalov, line 59: this seems to be repeating in several places, maybe we can generalize the owner change flows if that's feasible and has enough common shared code

biafra23:
We just found out, that `.removeOwnerSelect` is never use, because we do not select the user to be removed from a list. Could you delete this entry from `TrackingEvents.swift` line 287?

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
MouazAlzahabi: I thought it would be good as these two requests are called one after another, wdyt?

DmitryBespalov: just wanted to know your thinking. I think it's OK. If there would be 2 data tasks, then we'd need to cancel the 'get owner' when the 'get safe info' is cancelled or restarted.

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
MouazAlzahabi: What do you suggest instead ?

DmitryBespalov: maybe fall back to the result of the safe info and disable the 'replace/remove/add' actions?

MouazAlzahabi: I thought about it first, then I realized that it might be confusing for the user to not know why this options are disabled, no?

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/RemoveOwnerFlow.swift
MouazAlzahabi: I think this will require a larger refactor right ?, I would go with making it in another ticket

DmitryBespalov: ok

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
DmitryBespalov: and it is similarly confusing to see an error when you're loading the safe that owners are not there. Well, actually, will this just show a snackbar or will it show full screen "can't load safe" error?

DmitryBespalov (approved):
I'd suggest to merge and move the 'error handling' for this screen (when owners can't be loaded) to a separate issue.

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
DmitryBespalov: I suggest to move this to a separate issue / work item

Issue #2326, opened 2022-05-09
[iOS][addAsOwner]Not possible to go back to change data before execution from tx review


liliya-soroka:
**Bug description**
[iOS][addAsOwner]Not possible to go back to change data before execution from tx review
There is no way to go back to change data (required confirmation) on review screen 
**Steps To Reproduce**
1. safe is added and safe owner is added
2. generate a new key
3. complete backup
4. click X on success screen for backup
5. start add as an owner flow
6. click next on required confirmations page
Current result: Review screen is opened and there is no possibility to go back and change number of required confirmation 
**Expected Result**
The user should be able to go back and change the data
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/668aaba3-ba3b-47ae-a640-4a99f1f9432c)
**Device & App version**
 - Device/OS: iphone 11/15.4
 - App version: 3.11.0
 
**Environment**
staging
**Additional info**
relates to
https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-ios/2248

biafra23:
@liliya-soroka For me, there is a back button and it works. Can you provide the exact steps to reproduce?

 ![File (9).jpg](https://images.zenhubusercontent.com/5e97a1c48301b54c5cd938c1/79e3c8c7-5f89-4608-bf8b-251470c08132)

biafra23:
generate a new key-> pass backup -> click X button on success screen for backup -> start add as an owner flow => no header

liliya-soroka:
Verified
app version 3.12.0(1110)

liliya-soroka:
reopened in 3.13.0(1171)
1. safe is added
2. owner is added /not added ( please, fix the both cases)
3. create a key
4. go through backup process
5. start "request as an owner"
Current result: header is lost because of the success screen after backup
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/5ba92ab5-52a7-4875-be5a-c866932123e2)

liliya-soroka:
Reopened
still no header in 3.13.0 (1178)

liliya-soroka:
Verified
app version 3.13.0 (1180)

PR #2359, opened 2022-05-16
GH-2326 Show navigation bar on exit

biafra23:
Handles #2326

- unhide navbar on exit
- Prevent user from going back to success screen


DmitryBespalov (approved):
Let's see how this applies to success screens throughout the app. I think it will be OK. :shipit:

Issue #2364, opened 2022-05-17
Make the 'seed phrase' text entry switchable to 'secure text'


DmitryBespalov:
https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1551%3A16671

Right now the text is always shown as is. It would be good to have a 'secure' button that hides/unhides the text the same way as password fields.

liliya-soroka:
Verified
app version 3.12.0 (1112)

PR #2365, opened 2022-05-17
Gh 2364/secure text view

DmitryBespalov:
Handles #2364 

- Add the secure text for the PK/Seed inputs


biafra23:
It looks really good now with the multiline text input field. But there is a problem when deleting characters. The cursor jumps to the end. 

https://user-images.githubusercontent.com/246473/169037778-57a9b924-11a2-4f43-b315-29e2c6d44374.mov


I think the cursor position should be preserved while editing.

DmitryBespalov:
@biafra23 thank you for catching this!

DmitryBespalov:
should be fixed

biafra23:
There is one small thing left. When I insert a seed phrase via password manager, it is shown, even though the eye is crossed out. When I change a thing, it switches to the dots. See video: 

https://user-images.githubusercontent.com/246473/169066991-8cda9fd6-5d12-48e7-bb46-4e2712bb0f09.mov

DmitryBespalov:
Note to dummy self: find or create test cases for the part of the system that is changed and run through them after making the changes.

biafra23 (approved):

PR #2368, opened 2022-05-17
Gh 2326 - unhide too early

biafra23:
Handles #2326 

- do not unhide inside success screen
- add modal logic to UIFlow 
- enable and disable modal logic based on whether the presnter is nil

I considered having a ModalUIFlow base class. But we need most of the Flows in both variants.
I was looking for mixins or traits for Swift. 
Protocols can have default implementations, but those can only access to members. I couldn't make it work. The easiest was to decide in UIFlow by chekcing the presenter not to be nil.


DmitryBespalov (approved):
Looks good! 🚀

Issue #2360, opened 2022-05-16
[iOS] No ">" symbol on the safe setting for the safe with imported owner if another safe in "read-only" mode is added


liliya-soroka:
**Bug description**
No ">" symbol on the safe setting for the safe with imported owner if another safe in "read-only" mode is added
**Steps To Reproduce**
1. Add safe 1 and import safe owner
2. add safe 1 in read-only mode
3. check safe settings for the safe with owner
Current result: There is no > symbol near name and required confirmation but the option is available if click on the space without >

**Expected Result**
Make sure that > is displayed for the safes in "write" mode

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/4b9843e1-1cd1-4902-9faa-96b9b78055fe)
**Device & App version**
 - Device/OS: Iphone 11/15.4
 - App version: 3.12.0 (1099)
 
**Environment**
staging
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.12.0 (1113)

PR #2369, opened 2022-05-18
gh-2360 Added chevron

DmitryBespalov:
Handles #2360 

- Adds chevron image
- Disable safe modification actions when it is read-only


Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -243,7 +243,11 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
             return tableView.basicCell(name: name, indexPath: indexPath)
 
         case Section.RequiredConfirmations.confirmations(let name):
-            return tableView.basicCell(name: name, indexPath: indexPath, withDisclosure: !safe.isReadOnly, canSelect: !safe.isReadOnly)
+            return tableView.basicCell(name: name,
+                                       indexPath: indexPath,
+                                       withDisclosure: !safe.isReadOnly,
+                                       disclosureImage: UIImage(named: "arrow"),
```
biafra23: It seems the chevron is still missing when switching between readOnly and not read only safes. Seems to be a syncing issue.

Issue #2287, opened 2022-04-25
[ReplaceOwnerSettings] Adjust ReplaceOwnerController to connect all screens


DmitryBespalov:
# Overview
As a safe owner I want to create 'replace owner' transaction so I can change owner structure.

# What To Do
1. Adjust the existing ReplaceOwnerController to implement this use case
2. Reuse the screens from other flows (input address from 'Add Owner' flow, and existing 'review replace owner' and 'success' screens.

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/d42450d8-dadd-45b9-b6bd-af949215415d)

# Acceptance Criteria

- Able to create 'replace owner' transaction from safe settings

liliya-soroka:
Verified
app version 3.12.0 ( 1123)

PR #2373, opened 2022-05-18
Gh 2287/adjust replace owner controller to connect all screens

MouazAlzahabi:
Handles #2287


DmitryBespalov (approved):

biafra23:
I always get this error: 

image

What do I do to fix this? It is a safe on Rinkeby.

Issue #2392, opened 2022-05-23
[iOS] [add/replace owner]Show error if the safe themself is used as a new owner


liliya-soroka:
**Bug description**
Show error if the safe itself is used as a new owner
**Steps To Reproduce**
1. start adding an owner or replace owner flow on mobile
2. use the same safe as an owner 
Current result:
unexpected error is displayed 
**Expected Result**
From web ![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/98d35272-6a90-4cba-8ee0-c3d03c67a456)

**Device & App version**
 - Device/OS: iPhone 11/15.4
 - App version: 3.12.0 (1123)
 
**Environment**
staging
**Additional info**
(Optional)

liliya-soroka:
Reopened: no error is the safe themself is used as an owner
the error is displayed for the EOA

app version 3.13.0 (1150)

liliya-soroka:
Verified
app version 3.13.0 (1163)

Issue #2390, opened 2022-05-23
[change threshold]Disable "change threshold" for the 1 of 1 safe


liliya-soroka:
Disable "change threshold" for the 1 of 1 safe

Steps:
1. add safe 1 of 1 and import the owner
2. go to the safe settings 
3. go through the "change required confirmations" flow
Current result: the settings tx is created
Expected result: Disable "change threshold" for the 1 of 1 safe

liliya-soroka:
Reopened: the row color is changed on click.
Could we disable it as for the read-only safes?

app version 3.13.0 (1150)

liliya-soroka:
Verified
app version 3.13.0 (1163)

PR #2404, opened 2022-06-06
Disable change confirmations when owners count == 1

MouazAlzahabi:
Handles #2390 #2392

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
@@ -71,6 +71,7 @@ class EnterOwnerAddressViewController: UIViewController {
             self.addressField.setAddress(address, label: resolvedName)
             self.name = resolvedName
 
+            self.checkEqualToSafe()
             self.checkExisting()
         }
 
@@ -86,6 +87,11 @@ class EnterOwnerAddressViewController: UIViewController {
 
     private func clearErrors() {
         addressField.setError(nil)
+    }
+
+    private func checkEqualToSafe() {
+        guard safe.addressValue != address else { return }
+        handleError(error: "Cannot use Safe itself as owner.", text: address?.checksummed)
     }
 
     private func checkExisting() {
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
@@ -26,8 +26,6 @@ class EnterOwnerAddressViewController: UIViewController {
     private var address: Address?
     private var name: String?
 
-    private var addOwnerFlow: AddOwnerFlow!
-    
     override func viewDidLoad() {
         super.viewDidLoad()
 
--- a/Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
+++ b/Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
@@ -257,9 +257,10 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
             return tableView.basicCell(name: name, indexPath: indexPath)
 
         case Section.RequiredConfirmations.confirmations(let name):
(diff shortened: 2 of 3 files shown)


Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
```diff
@@ -71,6 +71,7 @@ class EnterOwnerAddressViewController: UIViewController {
             self.addressField.setAddress(address, label: resolvedName)
             self.name = resolvedName
 
+            self.checkEqualToSafe()
```
DmitryBespalov, line 72: Is this a generic view controller to enter address? Having checks for `equalToSafe` and `existing` seems like a wrong place

DmitryBespalov, line 72: Is this a generic view controller to enter address? Having checks for `equalToSafe` and `existing` seems like a wrong place

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -257,9 +257,10 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
             return tableView.basicCell(name: name, indexPath: indexPath)
 
         case Section.RequiredConfirmations.confirmations(let name):
+            let changeConfirmationDisabled = safe.isReadOnly || safe.ownersInfo?.count ?? 1 < 2 || safeOwners.count < 2
```
DmitryBespalov: Maybe we could move this logic to a controller or a flow so that if we want to start this flow from another place in the future we don't have to come back and dig this in the UI code

DmitryBespalov: Maybe we could move this logic to a controller or a flow so that if we want to start this flow from another place in the future we don't have to come back and dig this in the UI code

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -446,15 +447,17 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
             show(editSafeNameViewController, sender: self)
 
         case Section.RequiredConfirmations.confirmations(_):
-            guard let isReadOnly = safe?.isReadOnly else {
+            guard let safe = safe,
+                  !safe.isReadOnly,
+                  safe.ownersInfo?.count ?? 1 >= 2,
+                  safeOwners.count >= 2 else {
```
DmitryBespalov: same here, the logic check should be moved somewhere else. Maybe to the ChangeConfirmationsFlow initializer and make the initializer nillable

DmitryBespalov: same here, the logic check should be moved somewhere else. Maybe to the ChangeConfirmationsFlow initializer and make the initializer nillable

Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
MouazAlzahabi: this controller is used only in AddOwner/Replace owner flows, and we want to have this validation in those two places specifically

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
MouazAlzahabi: @DmitryBespalov  in that way I need to try create the flow in order to determine how to display UI, I will try another alternative

DmitryBespalov (approved):

Multisig/UI/Settings/OwnerKeyManagement/AddOwnerKey/EnterOwnerAddressViewController.swift
MouazAlzahabi: Were do you suggest to move it ?

DmitryBespalov: ok, after discussion, we decided to keep it here because it is fitting the current purpose.

Issue #2328, opened 2022-05-09
Onboarding screens on Mobile


sche:
Add onboarding screens on Mobile.

![image.png](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/1414e8cc-50a3-4516-925c-594c105b72f0)

**Links**
[Problem statement](https://github.com/safe-global/safe-pm/issues/56)
[Requirements](https://docs.google.com/document/d/1JnEgtub8_qKgmEl9N2GZ5hzdRQpL9ksXFtynYNJXO3M/edit#heading=h.p3arlzm44v2k)
[Design](https://www.figma.com/file/uywUJTuP658TcItB8XkGyC/Onboarding-screens)

# Diagrams
This is the logic of showing the first screen in the app depending on different conditions. The 'onboarding' screens is the new addition. This logic is implemented in the `SceneDelegate` class.

![Onboarding Activity Diagram.svg](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/36529e18-5f9b-4e81-a696-6edad4ad0f3a)

Below is the proposed classes for implementation of the onboarding screen:
![Onboarding Class Diagram.svg](https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/57ae064f-e657-4cbf-b859-3ff5b99c8f07)

# What to do
1. Create a container `OnboardingViewController` that uses or inherits from the `UIPageViewController`
2. Create a controller for a page, `OnboardingPageViewController`
3. Extract the "add safe", "create safe", "demo" buttons into a separate container view and use that in two places: in the page view controller, and in the existing `LoadSafeViewController`
4. Add tracking

<table>
  <tr>
   <td><strong>Tracking event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
   <td><strong>Description</strong>
   </td>
  </tr>
  <tr>
   <td>screen_onboarding_1
   </td>
   <td>NO
   </td>
   <td>User viewed the first onboarding screen
   </td>
  </tr>
  <tr>
   <td>screen_onboarding_2
   </td>
   <td>NO
   </td>
   <td>User viewed the second onboarding screen
   </td>
  </tr>
  <tr>
   <td>screen_onboarding_3
   </td>
   <td>NO
   </td>
   <td>User viewed the third onboarding screen
   </td>
  </tr>
  <tr>
   <td>user_skip_onboarding
   </td>
   <td>NO
   </td>
   <td>User skipped onboarding
   </td>
[...]

PR #2407, opened 2022-06-08
Gh 2328/onboarding screens on mobile

MouazAlzahabi:
Handles #2328


DmitryBespalov:
First of all, I love it! 🎉 

Nice swipe animation between pages! Nice "skip" animation!

However, we need to pay attention to details for these screens, because they are the first thing that new people see after installation. That's why it's important.

### Functional Issues
- When I tap on the page indicator to change the page, the indicator is changing but the page is not swiped  automatically. 
- Closing via "x" of the "create safe" modal is closing the onboarding. But when I do this for the "load safe", it doesn't close the onboarding. Inconsistent. I'm fine either way (whichever is easier to do), but let's make it consistent. 

### Usability Issues
@TanyaEfremova after arriving at the last screen, I had a feeling that I'm trapped and **have to** choose one of the options: create safe, load safe, or try demo. But I don't want any of those. Can there be a way to close the onboarding without choosing any of those options?

### Design Implementation Issues

@TanyaEfremova please double check if I've correctly indicated the issues.

On the left is the actual phone, on the right is the figma design.

splash

- [ ] 1. Seems like the logo should be bigger
- [ ] 2. The font size seems to be different
- [ ] 3. Size of the button, color of the title, and location of the button should be different
- [x] 4. Background color is used in the app, but a gradient is used in the design. The colors do not match.

screen-1

- [x] 1. Skip button should not be bold
- [x] 2. Font style of the title is different
- [x] 3. Font style of the text is different. 
- [x] 4. The major selling point is not highlighted in the text
- [x] 5. The inactive color of the page indicator is different

screen-2

- [x] Same as in the screen before

screen-3

- [x] Same issues that in the previous screens
- [x] The height of the buttons differ
- [x] Explore Demo button is bold but shouldn't. Plus, the text differs between "light" and "dark". "Completely new here?" vs "Completely new to Safe"

screen-1-dark

- [x] Background color in dark mode differs

DmitryBespalov:
- [x] Note: when I'm closing the app and coming back, the onboarding screens disappear

DmitryBespalov:
Reg. the changes to the "splash" and button position: this is all swift ui code and to change properly I'd need to re-write it in UIKit, which is not what to do at this point. Maybe when we completed all other tasks, I'd do it.

DmitryBespalov:
So, the comments to the first image in my comment above are won't fix for now.

TanyaEfremova:
> Can there be a way to close the onboarding without choosing any of those options?

@DmitryBespalov We could display a 'Close' icon on the last screen, where a user can start exploring the app and potentially load the keys in it. I wouldn't emphasis it more, as the primary actions are to either load existing Safe or create a new one, and should be more prominent.

See [Figma](https://www.figma.com/file/uywUJTuP658TcItB8XkGyC/Onboarding-screens?node-id=761%3A2554)


Other issues are correctly indicated! 

> The font size seems to be different

The main titles should be preferably as in the mocks, but we could also implement them in the later phases, as this is a new style.

Issue #2343, opened 2022-05-12
[RequestAddOwner] Intro screen after generating key


DmitryBespalov:
Intro screen to suggest request to add owner when the generated key created and there are no owner keys imported in the app for the selected safe.

https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=879%3A16132

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/12ad4aab-f9e4-479c-9aa0-d7737c9aecea" width=300 />

## What to do

1. Create the screen UI
2. Integrate it into the `GenerateKeyFlow`

## Acceptance Criteria
- If the safe is read-only, this screen is shown
- Otherwise, previously implemented "add key as owner" screen is shown.

MouazAlzahabi:
@DmitryBespalov @TanyaEfremova I don't see point of having close button + swipe to dismiss + Not this time button, I already removed the close button, please let me know if you don't agree with this

DmitryBespalov:
I do agree with this

DmitryBespalov:
@MouazAlzahabi btw, the swipe-to-dismiss would dismiss the whole add owner flow. So the "close" button would actually just skip this sub-flow

liliya-soroka:
Verified
app version 3.13.0 (1150)

PR #2408, opened 2022-06-09
Gh 2343/intro screen after generating key

MouazAlzahabi:
Handles #2343 

 screenshot
Simulator Screen Shot - iPhone 13 Pro - 2022-06-10 at 18 01 22

Design
Screen Shot 2022-06-10 at 6 03 49 PM


DmitryBespalov:
@MouazAlzahabi can you please attach a 'screenshot' vs 'design' image to the pull-request?

DmitryBespalov:
- The title's font style is not as it looks in figma
- The "not this time" button is too close to the bottom of the page

Multisig/UI/App/ViewControllerFactory.swift
```diff
@@ -84,9 +84,23 @@ enum ViewControllerFactory {
             UIBarButtonItem(barButtonSystemItem: .close, target: vc, action: #selector(CloseModal.closeModal))
     }
 
+    static func removeNavigationItem(_ vc: UIViewController) {
```
DmitryBespalov, line 87: 👍

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/CreateInviteOwnerIntroViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+        notThisTimeButton.setText("Not this time", .plain)
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        Tracker.trackEvent(.addAsOwnerIntro)
+    }
+
+    // Called when user swipes down the modal screen
+    func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
```
DmitryBespalov: can you check that this actually called? I think the swipe-to-dismiss is handled in another place (in navigation controller), but not sure.

When I tried to swipe down but then moved the finger up, the screen still swiped down, however I expected it to stay.

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/GenerateKeyFlow.swift
```diff
@@ -97,6 +109,14 @@ class GenerateKeyFlow: AddKeyFlow {
         push(flow: replaceOwnerFlow)
     }
 
+    func didInviteToAddKey(openKeyDetails: Bool = true) {
```
DmitryBespalov: I think this method is not needed, you could use the `didAddKeyAsOwner`. What is the reason that this duplicate method exists?

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/CreateInviteOwnerIntroViewController.swift
MouazAlzahabi: It turns out that it's general issue that we have unfortunately
Let's fix it separately

DmitryBespalov (approved):

Issue #2348, opened 2022-05-12
[RequestAddOwner] Create tracking events


DmitryBespalov:
### Share link

<table>
  <tr>
   <td><strong>Event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
   <td><strong>Description</strong>
   </td>
  </tr>
  <tr>
   <td>screen_add_key_as_owner_share_info
   </td>
   <td>chain_id (String): Chain id
   </td>
   <td>Info screen suggesting to share a universal link
   </td>
  </tr>
  <tr>
   <td>user_skip_share_link
   </td>
   <td>NO
   </td>
   <td>User skips sharing the universal link
   </td>
  </tr>
  <tr>
   <td>screen_add_owner_share_link
   </td>
   <td>chain_id (String): Chain id
   </td>
   <td>Screen with the universal link to share
   </td>
  </tr>
  <tr>
   <td>user_add_owner_share_link
   </td>
   <td>NO
   </td>
   <td>User opens share dialogue for universal link
   </td>
  </tr>
</table>

### Handle link

<table>
  <tr>
   <td><strong>Event</strong>
   </td>
   <td><strong>Parameters</strong>
   </td>
   <td><strong>Description</strong>
   </td>
  </tr>
  <tr>
   <td colspan="3" ><strong>User has safe with owner key</strong>
   </td>
  </tr>
  <tr>
   <td>screen_owner_from_link
   </td>
   <td>chain_id (String): Chain id
   </td>
   <td>Request to add owner from universal link when user has permission
   </td>
  </tr>
  <tr>
   <td>user_reject_owner_from_link
   </td>
   <td>NO
   </td>
   <td>User rejected adding owner from universal link
   </td>
  </tr>
  <tr>
   <td>screen_owner_from_link_choose_name
   </td>
   <td>chain_id (String): Chain id
   </td>
   <td>Choose new owner name
   </td>
  </tr>
  <tr>
   <td colspan="3" ><strong>User has read-only safe</strong>
   </td>
  </tr>
  <tr>
   <td>screen_owner_from_link_no_key
   </td>
   <td>chain_id (String): Chain id
   </td>
   <td>Request to add owner from universal link when user has no owner key
   </td>
  </tr>
  <tr>
   <td>user_owner_from_link_no_key_skip
   </td>
   <td>NO
   </td>
   <td>User without owner key skips adding owner from link
   </td>
  </tr>
  <tr>
[...]

PR #2425, opened 2022-06-15
GH-2348 - Add tracking events for sharing and handling links

biafra23:
Handles #2348

- Added all tracking events
- Events with the screen_ prefix automatically get the chainId as a parameter


DmitryBespalov:
@biafra23 it may happen that when the 'receive' screen is open, then the currently selected safe is on a different network than the safe to which the owner will be added. In that case, I think the chain_id should be the same as the chain of the safe in the link. Can this be specified so that it is not overridden with the automatic chain_id parameter?

biafra23:
@DmitryBespalov Can we call this `chain_id` taken from the link different? For example `permalink_chain_id`?

This way it is probably less confusing when analysing the data as the meaning of chain_id stays the same. And it'll be easier to implement.

DmitryBespalov:
Tracking doc https://docs.google.com/spreadsheets/d/1mj9iQIhpM-Pak7lQMb1dmxhQn1QdJoJyhbc730dZNzA/edit#gid=1949479240

DmitryBespalov (approved):

Multisig/UI/Settings/OwnerKeyManagement/RequestToAddOwner/ValidateRequestToAddOwnerViewController.swift
```diff
@@ -98,11 +98,13 @@ class ValidateRequestToAddOwnerViewController: UIViewController {
 
     func handleSafeNotAdded(info: SCGModels.SafeInfoExtended) {
         // safe must be added to the app to continue
+        Tracker.trackEvent(.screenOwnerFromLinkNoSafe)
         let noSafeVC = AddOwnerExceptionViewController.safeNotFound(
             address: parameters.safeAddress,
             chain: parameters.chain,
             onAdd: { [unowned self] in
-                // add safe
+                Tracker.trackEvent(.userOwnerFromLinkSafeNameAdded,
```
MouazAlzahabi, line 106: @DmitryBespalov @sche are you ok with using this event here or shall I rename it

Multisig/Cross-layer/Analytics/TrackingEvent.swift
```diff
@@ -310,6 +310,24 @@ enum TrackingEvent: String, Trackable {
...
+    // Handle link
+    case screenOwnerFromLink                        = "screen_owner_from_link"
+    case userRejectOwnerFromLink                    = "user_reject_owner_from_link"
+    case screenOwnerFromLinkChooseName              = "screen_owner_from_link_choose_name"
+    case screenOwnerFromLinkNoKey                   = "screen_owner_from_link_no_key"
+    case userOwnerFromLinkNoKeySkip                 = "user_owner_from_link_no_key_skip"
+    case userOwnerFromLinkNoKeyAddIt                = "user_owner_from_link_no_key_add_it"
+    case screenOwnerFromLinkNoSafe                  = "screen_owner_from_link_no_safe"
+    case userOwnerFromLinkNoSafeSkip                = "user_owner_from_link_no_safe_skip"
+    case screenOwnerFromLinkAddSafeName             = "screen_owner_from_link_add_safe_name" // not used
```
biafra23, line 328: Is it a bug, that this is not used? Or does it not make sense? Should we delete it?

MouazAlzahabi: This is because we are adding the safe with default name directly, so I assume we don't need that event now, but I kept it for future

Multisig/UI/Settings/OwnerKeyManagement/RequestToAddOwner/ValidateRequestToAddOwnerViewController.swift
MouazAlzahabi: @biafra23 See here

DmitryBespalov: ok

Multisig/Cross-layer/Analytics/TrackingEvent.swift
```diff
@@ -310,6 +310,24 @@ enum TrackingEvent: String, Trackable {
...
 
+    // Share link
+    case screenAddKeyAsOwnerShareInfo               = "screen_add_key_as_owner_share_info"
+    case userSkipShareLink                          = "user_skip_share_link"
+    case screenAddOwnerShareLink                    = "screen_add_owner_share_link"
+    case userAddOwnerShareLink                      = "user_add_owner_share_link"
+
+    // Handle link
+    case screenOwnerFromLink                        = "screen_owner_from_link"
+    case userRejectOwnerFromLink                    = "user_reject_owner_from_link"
```
DmitryBespalov, line 321: you know, maybe at some point we should do a refactor/rename and rename all these cases to look exactly as their values - that way it would be easier to search in the codebase

DmitryBespalov (approved):

Multisig/Cross-layer/Analytics/TrackingEvent.swift
biafra23: That would be so great! What direction?

`case userRejectOwnerFromLink                    = "userRejectOwnerFromLink"`
or
`case user_reject_owner_from_link                    = "user_reject_owner_from_link"`

Issue #2393, opened 2022-05-23
[RequstAddOwner] Integrate screens for link receiver


DmitryBespalov:
Connect all screens to integrate the flow when a person opens the link to add owner for a safe.

# What To Do

1. Modify "AddOwnerFlow":
  * If flow has new owner's name, then proceed to entering confirmations
  * otherwise start from "enter name" screen.

2. Modify "ReplaceOwnerFlow"
  * If flow doesn't have new owner's name, then show first "enter name"
  * else show the "select owner to replace" screen.

3. Integrate the "receive link" screen to have 3 outcomes / callbacks
  * When "add owner" selected, start AddOwnerFlow
  * When "replace owner" selected, start ReplaceOwnerFlow
  * When skipped, close the receive link screen.

4. The "receive link" screen must be opened in a modal from the MainTabBar.

5. Make sure that the "exceptional" cases handled:
  * When the owner already exists in that safe
  * When there's no such safe --> show the appropriate screen
  * When there's safe but it is read only --> show the exceptional screen
  * When the exceptional screens lead to added safe or added owner key, re-evaluate whether the owner can be added and then show again either exceptional screen or the "receive link intro" screen.

liliya-soroka:
Verified
app version 3.13.0 (1171)

Issue #2346, opened 2022-05-12
[RequestAddOwner] Add requested owner address to the address book


DmitryBespalov:
This screen gets a name from the user so that the newly added owner address would have a new entry in the address book when the flow finishes.

https://www.figma.com/file/2EgEwJHWRBx7KdGdPRezgn/Owner-Keys?node-id=1910%3A15633

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/0fee2d21-ab41-44d5-b771-062e569c8704" width=300 />

# What to do

1. Reuse existing screen.
2. This screen adds or edits the name of the owner address. 

# Acceptance criteria
1. If not exists in address book, then user needs to enter a name
2. If exists in address book, then pre-fill the name in the text field.

DmitryBespalov:
fyi Reusing estimate from the "Intro screen" as it seems to have similar complexity

liliya-soroka:
Verified
an owner is added to the addressbook if the adding flow was opened from the shared link only.
For "add owner" and "replace owner" flow the separate issue is created

app version 3.13.0 (1171)

PR #2428, opened 2022-06-16
Add flow to replace&add owner from settings

MouazAlzahabi:
Handles #2346 #2393 

- Add flows

Missing: 
- We still needs to adjust tracking

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReplaceOwnerFromInviteFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReplaceOwnerFromInviteFlow.swift
@@ -1,0 +1,49 @@
+//
+//  ReplaceOwnerFromInviteFlow.swift
+//  Multisig
+//
+//  Created by Moaaz on 6/16/22.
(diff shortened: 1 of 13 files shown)


Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReplaceOwnerFromInviteFlow.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  ReplaceOwnerFromInviteFlow.swift
+//  Multisig
+//
+//  Created by Moaaz on 6/16/22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+class ReplaceOwnerFromInviteFlow: ReplaceOwnerFlow {
```
DmitryBespalov, line 22: Can you please add a comment with the intended steps for this and other flows? You can see example in the other flows.

This way it is documented what screen sequence is implemented, and when you read the code you don't need to jump around the class hierarchy

DmitryBespalov (approved):
I'm super excited about this, especially to test the whole flow end-to-end where one person generates an owner, sends a link, and another person adds it to their safe

Issue #2446, opened 2022-06-22
[iOS][unexpected delegatecall warning] No warning on the actions in some case


liliya-soroka:
# **Bug description**
No warning on the actions in some case

# Expected Behavior
**Change To Requirements**
Any nested delegate call should be warned about because none of our interfaces create nested multi-send transactions. In general, it does not make sense to have nested multi-send transactions, so such cases should be avoided.

# **Steps To Reproduce**
1. check https://gnosis-safe.io/app/eth:0x583788f490a278DB2a8d6D7226264b4985f75678/transactions/multisig_0x583788f490a278DB2a8d6D7226264b4985f75678_0x6491b8096381f03e6e9d5bf771665d200bf78d475e9826ccc552ce9467127fdf tx
**Expected Result**
It has nexted actions with unexpected delegate calls but there is no warning on action
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/3d23ab73-1c31-4d0d-ab92-f8d1584bed51)
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/29db7680-0792-4a56-9750-0c563aa1a6ed)

**Device & App version**
 - Device/OS: iPhone 11/15.4
 - App version: 3.13.0 (1178)
 
**Environment**
prod
**Additional info**
(Optional)

biafra23:
@DmitryBespalov There is no warning on the multiSend entries because the to-address of the DELEGATE operation is "trusted".
I do not understand why there is a warning ot the  outer tx. Here is a link to the tx https://safe-client.staging.gnosisdev.com/v1/chains/1/transactions/multisig_0x583788f490a278DB2a8d6D7226264b4985f75678_0x6491b8096381f03e6e9d5bf771665d200bf78d475e9826ccc552ce9467127fdf

liliya-soroka:
Verified
app version 3.14.0 (1213)

PR #2454, opened 2022-06-23
GH-2446 (main) Fixes warning for DELEGATES and also show bigger icon in mult…

biafra23:
Handles #2446

- Fixes warning for DELEGATES and also show bigger icon in mult…
- Against `main` branch


Multisig/Data/Services/Utils/DelegateWarningCalculator.swift
```diff
@@ -76,7 +76,7 @@ class DelegateWarningCalculator {
     ///   - addressInfoIndex: name index of addresses in the decoded transaction data
     /// - Returns: true if transaction is untrusted, false otherwise.
     static func isUntrusted(multiSendTx: MultiSendTx, addressInfoIndex: AddressInfoIndex?) -> Bool {
-        let result = multiSendTx.operation == .delegate && addressInfoIndex?.values[multiSendTx.to]?.addressInfo == nil
+        let result = multiSendTx.operation == .delegate
```
DmitryBespalov, line 78: this would make any delegate operation untrusted. I think this is not correct

DmitryBespalov: the problem is that if the transaction has the flag untrusted, it should still be untrusted. 

I think the fix should be in a different code line

DmitryBespalov: it seems like the line 23 is not triggered?

biafra23: This is only called on MultiSendTxs that have the DELEGATE operation. According to Ivan, they should be warned about.

DmitryBespalov: I have checked the updates, and indeed my understanding was outdated when I was writing the ticket description.

Your change is correct ✅

DmitryBespalov: However, that means we need to remove all checks of the name in the calculator file and in the tests 🤦

DmitryBespalov: remove the `else` branch completely on line 33

DmitryBespalov:
up-to-date requirements:
* warn when receiving the `trustedDelegateCallTarget`
* warn on any nested "delegate" operation in multi-sends

biafra23:
I modified the code and added the TX from the bug report for unit tests.

Multisig/Data/Services/Utils/DelegateWarningCalculator.swift
```diff
@@ -10,15 +10,17 @@ class DelegateWarningCalculator {
     private static let TRUSTED = false
 
     // MultiSendTx is untrusted: operation = delegate && 'to' address is not 'known'
-        // 'resolving the name': look up in the address info index
+    // 'resolving the name': look up in the address info index
     // Safe Tx is untrusted: operation = delegate && 'to' is not 'known'
-        // 'resolving name': 'to' object's .name is nil
+    // 'resolving name': 'to' object's .name is nil
```
DmitryBespalov, line 14: the comment should change, too, now it is not synced with the code.

MultiSendTx is untrusted iff there's any "delegate" operation in its transactions.

Multisig/Data/Services/Utils/DelegateWarningCalculator.swift
```diff
@@ -57,9 +57,9 @@ class DelegateWarningCalculator {
                 // parameter is trusted when all sub-transactions are trusted.
                 return allTransactionsTrusted
 
-            // if it's nil (.none) or not decoded, then we say it's trusted
+                    // if it's nil (.none) or not decoded, then we say it's trusted
             case .none, .unknown:
-                return true
+                return UNTRUSTED
```
DmitryBespalov: hm, the comment says it's trusted, but the retunred value is "UNTRUSTED".

I think in this code block, it checks whether the tx is trusted, and not that it is untrusted.

DmitryBespalov (commented):
sorry but some more changes needed

Multisig/Data/Services/Utils/DelegateWarningCalculator.swift
biafra23: You're right. This should remain `true` and not be replaced by the names constant

DmitryBespalov (approved):

Issue #2462, opened 2022-07-02
[UniversalLinks] Share transaction details URL


MouazAlzahabi:
# Description

We want to give ability to share link for transaction as we have in our web interface.

[Figma](https://www.figma.com/file/ITvXrFnLET8cLrA43I3xJD/Share-a-TX-link?node-id=304%3A648)
<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/d8c3da9f-b2cb-41bd-afe7-0b2685dea6a6" width=300 />

# What To Do

1. Add Share button on top right corner of transaction details whenever there is transaction_id value in transaction details object
2. use the following schema to generate the share link 

```
https://gnosis-safe.io/app/{chain prefix}:{safe_address}/transactions/{client_gateway_transaction_id}
```

# Acceptance criteria

Get sharable link that work for both, mobile and web interface

DmitryBespalov:
@MouazAlzahabi fyi I adjusted the requirements here: added figma link and screenshot, and corrected the requirement description (was "top left" corner, which was incorrect)

liliya-soroka:
Verified
app version 3.14.0 (1197)

PR #2465, opened 2022-07-05
Add share url

MouazAlzahabi:
Handles #2462


DmitryBespalov (approved):

biafra23 (commented):
Where is the check, that we have a transaction_id? Or do we always have one, when in this screen?

MouazAlzahabi:
> Where is the check, that we have a transaction_id? Or do we always have one, when in this screen?

@biafra23 it turns out that transaction details option always has transaction id, it's not optional, in the past I remember we used to had it optional, I don't remember when this has been changed

biafra23 (approved):

Issue #2460, opened 2022-07-02
[DecodedData] Create PreviewTransaction request


MouazAlzahabi:
# Description

In order to fetch details of incoming wallet connect transaction details we have introduced a new CGW endpoint, we need to consume that endpoint. 

# What To Do

1. Create PreviewTransaction request to fetch WalletConnect incoming transaction details.
2. Extend SafeClientGatewayService to send the request and return response 

# Acceptance Criteria
* Add new test for SafeClientGatewayServiceIntegrationTests to test the new added method 

# Links

[Endpoint proposal](https://github.com/safe-global/safe-client-gateway/issues/840#issuecomment-1171257033)

liliya-soroka:
waiting for https://github.com/safe-global/safe-client-gateway/pull/924 to check

liliya-soroka:
Verified
app version 3.15.0 (1248)

PR #2466, opened 2022-07-05
Create request

MouazAlzahabi:
Handles #2460


Multisig/UI/WalletConnect/Server/Safes/Incoming Transaction/MockJSONHttpClient.swift
```diff
@@ -0,0 +1,66 @@
...
+            let txData = jsonData("MultiSendApproveMultihopBatchSwapExactIn")
+            do {
+                let output: T.ResponseType = try self.response(from: txData)
+                completion(.success(output))
+            } catch {
+                completion(.failure(error))
+            }
+        } else {
+            print("---> TransactionPreviewRequest | Unexpected request... -> error")
+            //completion(.failure(InvalidRequestError))
```
biafra23: What is the easiest way to creat a new error class and return it here. For example `InvalidRequestError`?

Multisig/Data/Services/Safe Client Gateway Service/TransactionPreviewRequest.swift
```diff
@@ -0,0 +1,78 @@
...
+        // Will always return the mocked response in completion
+        asyncExecuteMock(request: TransactionPreviewRequest(safe: transaction.safe!,
+                                                               sender: sender,
+                                                               transaction: transaction,
+                                                               chainId: chainId),
+                            completion: completion)
+    }
+
+    @discardableResult
+    func previewTransaction(transaction: Transaction, sender: AddressString, chainId: String) throws -> SCGModels.TransactionDetails? {
```
biafra23, line 71: This is not mocked. Do we need it?

Multisig/UI/WalletConnect/Server/Safes/Incoming Transaction/WCIncomingTransactionRequestViewController.swift
```diff
@@ -215,6 +215,23 @@ class WCIncomingTransactionRequestViewController: UIViewController {
 
         tableView.estimatedRowHeight = BasicCell.rowHeight
 
+        // TODO Is this the right place for this call?
+         App.shared.clientGatewayService.asyncPreviewTransaction(
```
biafra23: Is this the right place for this call?

Multisig/UI/WalletConnect/Server/Safes/Incoming Transaction/MockJSONHttpClient.swift
DmitryBespalov: for debug issues, just use a String, I think it is defined as an error protocol implementation somewhere, so you can use it

Multisig/UI/WalletConnect/Server/Safes/Incoming Transaction/WCIncomingTransactionRequestViewController.swift
DmitryBespalov: yes I think so

DmitryBespalov (approved):

Issue #2410, opened 2022-06-14
[AssetContractLink] Link to Asset Contract 


DmitryBespalov:
# Description

As a user,
I want to see the token contract of any asset I have,
So that I'm confident about my token operations.

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/6ccba02d-0d7a-4425-90ef-ea3466610aed" width=300 />

# What To Do

1. Add a link to the contract address for an asset when tapping on the asset row

# Acceptance Criteria

* When tapping on a row, the app opens the webview of the token contract in block explorer
* If token name is too big to fit, the token name is ellipsized at the trailing end, the 'link' icon still shown
* Remove previous functionality of starting "send assets" when tapping on an asset row

# Resources

* [Figma](https://www.figma.com/file/wwa8TTJwjV007sTuHI4i8q/Tokens-management?node-id=26%3A5535)
* [Problem Statement](https://github.com/safe-global/safe-pm/issues/70)

liliya-soroka:
https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-ios/2474

DmitryBespalov:
commented on the issue itself: requirements changed, this is expected behavior

liliya-soroka:
Verified
app version 3.14.0(1197)

PR #2467, opened 2022-07-05
Add browse token address url

MouazAlzahabi:
Handles #2410 
Simulator
simulator_screenshot_9C5AFE1C-F2AB-4447-8F3E-C4048FFE91CA
Figma
Screen Shot 2022-07-05 at 3 50 39 PM

### DIFF
--- a/Multisig/Assets.xcassets/icon-external-link.imageset/Contents.json
+++ b/Multisig/Assets.xcassets/icon-external-link.imageset/Contents.json
@@ -18,5 +18,8 @@
   "info" : {
     "author" : "xcode",
     "version" : 1
+  },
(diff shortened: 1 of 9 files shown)


biafra23 (approved):
I find the name `browseIcon ` very generic. Maybe something like `browseTokenInfo` or `linkToTokenInfo` would be better?

MouazAlzahabi:
> I find the name `browseIcon ` very generic. Maybe something like `browseTokenInfo` or `linkToTokenInfo` would be better?

Well it's actually just an icon for browse, even the used icon is a generic icon that's used all over the app, I think it's fine to keep it

Issue #2436, opened 2022-06-20
[iOS] The name of the new owner is not used on Review screen and not saved to the addressbook for "Add owner" and "replace owner"


liliya-soroka:
**Bug description**
The name of the new owner is not used on Review screen and not saved to the addressbook for "Add owner" and "replace owner"

Precondition:
1. the safe with one owner( key) is added in the app
**Steps To Reproduce**
1. go to the safe settings and start "Add owner" or "replace owner" flows
2. add a new owner address( make sure that no such owner in addressbook or in the owners list)
3. enter name for a new owner
4. go to the review screen
Current result: Just entered on the previous step name is not displayed on review screen

5. finish "Add owner" or "replace owner" flow ( create tx but not execute)
6. go to the addressbook 
Current result: No new entity with new owner name
**Expected Result**
according to https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-ios/2346, I would expect that as soon as a name for the new owner is entered the new entity is created in the addressbook 
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/15.4
 - App version: 3.13.0( 1171)
 
**Environment**
staging 
**Additional info**
(Optional)

JagoFigueroa:
Morning! I think there is an issue on the replace flow (please, let me know if the behaviour that I will share now is the expected one or if it is better to approach this in another ticket).

When replacing an owner, I am asked a name for the "replacer" as expected but I don't see it on the review screen:

Case 1, nothing on my addressbook:
![IMG_0632.PNG](https://images.zenhubusercontent.com/625e824cecd9c30f320dd106/436f22c6-b721-4bd5-9b23-1df8ab50fa3f)

Case 2, address to be replaced is on my addressbook: 
![IMG_0633.PNG](https://images.zenhubusercontent.com/625e824cecd9c30f320dd106/e05294db-c09e-4064-b828-091664c572a7)

Should we be seeing the "replacer" name in these cases? Cheers!

JagoFigueroa:
Reinstalling the build seems to have fixed the issue above, giving it a second go of testing and will close the ticket if all is bueno :)

PR #2470, opened 2022-07-07
GH-2436 - Add new owner to address book earlier

biafra23:
Handles #2436

- Add Address book entry when showing the success screen
- Pass new name to ViewControllers to show it in review screen


MouazAlzahabi:
> Handles #2436
> 
> Changes proposed in this pull request:
> 
> * Add Address book entry when name is confirmed

@biafra23 , I don't think that we should go with this solution, after we confirm the name we still have the possibility to cancel the whole flow, in that situation I think we should avoid any change to the app state, and saving a new entity to the address book is not comply with this

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/AddOwnerFlowFromSettings.swift
```diff
@@ -64,6 +64,7 @@ class AddOwnerFlowFromSettings: AddOwnerFlow {
             trackingEvent: .addOwnerSpecifyName
         ) { [unowned self] name in
             newAddressName = name
+            AddressBookEntry.addOrUpdate(newOwner!.checksummed, chain: safe.chain!, name: name)
```
DmitryBespalov: I'm a bit unsure if this is the right fix for the issue. Can we pass on the name for that address to further screens to display it in Review or if it's added here, then remove it when the flow is cancelled and the name was added?

This is my view of this change:

pros:
 - if the flow is unfinished (cancelled) and then repeated, the name will be shown in the UI already
 - the name is shown on review screen automatically

cons:
  - if the flow is cancelled, then there's new entry in address book which was not used, maybe it would be a mistake to have it then?

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/AddOwnerFlowFromSettings.swift
```diff
@@ -76,6 +77,7 @@ class AddOwnerFlowFromSettings: AddOwnerFlow {
 
     override func success() {
         assert(transaction != nil)
+        // TODO remove this here?
         AddressBookEntry.addOrUpdate(newOwner!.checksummed, chain: safe.chain!, name: newAddressName!)
```
DmitryBespalov, line 96: I think TODO is correct, if we add it earlier, then updating name here is not needed

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReplaceOwnerFromSettingsFlow.swift
```diff
@@ -64,6 +64,7 @@ class ReplaceOwnerFromSettingsFlow: ReplaceOwnerFlow {
             trackingEvent: .replaceOwnerNewOwnerName
         ) { [unowned self] name in
             newAddressName = name
+            AddressBookEntry.addOrUpdate(newOwner!.checksummed, chain: safe.chain!, name: name)
```
DmitryBespalov: the same concerns as above

biafra23:
> @biafra23 , I don't think that we should go with this solution, after we confirm the name we still have the possibility to cancel the whole flow, in that situation I think we should avoid any change to the app state, and saving a new entity to the address book is not comply with this

Isn't that exactly what @liliya-soroka wanted to happen? That the name is saved even if the flow is not completed. 
@liliya-soroka Please clarify!

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/AddOwnerFlowFromSettings.swift
biafra23: In the latest implementation, this is the right place. So I removed the comment.

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/AddOwnerFlowFromSettings.swift
biafra23: The latest implementation saves the address book entry in the success screen.

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReplaceOwnerFromSettingsFlow.swift
biafra23: It's gone

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/AddOwnerFlowFromSettings.swift
```diff
@@ -74,20 +74,27 @@ class AddOwnerFlowFromSettings: AddOwnerFlow {
...
+            transaction = txDetails
+            success()
+        }
+        show(reviewVC)
+    }
+
     override func success() {
         assert(transaction != nil)
+        // TODO remove this here?
         AddressBookEntry.addOrUpdate(newOwner!.checksummed, chain: safe.chain!, name: newAddressName!)
```
DmitryBespalov, line 96: is this intended? the other comment was saying you removed the comment...

biafra23: ups

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReviewReplaceOwnerTxViewController.swift
```diff
@@ -88,10 +91,12 @@ class ReviewReplaceOwnerTxViewController: ReviewSafeTransactionViewController {
...
                                                 chainId: safe.chain!.id!)
 
+            let newName = newAddressName ?? newOwnerName
+
             cell.set(
-                newOwner: AddressInfo(address: owner, name: newOwnerName),
+                newOwner: AddressInfo(address: owner, name: newName),
                 oldOwner: AddressInfo(address: ownerToBeReplaced, name: oldOwnerName),
-                prefix: safe.chain?.shortName
+                    prefix: safe.chain?.shortName
```
MouazAlzahabi: Please fix line alignment

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReviewAddOwnerTxViewController.swift
```diff
@@ -74,8 +76,8 @@ class ReviewAddOwnerTxViewController: ReviewSafeTransactionViewController {
         let (name, _) = NamingPolicy.name(for: owner,
                                                     info: nil,
                                                     chainId: safe.chain!.id!)
-
-        cell.set(owner: AddressInfo(address: owner, name: name), action: .addingOwner, prefix: safe.chain?.shortName)
+        let newName = newAddressName ?? name
```
MouazAlzahabi, line 79: I think this should be the opposite
`let newName = name ?? newAddressName `

This way owner keys table will have more priority than new name which might be address book somehow 
@DmitryBespalov is this aligned with our requiremet ?

MouazAlzahabi (changes requested):

Multisig/UI/Settings/OwnerKeyManagement/AddKeyAsNewOwner/ReviewAddOwnerTxViewController.swift
biafra23: `newAddressName` is the name just entered by the user. IMHO that should have precedence of any `name` form the database. 

Only if `newAddressName` is nill, then the fallback `name` is used.

We could make this more efficient by not calling `NamingPolicy.name()` if newAddressName is not nil.

biafra23:
```
        if let newName = newAddressName {
            cell.set(owner: AddressInfo(address: owner, name: newName), action: .addingOwner, prefix: safe.chain?.shortName)
        } else {
            let (name, _) = NamingPolicy.name(for: owner, info: nil, chainId: safe.chain!.id!)
            cell.set(owner: AddressInfo(address: owner, name: name), action: .addingOwner, prefix: safe.chain?.shortName)
        }
```

Like this. Is it possible to have only one call to `cell.set()`? And without a var declaration before the block.

MouazAlzahabi (approved):

Issue #2463, opened 2022-07-02
[UniversalLinks] Support deep linking for missing routes


MouazAlzahabi:
# Description
We want to support deep linking for the following screens

- [x] Assets balances
- [x] Transactions queue
- [ ] Transactions history
- [ ] Transaction details

# What To Do

1. Add new navigation methods to NavigationRoute 
2. Adjust method canNavigate in MainTabBarViewController.swift line 380 to handle the new added cases 

## Transaction details

    https://gnosis-safe.io/app/{chain prefix}:{safe_address}/transactions/{client_gateway_transaction_id}

## Asset Balances

    https://gnosis-safe.io/app/{chain prefix}:{safe_address}/balances

## Transaction Queue
    https://gnosis-safe.io/app/{chain prefix}:{safe_address}/transactions/queue

## Transaction History
    https://gnosis-safe.io/app/{chain prefix}:{safe_address}/transactions/history

## NOTE
For the staging environment, the app must support using the `safe-team.staging.gnosisdev.com` host instead of the `gnosis-safe.io` host.
 
# Acceptance Criteria

- We can open transaction mentioned screens by using shared link from web interface
- If the safe is added to the app, and corresponding link is opened, the app opens at the intended screen
- If the safe is not added to the app, then the link is ignored (nothing happens)

# Links
* [Requirements](https://docs.google.com/document/d/1aoW-HZDeQCRpMVMaHOUZQZyBNMWcqCo51IEkq3zEBlE/edit#heading=h.t0of4xmer2ue)

DmitryBespalov:
@MouazAlzahabi I have updated the requirements with the link URL formats

liliya-soroka:
Verified
the issues for staging app and for incoming tx are reported separately 

app version 3.14.0 (1204)

PR #2471, opened 2022-07-07
Gh 2463/support deep linking for missing routes

MouazAlzahabi:
Handles #2463


Multisig/UI/App/NavigationRouter.swift
```diff
@@ -39,6 +39,36 @@ class DefaultNavigationRouter: NavigationRouter {
     func navigate(to route: NavigationRoute) {
         sceneDelegate?.navigate(to: route)
     }
+
+    private let pattern = "^https://gnosis-safe.io/app/([-a-zA-Z0-9]{1,20}):(0x[a-fA-F0-9]{40})/([-a-zA-Z0-9]{1,20})(/[-a-zA-Z0-9_]{1,118})?"
```
DmitryBespalov, line 43: just a note: this will break when the tx id format will change (the last term)

DmitryBespalov (approved):

Multisig/UI/App/NavigationRouter.swift
MouazAlzahabi: Will it changed? 
any suggestion to get ride of this ?

DmitryBespalov: thinking about this... perhaps not.

good to go for  now.

elgatovital (approved):

Issue #2441, opened 2022-06-21
[iOS] [addOwnerFromLink]Flow is not started if the pop-up window is opened in the app


liliya-soroka:
**Bug description**
Flow is not started if the pop-up window is opened in the app
**Steps To Reproduce**
1. open app and open Switch safe
2. try to open the link https://gnosis-safe.io/app/rin:0x1230B3d59858296A31053C1b8562Ecf89A2f888b/addOwner?address=0xdBb8af3026040562e2ca9175C9E8856f614a154a to start "add owner" flow

Current result: Nothing is happened after click on the link as another pop-up window is opened in the app
**Expected Result**
Make sure that the flow will start even if another pop-up window was opened
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/15.4
 - App version: 3.13.0 (1171)
 
**Environment**
staging
**Additional info**
(Optional)

MouazAlzahabi:
@liliya-soroka @DmitryBespalov  This is a general issue we have over the app when we try to present modal view over another, try same behavior with WalletConnect incoming tx

MouazAlzahabi:
So the solution to present last presented view and dismiss the current presented one

JagoFigueroa:
tested and verified in v3.14.0 (1213) thanks!

PR #2484, opened 2022-07-14
gh-2441 Dismiss popup screen if any before showing flow

elgatovital:
Handles #2441

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
@@ -43,12 +43,10 @@ class UIFlow {
 
         presenter = vc
 
-        if presenter.presentedViewController != nil {
-            presenter.presentedViewController!.dismiss(animated: true)
-            presenter.present(nav, animated: true)
-        } else {
-            presenter.present(nav, animated: true)
+        if let presentedViewController = presenter.presentedViewController {
+            presentedViewController.dismiss(animated: true)
         }
+        presenter.present(nav, animated: true)
     }
 
     func push(from vc: UIViewController) {
--- a/Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
@@ -44,7 +44,8 @@ class UIFlow {
         presenter = vc
 
         if presenter.presentedViewController != nil {
-            presenter.presentedViewController!.present(nav, animated: true)
+            presenter.presentedViewController!.dismiss(animated: true)
+            presenter.present(nav, animated: true)
         } else {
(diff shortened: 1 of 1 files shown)


Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
```diff
@@ -42,7 +42,12 @@ class UIFlow {
         ViewControllerFactory.addCloseButton(rootVC)
 
         presenter = vc
-        presenter.present(nav, animated: true)
+
+        if presenter.presentedViewController != nil {
```
DmitryBespalov: sorry but this is wrong.. it must show as a modal, and not `show`. If the `show` needed, then there's another method for it, `push`.

Instead of showing as a `show` perhaps it's better presnting on presented viewcontroller. Or even better, get (recursively) the most-presented viewcontroller and present from that.

DmitryBespalov (changes requested):

Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
DmitryBespalov: BUT. Moreover, this is not how we agreed to do it before, can you explain why this solution is not used?

https://github.com/safe-global/safe-ios/issues/2441#issuecomment-1173786524

elgatovital: presenting on presented view controller works

elgatovital: Adjusted solution to dismiss currently presented controller before presenting flow

Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
```diff
@@ -43,12 +43,10 @@ class UIFlow {
 
         presenter = vc
 
-        if presenter.presentedViewController != nil {
-            presenter.presentedViewController!.dismiss(animated: true)
-            presenter.present(nav, animated: true)
-        } else {
-            presenter.present(nav, animated: true)
+        if let presentedViewController = presenter.presentedViewController {
+            presentedViewController.dismiss(animated: true)
```
DmitryBespalov: sorry, but this will not work, because the dismiss animation will prevent the presentation. You need to dismiss with completion block and then present from the completion block.

Also, beware of retain cycles in completion block (unowned/self needed)

DmitryBespalov (approved):

Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
elgatovital: it worked with presenting right after dismissal, but now it's bulletproof

Issue #2295, opened 2022-04-25
[iOS] [backup]Connect key flow is closed without enabling push notification dialog if the "set up passcode" window is swiped down


liliya-soroka:
**Bug description**
Connect key flow is closed without enabling push notification dialogue if the "set up passcode" window is swiped down
The same for ledger nano X connection
**Steps To Reproduce**
1. add a new key
2. select connect via WC
3. select mobile wallet app to connect 
4. enter key name and click add
Current result: set up passcode window appears

5. swipe set up passcode window down
Current result: add key dialogue is closed with added wc key
**Expected Result**
Enable push notification window should be displayed 

**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/15.4
 - App version: 3.10.0 (1047)
 
**Environment**
prod
**Additional info**
(Optional)

Relates to
https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/gnosis/safe-ios/2202

DmitryBespalov:
Alternatives:
- instead of closing the window, just go to the next step in the flow
- disable "swipe down"
- do nothing, because swiping down closes the whole flow, so it's a valid action


For me, disabling "swipe down" is a viable solution

liliya-soroka:
Verified
app version 3.15.0 (1248)

PR #2511, opened 2022-07-20
gh-2295 Prevent premature connect key flow closure when swiping down passcode creation popup

elgatovital:
Handles #2295

- Disable swipe down gesture


Multisig/UI/Settings/OwnerKeyManagement/Backup/CancellableNavigationController.swift
```diff
@@ -16,6 +16,8 @@ class CancellableNavigationController: UINavigationController, UIAdaptivePresent
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         presentationController?.delegate = self
+        // disable swipe down gesture
+        isModalInPresentation = true
```
DmitryBespalov: can you please make this configurable and make it only for the "generate key" flow or a subflow (even better)?

the reason is that, if I'm not mistaken, then this code is used for all modal flows, and it would be too much to disallow any modal to be closed with swipe down.

DmitryBespalov (approved):

Issue #2482, opened 2022-07-13
☔️ Test Automation Umbrella


DmitryBespalov:
TBD

This is an umbrella issue for the test automation project on iOS

PR #2513, opened 2022-07-22
gh-2482 Backup flow test

elgatovital:
Handles #2482


DmitryBespalov:
My advice:
- Use the `throws` in the test function declaration so that you don't have to do `try!` or `do {} catch {}`
- Extract repeating test code to a utility function
- Create methods in the view controller so that test could manipulate it from the interface and without peeking inside the view controller structure, i.e. encapsulate

DmitryBespalov (approved):

Issue #2515, opened 2022-07-25
[Claiming app Banner] Display banner on Ethereum and Rinkeby only


liliya-soroka:
# Description
Display banner on Ethereum and Rinkeby only
As a user,
I want to see the banner only on the networks with available claiming app ,
so that I open claiming app from the banner.

# What To Do

1. Banner must only show for the Ethereum  and Rinkeby network Safes because the airdrop is only on Ethereum and Rinkeby ( for testing purposes).

# Acceptance Criteria

* The banner is displayed only on Ethereum and Rinkeby networks 

# Resources

liliya-soroka:
Verified
app version 3.15.0 (1241)

Issue #2499, opened 2022-07-18
[SAFE] Delegate To Custom Address


DmitryBespalov:
# Delegate To Custom Address

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/426f1835-fefe-4977-a052-6f1546b8e5e7" width=300 />

https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=341%3A6133

## Acceptance criteria

- Uses standard component to enter address from clipboard, qr code, address book, or ens name
- This address will be used for setting a delegate in the delegate registry for this safe.

liliya-soroka:
Verified
app version 3.15.0 (1241)

PR #2516, opened 2022-07-26
gh-2499 Enter custom address

elgatovital:
Handles #2499
Handles #2515
Handles #2518

- Show claim banner only for mainnet and rinkeby safes
- Screen for entering custom delegate address
- Claim not available screen

<table>
<tr>
<td>
<image src="https://user-images.githubusercontent.com/17750984/180926920-40f28a02-d9d2-4a83-a625-127e875a3c71.png" width=400 />
</td>
<td>
<image src="https://user-images.githubusercontent.com/17750984/180926995-cdf7cb7c-671b-4359-9006-92b566b36812.png" width=400 />
</td>
<td>
<image src="https://user-images.githubusercontent.com/17750984/181192179-f50358f7-50e0-4cb7-8df5-7b8ab0317dfd.png" width=400 />
</td>
<td>
<image src="https://user-images.githubusercontent.com/17750984/181192826-a6c5a77a-1c7f-46f0-b69d-ef8f6853ed75.png" width=400 />
</td>
</tr>
</table>

### DIFF
--- a/Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
+++ b/Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
@@ -26,7 +26,7 @@ class EnterCustomAddressViewController: UIViewController {
     private var address: Address? { addressField?.address }
 
     var mainnet: Bool = true
-    var onContinue: (() -> ())?
+    var onContinue: ((_ address: Address) -> ())?
 
 
     override func viewDidLoad() {
@@ -162,7 +162,7 @@ class EnterCustomAddressViewController: UIViewController {
     }
 
     @IBAction func didTapContinueButton(_ sender: Any) {
-        onContinue?()
+        onContinue?(address!)
     }
 }
 
--- a/Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
+++ b/Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
@@ -1,0 +1,188 @@
+//
+//  EnterCustomAddressViewController.swift
+//  Multisig
+//
+//  Created by Vitaly on 24.07.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+import Web3
+
+class EnterCustomAddressViewController: UIViewController {
+
+    @IBOutlet private weak var continueButton: UIButton!
+    @IBOutlet private weak var addressField: AddressField!
+
+    private var stepLabel: UILabel!
+    private var stepNumber: Int = 2
+    private var maxSteps: Int = 3
+
+    private var debounceTimer: Timer!
+    private let debounceDuration: TimeInterval = 0.250
+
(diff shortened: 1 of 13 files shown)


DmitryBespalov:
can the address become 2 lines? The address is not fully shown, but when adding safe, we have the address fully shown in 2 lines

Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+            chain = SCGModels.Chain.mainnetChain()
+        } else {
+            chain = SCGModels.Chain.rinkebyChain()
+        }
+
+        trackingParameters = { ["chain_id" : chain.chainId.description] }()
+    }
+
+    private func didTapAddressField() {
+        let alertVC = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
```
DmitryBespalov, line 61: Can you explain more why this is re-implemented? I think we already have a view controller that gets address from the user

Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+//  Multisig
+//
+//  Created by Vitaly on 24.07.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+import Web3
+
+class EnterCustomAddressViewController: UIViewController {
```
DmitryBespalov, line 12: In general, I think we already have such a screen that asks for an address from the user. We should have used that one

Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
elgatovital: what view countroller do you mean? Enter safe address screen and address book screen use same approach as here. This could be refactored and extracted to a helper class, but I would do this as a separate task.

Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
elgatovital: We do have different screens for address input but they differ in number of small details. I would not generalize this screen and keep it as simple as possible. This will allow us to be more flexible in the future, should any additional requirements / changes come that are related to this screen / flow.

elgatovital:
> can the address become 2 lines? The address is not fully shown, but when adding safe, we have the address fully shown in 2 lines

yes. I will adjust the layout.

Multisig/UI/ClaimToken/EnterCustomAddressViewController.swift
elgatovital: ticket for refactoring task: https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-ios/2519

DmitryBespalov (approved):

Issue #2548, opened 2022-08-15
[SAFE] Adjust Delegate List screen


DmitryBespalov:
- [x]  @TanyaEfremova : review images from delegates
    - [x]  @DmitryBespalov : build an HTML to review the images
- [x]  Fix: when click on “48 delegates”, it highlights the row with gray background
- [x] If image of a guardian not found, then show a placeholder image (NOT identicon/blockie)
- [x]  Not Found screen
    - [x] icon for delegate not found changed
    - [x]  Change the empty state screen style (across the app)
        - [x] Text style changed
- [x] When keyboard appears, the list is not scrolled to the end

https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=336%3A7418

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/0f09924c-aaf4-4e36-95a8-8ae257e5d193" width=300 />

liliya-soroka:
Verified
but the issue 2561 is reopened with details

app version 3.16.0 (1277)

PR #2563, opened 2022-08-18
Fix issues

MouazAlzahabi:
Handles #2548


Multisig/UI/ClaimToken/Select Guardian/GuardianTableViewCell.swift
```diff
@@ -31,11 +31,13 @@ class GuardianTableViewCell: UITableViewCell {
...
             label: guardian.name,
             imageUri: guardian.imageURL,
+            placeholderImage: "ico-no-delegate-placeholder",
             badgeName: nil,
             prefix: nil)
 
         addressInfoView.copyEnabled = false
         
         descriptionLabel.text = guardian.reason
+        layoutIfNeeded()
```
elgatovital, line 41: why would it be needed? What makes it necessary to call it explicitly?

elgatovital (approved):

Multisig/UI/ClaimToken/Select Guardian/GuardianTableViewCell.swift
MouazAlzahabi: Here we are in reusable view, when you replace it with longer text the size of cell is not adjusted without explicit call

Issue #2557, opened 2022-08-15
[SAFE] get data for the claim screen


DmitryBespalov:
- [x]  claimed amount

Single account vesting data format is specified in the [Notion Doc](https://www.notion.so/gnosis-safe/Safe-Claiming-App-bc121fbfe5a249368b6d0779a7912c84#8e844508f00c4a72a4f13c38e001b85c). 

For convenience, copy is here:

```
Filename: <ACCOUNT_ADDRESS (checksummed)>.json

[
	  {
	      "tag": "User Vesting",
	      "account": "ACCOUNT_ADDRESS (checksummed)",
	      "chainId": 4,
	      "contract": "VESTING_CONTRACT_ADDRESS (checksummed)",
	      "vestingId": "0x7ebb10e83e90cb4a8ad1ce330d59b2e71e1efd5f3b539da6db76b4de20021380",
	      "durationWeeks": 416,
	      "startDate": 1531562400,
	      "amount": "1000000000000330000000",
	      "curve": 0,
	      "proof": ["0x...", "0x...."]
	  },
	  ...
]
```

DmitryBespalov:
@elgatovital will you add the "tag" to the allocations (single file jsons) or should I not count on it?

elgatovital:
@DmitryBespalov you can count on it. tag=["ecosystem" | "user"]

elgatovital:
@DmitryBespalov regenerated vesting files and added tag field

DmitryBespalov:
Thank you!

liliya-soroka:
Verified
app version 3.17.0 (1312)

PR #2574, opened 2022-08-22
gh-2557 Pull data from contracts, part 1

DmitryBespalov:
Handles #2557

- Refactor eth_call to a separate RpcClient class
- Get SafeToken.paused
- Get Airdrop.Vesting by vesting hash (vesting id)
- Get DelegateRegistry.delegate by delegator address


biafra23:
I would prefer not to have spaces in file or directory names. But apparently that ship has sailed. But do we need to have a trailing space in a directory name like: [Multisig/Data/Services/Safe Claiming Service /GuardiansRequest.swift](https://github.com/safe-global/safe-ios/pull/2574/files#diff-2dd592adf754cc4891e0f96711e2c92db5f199a3f5e83f9ebbfda745ae86c776)

Multisig/Logic/SafeTransactionController.swift
```diff
@@ -123,56 +123,17 @@ class SafeTransactionController {
...
-                let addresses = returns._arg0.elements.compactMap(Address.init)
-                dispatchOnMainThread(completion(.success(addresses)))
-            } catch {
-                dispatchOnMainThread(completion(.failure(error)))
+            input: GnosisSafe_v1_3_0.getOwners()
+        ) { contractResult in
+            let result = contractResult.map { returns in
+                returns._arg0.elements.compactMap(Address.init)
             }
+            dispatchOnMainThread(completion(result))
```
MouazAlzahabi: We have the same line in RPC client send method

MouazAlzahabi (approved):

DmitryBespalov:
>I would prefer not to have spaces in file or directory names. But apparently that ship has sailed. But do we need to have a trailing space in a directory name like: [Multisig/Data/Services/Safe Claiming Service /GuardiansRequest.swift](https://github.com/safe-global/safe-ios/pull/2574/files#diff-2dd592adf754cc4891e0f96711e2c92db5f199a3f5e83f9ebbfda745ae86c776)

Oh, I didn't notice that it was like that. I was using the existing folder structure. Will change in the next PR

Multisig/Logic/SafeTransactionController.swift
DmitryBespalov: yeah, true that it's enough if it's just in one place already

PR #2580, opened 2022-08-23
gh-2557 Added calls to get more data

DmitryBespalov:
Handles #2557

- New api call
- Refactor the namesof methods to be more swifty

### DIFF
--- a/Multisig/Data/Services/Safe Claiming Service/AllocationsRequest.swift
+++ b/Multisig/Data/Services/Safe Claiming Service/AllocationsRequest.swift
@@ -1,0 +1,31 @@
+//
+//  AllocationsRequest.swift
+//  Multisig
+//
+//  Created by Dmitry Bespalov on 23.08.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+struct AllocationsRequest: JSONRequest {
+    var account: Address
+
+    var httpMethod: String { "GET" }
+
+    var urlPath: String {
+        "/claiming-app-data/resources/data/allocations/\(account.hexadecimal).json"
+    }
+
+    typealias ResponseType = [Allocation]
+
+    func encode(to encoder: Encoder) throws {
+        // empty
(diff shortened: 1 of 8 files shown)


Multisig/Data/Services/Safe Claiming Service/AllocationsRequest.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    var httpMethod: String { "GET" }
+
+    var urlPath: String {
+        "/claiming-app-data/resources/data/allocations/\(account.hexadecimal).json"
+    }
+
+    typealias ResponseType = [Allocation]
+
+    func encode(to encoder: Encoder) throws {
```
MouazAlzahabi, line 22: Should we have this method ?

Multisig/Data/Services/Safe Claiming Service/SafeClaimingService.swift
```diff
@@ -19,6 +19,7 @@ class SafeClaimingService {
         self.url = URL(string: "https://5afe.github.io/")!
         httpClient = JSONHTTPClient(url: url, logger: logger)
         httpClient.jsonDecoder.dateDecodingStrategy = .millisecondsSince1970
+        httpClient.jsonDecoder.keyDecodingStrategy = .convertFromSnakeCase
```
MouazAlzahabi: Isn't it better if we have this on the source files directly from now?

MouazAlzahabi (approved):

Multisig/Data/Services/Safe Claiming Service/AllocationsRequest.swift
DmitryBespalov: needed to implement encodable

Multisig/Data/Services/Safe Claiming Service/SafeClaimingService.swift
DmitryBespalov: ok, will update the data

Issue #2546, opened 2022-08-15
[SAFE] Adjust Get Started screen


DmitryBespalov:
- [x]  Text of the button changed.
- [x]  Screen structure has to be re-implemented
- [x]  Add scroll view for instructions to be visible on small screens

https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=341%3A6794

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/601c070e-7fd5-46ba-a20e-1a873c8057a3" width=300 />

DmitryBespalov:
QA note: background was removed because it was conflicting with the text in the dark mode

https://github.com/safe-global/safe-ios/pull/2607#issuecomment-1234041399

liliya-soroka:
Verified
app version 3.17.0 (1301)

PR #2607, opened 2022-08-31
Gh 2546/adjust get started screen

biafra23:
Handles #2546

- Background had to be remoced because there is no dark mode asset for it

light | dark
-----|-----
image| image


DmitryBespalov (approved):

elgatovital (approved):

TanyaEfremova:
The waves in the dark mode are too bright which reduces the readability drastically. We could remove them from the background in both modes, if it requires extra effort to make them less visible. They are not the main priority for the UI.

biafra23:
I removed the background and updated the screenshots

PR #2609, opened 2022-09-01
Integrate Keystone - connect Keystone Wallet

zhiying-fan:
### Description
This PR adds basic support for connect Keystone Wallet and get extended public key from Keystone Wallet.

### Changes
- Add `ConnectKeystoneFlow` which is inherited from `AddKeyFlow`, this is the main file I added.
- Extend `KeyPickerController` and `SelectOwnerAddressViewModel` to support public key.
- Add a new method in `KeyInfo` to import public key for keystone.

### Dependencies

- [URKit](https://github.com/BlockchainCommons/URKit): Decode QR code based on UR type. One dependency is `WolfBase`.
- [URRegistry](https://github.com/KeystoneHQ/ur-registry-ios): Our library used to get public key from UR and uncompress public key.

There are some TODOs will be finished in next few weeks.


Multisig/Logic/Models/PrivateKey/web3swift/BIP32HDNode.swift
```diff
@@ -235,7 +235,7 @@ extension HDNode {
             let newNode = HDNode()
             newNode.chaincode = cc
             newNode.depth = self.depth + 1
-            newNode.publicKey = pubKeyCandidate
+            newNode.publicKey = newPublicKey
```
zhiying-fan, line 238: This is a bug which is already fixed in the `web3swift` library itself, please see detail here: https://github.com/skywinder/web3swift/pull/423

zhiying-fan:
Will create a new one and include all of the work.

Issue #2591, opened 2022-08-29
[SAFE] "What is Safe Token?" screen


DmitryBespalov:
- [ ] Scrollable on small screens
- [ ] Detail screens open as modals, half-screen (see how it fits on small devices)

### Content 
[Education Presentation](https://docs.google.com/presentation/d/1LfJJ6RkHvRAlWN7rmaFy1XuSyxEL7diI78UPBgYASJY/edit#slide=id.g142e0570c2d_0_54): 

**What exactly is Safe token and what does it govern?**
$SAFE is an ERC-20 governance token that stewards infrastructure components of the Safe ecosystem, including:

**Safe Protocol**
Safe Deployments (core smart contract deployments across multiple networks)
Curation of “trusted lists” (Token lists, dApp lists, module lists)

**Interfaces**
Decentralized hosting of a Safe frontend using the safe.eth domain
Decentralized hosting of governance frontends

**On-chain assets**
ENS names
Outstanding Safe token supply
Other Safe Treasury assets (NFTs, tokens, etc.)

**Tokenomics**
Ecosystem reward programs
User rewards
Value capture
Future token utility

[Figma](https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=1232%3A10336)
<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/d1f6ab6f-a30c-4677-9246-81720ca8734e" width=300 />

[Figma](https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=1232%3A10588)
<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/72836982-cd7b-4aec-aff4-26fa60b56ab1" width=300 />

Next Screen: [Navigating DAO](https://app.zenhub.com/workspaces/mobile-engineering-60cc78740f488d000e996335/issues/safe-global/safe-ios/2592)

liliya-soroka:
Verified

app version 3.17.0 (1325)

PR #2618, opened 2022-09-05
Gh 2591/what is safe token

MouazAlzahabi:
Handles #2591


DmitryBespalov:
@MouazAlzahabi that's a good progress!

Probably you were not aware or missed, but this does not implement the issue as specified in the ticket (https://github.com/safe-global/safe-ios/issues/2591):

```
Scrollable on small screens
Detail screens open as modals, half-screen (see how it fits on small devices)
```

The designs in Figma do not show this, but - instead of opening detail text in a navigation controller, it's better to show them as a modal popover (half-screen size). 

Were there any reason you went with the 'push' navigation?

DmitryBespalov:
image

The chevron seems too small compared to Figma

DmitryBespalov:
there's no space between text paragraphs

image


but there's space in the design (in 3 out of 4 detail screens)

image

DmitryBespalov:
In 'navigating safe dao' text

image

is different from the designs (ellipsis missing)

image

DmitryBespalov:
If you solve paragraph spacing issue, it would be also good to adjust it in the `Legal Disclaimer` screen, so that it is easier to read.

image

Compare with the designs:

image

DmitryBespalov (approved):

Issue #2619, opened 2022-09-06
[iOS] Wrong passcode after updating to 3.16.0


DmitryBespalov:
**Bug description**
App does not recognize passcode after it was updated to 3.16.0 from previous version

**Steps To Reproduce**
1. Install 3.15.0. Load demo safe.
2. Set up a passcode. Without biometry.
3. Update to 3.16.0
4. Try to log in with the passcode
---
"Wrong passcode" error

**Expected Result**
Passcode still works

**Screenshots**
<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/c438f2cd-e28d-4b69-a5f4-333ee4ea59e8" width=300 />

**Device & App version**
 - Device/OS: 
 - App version: 3.16.0 (production)
 
**Environment**
Production

**Additional info**
(Optional)

DmitryBespalov:
3.14.0 → 3.15.0 updating doesn't break passcode
3.15.0 →3.16.0 update does break passcode

liliya-soroka:
Verified
app version 3.16.1

PR #2622, opened 2022-09-06
Fix issue

MouazAlzahabi:
Handles #2619

### DIFF
--- a/Multisig/Logic/AuthenticationController.swift
+++ b/Multisig/Logic/AuthenticationController.swift
@@ -112,6 +112,7 @@ class AuthenticationController {
     }
 
     private func derivedKey(from plaintext: String) -> String {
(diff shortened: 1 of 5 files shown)


Multisig/Logic/AuthenticationController.swift
```diff
@@ -112,7 +112,7 @@ class AuthenticationController {
     }
 
     private func derivedKey(from plaintext: String) -> String {
-        let salt = "Safe Multisig Passcode Salt"
+        let salt = "Gnosis Safe Multisig Passcode Salt"
```
DmitryBespalov, line 116: oh shit. @MouazAlzahabi please add a comment before this line that this must be kept the same so that the passcodes can be decoded from earlier versions

DmitryBespalov (approved):

biafra23 (approved):

Issue #2553, opened 2022-08-15
[SAFE] Adjust Review Claim screen


DmitryBespalov:
- [x]  Design changes
    - [x]  “You’re claiming”, “Delegating to”
- [x]  MultiSend List
    - [x]  Update to the designs - for all multi-send details
    - [x]  Network banner
    - [x]  Change ellipsis algo: should be in the middle

https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=547%3A8370

DmitryBespalov:
Note on MultiSend screens: for some reason, when opening the screen, the large title is not displayed automatically. I have spent some time on trying to make it work, but I found no way to enable it.

I leave it as is. If you scroll those screens just a little, the large title appears, otherwise it is a short title.

I have also not changed the title of the multi-send screen to "Contract Interaction", because it needs to be "MultiSend" since that is the name of the method call. The designs are incorrect in this.

liliya-soroka:
Verified
app version 3.17.0 (1326)

PR #2628, opened 2022-09-07
Gh 2553/review screen

DmitryBespalov:
Handles #2553

- Header with information (in the style of other review screens)
- MultiSend item style change


elgatovital (approved):

Multisig/UI/ClaimToken/ReviewClaimTokensHeaderCell.swift
```diff
@@ -0,0 +1,83 @@
...
+
+import UIKit
+
+class ReviewClaimTokensHeaderCell: UITableViewCell {
+    @IBOutlet private weak var amountView: TokenInfoView!
+    @IBOutlet private weak var delegateView: AddressInfoView!
+    @IBOutlet private weak var fromView: AddressInfoView!
+    @IBOutlet private weak var toView: AddressInfoView!
+
+    var showsDelegate: Bool = true {
```
MouazAlzahabi, line 17: Maybe instead of this we could hide delegate view always until setDelegate called with value ?

Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
```diff
@@ -149,19 +149,20 @@ class ReviewSafeTransactionViewController: UIViewController {
...
                         chainId: self.safe.chain!.id!
                     ) { result in
-                        switch result {
-                        case .success(let response):
-                            self.transactionPreview = response
-                            self.onSuccess()
-                        case .failure(let error):
-                            DispatchQueue.main.async { [weak self] in
-                                guard let `self` = self else { return }
+                        DispatchQueue.main.async {
```
MouazAlzahabi, line 160: just for info, does this solved any noticeable issue so far ?

Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
```diff
@@ -149,19 +149,20 @@ class ReviewSafeTransactionViewController: UIViewController {
                     self.safeTxGas = UInt256String(estimatedSafeTxGas)
                 }
 
-                if self.shouldLoadTransactionPreview {
+                if self.shouldLoadTransactionPreview, let transaction = self.createTransaction() {
```
MouazAlzahabi, line 152: check the beginning of this method (line 113) it contains the same check, is it necessary to do it again ?

MouazAlzahabi (approved):

Multisig/UI/ClaimToken/ReviewClaimTokensHeaderCell.swift
DmitryBespalov: can you explain your reasoning a bit, i.e. why this would be better?

Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
DmitryBespalov: no noticeable issue, but I noticed this is called on background thread, but the onSuccess is calling UI-related code

Multisig/UI/Transaction/InitiateTransaction/ReviewSafeTransactionViewController/ReviewSafeTransactionViewController.swift
DmitryBespalov: yes because the safeTxGas and safeNonce have changed in the lines above, and the transaction is outdated now

DmitryBespalov: however, those parameters seem irrelevant for the preview endpoint, but still, I was trying to debug it and found that it is using not updated values after estimation

Multisig/UI/ClaimToken/ReviewClaimTokensHeaderCell.swift
DmitryBespalov: to me those are equivalent and not much of a difference

MouazAlzahabi: with the current implementation in order to present delegate we need to set it first then set showsDelegate, with my suggestion by just setting the delegate it should be showed

Issue #2554, opened 2022-08-15
[SAFE] Adjust Transaction Creation Success screen


DmitryBespalov:
- [x]  Done → Close success, open Tx Details
- [x]  Share Transaction → Open share screen
- [x]  Close → Close screen 
- [x] Twitter box

https://www.figma.com/file/a0BQ5MEgKe6qsZXC5XTWHc/Claim-S-Tokens-Mobile-App?node-id=791%3A8289

<img src="https://images.zenhubusercontent.com/5c361cfa7727940fef83693b/418f0cdf-6d49-4f51-afd4-4518d74b3114" width=300 />

liliya-soroka:
Verified
app version 3.17.0 (1332)

PR #2631, opened 2022-09-07
gh-2554 success screen changes

DmitryBespalov:
Handles #2554 

- UI is scrollable on small screens
- Tweet text as in figma (3 cases)


Multisig/UI/UI Library/String+Highlight.swift
```diff
@@ -37,25 +37,13 @@ extension String {
...
                         textToHighlight: String? = nil) -> NSAttributedString {
-            var style = NSMutableParagraphStyle()
-            style.lineSpacing = 10
-
-            var attributedString = NSMutableAttributedString(
-            string: self,
-            attributes: [
-                .font: UIFont.gnoFont(forTextStyle: originalStyle),
-                .foregroundColor: originalStyle.color!,
-                .kern: -0.41,
```
DmitryBespalov, line 48: I removed it because I thought that in Figma it is an 'automatic' style, i.e. we just use our system font without kerning.

Kerning is set in the GNOTextSTyle

MouazAlzahabi (approved):

Issue #2643, opened 2022-09-08
[iOS][claiming app] Interact with contract(Multisend) on mobile is diff from the web(multisendonly)


liliya-soroka:
**Bug description**
 Interact with contract(Multisend) on mobile is diff from the web(multisendonly)
**Steps To Reproduce**
1. claim Safe tokens on mobile
Current result: interact contract = rin:0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761 ( I think it's our Multisend) - https://safe-team.dev.gnosisdev.com/app/rin:0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD/transactions/multisig_0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD_0xcc9b560442fe5eec1839a622478f081799c0ad91826f7f9eb7c30d214f876793

2. claim safe tokens on web
Current result: interact contract = rin:0x40A2aCCbd92BCA938b02010E17A5b8929b49130D( Gnosis Safe: MultiSendCallOnly) - https://safe-team.dev.gnosisdev.com/app/rin:0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD/transactions/multisig_0xf2565317F3Ae8Ae9EA98E9Fe1e7FADC77F823cbD_0x7f665b011c24e7587a5a8227ac4531322ed4f3f5e0375b628a5884cd023f0606

**Expected Result**
Please, check what contract should be used for the safe with contract >= 1.3.0 and for the old safes

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.17.0 (1325)
 
**Environment**
staging
**Additional info**
(Optional)

DmitryBespalov:
As far as I tried it, `redeem` will only work with delegate call multi-send [(contract requirement)](https://github.com/safe-global/safe-token/blob/2e176db9c2801afc1579ac00ff9c539ba840ef0f/contracts/Airdrop.sol#L56)

So, I continued to use the full-featured multi-send for other types of transactions. 

Safes with version < 1.3.0 use 1.1.1 multi-send contract

DmitryBespalov:
I believe this is not a bug

DmitryBespalov:
After analysis and discussion, *this is a bug*.

Currently, we create the following transaction calls:
```
Safe.delegateCall( MultiSend.delegateCall( Airdrop.redeem() ) )
...
```

But instead we should do:
```
[...]

PR #2645, opened 2022-09-08
call MultiSendCallOnly instead of MultiSend

MouazAlzahabi:
Handles #2643


elgatovital (approved):

DmitryBespalov (changes requested):
this will break tests, the tests need to be adjusted

DmitryBespalov:
Also, need to double-check that our single transactions are not delegate calls

DmitryBespalov:
Ok, it did not break tests because there was no tests for this (in AllTests scheme).

Still, I reverted to use MultiSend for contract versions < 1.3.0

DmitryBespalov (approved):

PR #2663, opened 2022-09-09
gh-2154 Fixed gas estimation on polygon

DmitryBespalov:
Handles #2154

https://docs.polygon.technology/docs/develop/eip1559-transactions/how-to-send-eip1559-transactions/

https://docs.polygon.technology/docs/edge/get-started/json-rpc-commands/#eth_call

- Not using "pending" for Polygon
- Using only "maxPriorityFeePerGas" for Polygon transactions

### DIFF
--- a/Packages/Ethereum/Sources/Ethereum/EthRpc1.swift
+++ b/Packages/Ethereum/Sources/Ethereum/EthRpc1.swift
@@ -334,6 +334,9 @@ public enum EthRpc1 {
         /// from address
         public var from: EthRpc1.Data? = nil
 
+        enum CodingKeys: String, CodingKey {
+            case from, to, gas, gasPrice, maxPriorityFeePerGas, maxFeePerGas, value, data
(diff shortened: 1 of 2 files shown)


Packages/Ethereum/Sources/Ethereum/EthRpc1.swift
```diff
@@ -334,6 +334,9 @@ public enum EthRpc1 {
         /// from address
         public var from: EthRpc1.Data? = nil
 
+        enum CodingKeys: String, CodingKey {
```
elgatovital, line 337: are these used somewhere?

DmitryBespalov: they are automatically picked up by Swift JSON Encoder, so only keys listed here will be encoded ('type' is not encoded)

elgatovital (approved):

Issue #2656, opened 2022-09-09
[iOS][claiming app]delegates list is broken when nothing is found using search 


liliya-soroka:
**Bug description**
delegates list is broken when nothing is found using search 

Precondition:
make sure that you claimed at least once and the flow is started from the claiming screen for you
**Steps To Reproduce**
1. go back to the delegates list
2. search for a delegate that doesn't exist
Current result: the list is hidden partly under black screen instead of no delegate text
**Expected Result**
no delegates should be displayed
**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/5cf99d2d-4b3e-4044-bac2-bb2b7ba903d4)
**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.17.0 (1332)
 
**Environment**
staging
**Additional info**
(Optional)

biafra23:
I cannot reproduce this. Neither on iPad nor on iPhone SE

MouazAlzahabi:
nor too

liliya-soroka:
Verified
app version 3.17.0 (1346)

PR #2667, opened 2022-09-12
Reimplement search logic

MouazAlzahabi:
Handles #2656

### DIFF
--- a/Multisig/UI/ClaimToken/Select Guardian/GuardianListViewController.swift
+++ b/Multisig/UI/ClaimToken/Select Guardian/GuardianListViewController.swift
@@ -12,23 +12,31 @@ import UIKit
 class GuardianListViewController: LoadableViewController {
     private var currentDataTask: URLSessionTask?
     private enum Section {
-        case guardiansCount(count: Int)
-        case guardians(items: [Guardian])
+        case guardiansCount
+        case guardians
     }
 
     private var sections: [Section] = []
 
     private var guardians: [Guardian] = []
+    var filteredGuardians: [Guardian] = []
 
     var onSelected: ((Guardian) -> ())?
     var onReloaded: (() -> ())?
 
     private var searchController: UISearchController!
-    private var resultsController: GuardianSearchResultController!
 
     var controller: ClaimingAppController!
     var safe: Safe!
     var selectedDelegate: Address?
+
+    var isSearchBarEmpty: Bool {
+        searchController.searchBar.text?.isEmpty ?? true
+    }
+
+    var isFiltering: Bool {
+        searchController.isActive && !isSearchBarEmpty
+    }
 
     convenience init() {
         self.init(namedClass: Self.superclass())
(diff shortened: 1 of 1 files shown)


Multisig/UI/ClaimToken/Select Guardian/GuardianListViewController.swift
```diff
@@ -63,22 +68,23 @@ class GuardianListViewController: LoadableViewController {
...
-        }
+    private func makeSections(items: [Guardian]) {
+        self.sections = []
+        guard !items.isEmpty else { return }
 
         var sections = [Section]()
 
-        sections.append(.guardiansCount(count: items.count))
-        sections.append(.guardians(items: items))
+        if !isFiltering {
```
elgatovital, line 81: what about normal state? Why not adding .guardiansCount then?

Multisig/UI/ClaimToken/Select Guardian/GuardianListViewController.swift
```diff
@@ -120,17 +126,28 @@ class GuardianListViewController: LoadableViewController {
...
     func updateSearchResults(for searchController: UISearchController) {
-        guard let resultsController = searchController.searchResultsController as? GuardianSearchResultController else {
-            return
-        }
-        if !searchController.searchBar.text!.isEmpty {
+        if !isSearchBarEmpty {
             Tracker.trackEvent(.userClaimChdelSearch)
+
+        } else if !isSearchBarEmpty && filteredGuardians.isEmpty {
+            Tracker.trackEvent(.screenClaimChdelNf)
```
elgatovital, line 133: looks like a typo in a screen name

Multisig/UI/ClaimToken/Select Guardian/GuardianListViewController.swift
```diff
@@ -120,17 +126,28 @@ class GuardianListViewController: LoadableViewController {
...
         }
+
         let terms = searchController.searchBar.text!
             .trimmingCharacters(in: .whitespacesAndNewlines)
             .components(separatedBy: " ") as [String]
 
-        resultsController.filteredGuardians = guardians.filter { guardian in
+        filterContentForSearchText(terms)
+        makeSections(items: isFiltering ? filteredGuardians : guardians)
```
elgatovital, line 141: why is this distinction whether result set is being filtered needed?

elgatovital: filtered guardians could always be used and in case of unfiltered state filtered guardians = unfiltered guardians. This way some if checks could be saved.

DmitryBespalov (changes requested):
I think that reimplementing the search to fix the ui gap is incorrect solution, because it does what the system already is doing.Observing that the gap looks like a missing navigation bar (large size), and searching on stack overflow, I would first try other way: https://stackoverflow.com/questions/57521967/ios-13-strange-search-controller-gap

Disable large title and 'extendedLayoutIncludesOpaqueBars = true' on the search results view controller

Issue #2699, opened 2022-10-04
Change staging service URL


DmitryBespalov:
Change client gateway URL:

safe-client.staging.gnosisdev.com → safe-client.staging.5afe.dev

**Does this server name change affect SSL pinning? Should it?**

liliya-soroka:
Verified
app version 3.18.0 (1389)

PR #2701, opened 2022-10-04
GH-2699 - Updated CGW url for staging and dev

biafra23:
Handles #2699 

- gnosisdev.com -> 5afe.dev

**Does this server name change affect SSL pinning? Should it?**


DmitryBespalov (approved):

DmitryBespalov:
@biafra23 re ssl pinning: the 5afe.dev is not pinned in the app, it needs to be added, and the gnosisdev.com needs to be removed.

This is how to do in the iOS: 
https://datatheorem.github.io/TrustKit/documentation/Classes/TrustKit.html

DmitryBespalov:
In Info.plist there's an "TSKConfiguration" entry that holds the pinning hashes for the pinned domains

Issue #2709, opened 2022-10-06
[Rebrand] Fix known issues (General)


TanyaEfremova:
- [x] Dark Mode: some backgrounds use legacy colours (or system background colours). E.g.: Background in importing owner keys ![IMG_3636.PNG](https://images.zenhubusercontent.com/6183a79a848bb3312854c951/7cdeccd1-a9b0-48a5-b699-d717b6236387) 

- [ ] Remove separators in the end of the list items and section headers, e.g. see connecting wallets
![IMG_3637.PNG](https://images.zenhubusercontent.com/6183a79a848bb3312854c951/3ce65e35-8fae-48e0-92a7-33ab3fc48065)

- [ ] The new typography is not applied in the app

liliya-soroka:
Verified
app version 3.18.0 (1394)

PR #2713, opened 2022-10-09
Fix background and update fonts

MouazAlzahabi:
Handles #2709


DmitryBespalov:
epic work going on 👏 keep going!

Multisig/App/AppDelegate.swift
```diff
@@ -63,6 +63,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             UINavigationBar.appearance().scrollEdgeAppearance = appearance
         }
 
+        UITableView.appearance().backgroundColor = .backgroundPrimary
```
MouazAlzahabi: @DmitryBespalov are you fine to use the Appearance protocol?

DmitryBespalov: I don't remember having issues in the past except that we need to be aware that there's this global appearance is set when debugging. In the past it was difficult to find why something is having non-standard setting, and it turned out to be because of appearance, which I wasn't aware of. 

So, nothing against it at this point 👍

elgatovital:
@MouazAlzahabi there are some conflicts in Multisig.xcodeproj/project.pbxproj and build is failing

DmitryBespalov:
This screen has old fonts

DmitryBespalov:
Please check the navigation bar text fonts and the network ribbon text

DmitryBespalov:
assets-difference

DmitryBespalov:
claim-banner

DmitryBespalov:
banner

DmitryBespalov:
load-enter-address

DmitryBespalov:
error-colors

DmitryBespalov:
enter-ens

DmitryBespalov:
settings table section style
transactions table section headers

DmitryBespalov:
> error-colors

Ok, so I checked the color with a screenshot made from the Simulator - color matches the Assets' and Figma color. BUT when I pick the color from the simulator screen itself, OR when I pick the color from a screenshot of the device - the color is slightly different.

This has to do something with a color management of the device / simulator and I don't see a way to control it. 

With the naked eye I see the difference, but apparently I don't see a way to change it.

DmitryBespalov (approved):

Issue #2702, opened 2022-10-05
[Rebrand] Fix known issues in Onboarding


MouazAlzahabi:
- [x] Onboarding steps icons not match 
- [x] Onboarding steps titles highlighting is not there

liliya-soroka:
Verified
app version 3.18.0 (1404)

PR #2720, opened 2022-10-20
Highlight title

MouazAlzahabi:
Handles #2702

### DIFF
--- a/Multisig/UI/App/Onboarding/OnboardingStepCollectionViewCell.swift
+++ b/Multisig/UI/App/Onboarding/OnboardingStepCollectionViewCell.swift
@@ -17,8 +17,16 @@ class OnboardingStepCollectionViewCell: UICollectionViewCell {
     static let identifier = "OnboardingStepCollectionViewCell"
 
     func configure(step: OnboardingStep) {
-        titleLabel.text = step.title
-        titleLabel.setStyle(.title1)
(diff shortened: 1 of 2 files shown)


Multisig/UI/App/Onboarding/OnboardingStepCollectionViewCell.swift
```diff
@@ -17,8 +17,16 @@ class OnboardingStepCollectionViewCell: UICollectionViewCell {
...
-        titleLabel.text = step.title
-        titleLabel.setStyle(.title1)
+        if let highlightedText = step.title.highlightedText {
+            titleLabel.attributedText = step.title.text.highlightRange(
+                originalStyle: .title1,
+                highlightStyle: .title1.color(.primary),
+                textToHighlight: highlightedText
+            )
+        } else {
+            titleLabel.text = step.title.text
```
DmitryBespalov, line 27: Is there a situation when there's no highlighted text provided?

DmitryBespalov (approved):

Multisig/UI/App/Onboarding/OnboardingStepCollectionViewCell.swift
MouazAlzahabi: General case yes, but actually we don't have it right now

Issue #2721, opened 2022-10-21
Crash when duplicate transaction hash in the data


DmitryBespalov:
[Crashlytics](https://console.firebase.google.com/u/1/project/safe-firebase-mainnet/crashlytics/app/ios:io.gnosis.multisig.prod.mainnet/issues/cf9e16543fb9560776f9b7d48f033efe?time=last-seven-days&sessionEventKey=4e3a77ad09f14dad8f9dd3727bdb256e_1735176487211692983)

[Support ticket](https://github.com/5afe/safe-support/issues/42)

It seems that this would happen if user manages to execute the same safe transaction twice, i.e. submit it twice with the same parameters. 

I have not found how to reproduce it via user interface, however, I could reproduce the crash if I programmatically create duplicate transactions in the database.

liliya-soroka:
closed, will be checked after release by the user

PR #2722, opened 2022-10-21
Gh 2721/patch duplicate eth tx hash

DmitryBespalov:
Handles #2721 

- Fixes crash due to duplicate tx hash values and prevents creation of duplicates

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
@@ -779,6 +779,21 @@ class CreateSafeFormUIModel {
 
         // save the tx information for monitoring purposes
         let context = App.shared.coreDataStack.viewContext
+
+        // prevent duplicates
+        let ethTxHash = txHash.storage.storage.toHexStringWithPrefix()
+
+        let fr = CDEthTransaction.fetchRequest().by(ethTxHash: ethTxHash, chainId: chain.id!)
+        do {
+            let toRemove = try context.fetch(fr)
+            for tx in toRemove {
+                context.delete(tx)
+            }
+        } catch {
+            LogService.shared.error("Failed to save transaction for monitoring: \(error)")
+            return
+        }
+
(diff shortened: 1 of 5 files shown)


Multisig/UI/Transaction/ExecuteTransaction/PendingTransactionMontior.swift
```diff
@@ -128,10 +128,10 @@ class ChainPendingTransactionMonitor {
         // the core data objects might change while we're making the network request
         // therefore we need to remember the data that will tell us that we still should udpate the objects
         // after network request is completed.
-        let cdTxByTxHash: [String: (status: String, updatedAt: Date?, ethTx: CDEthTransaction)] = Dictionary(
-            uniqueKeysWithValues: cdEthTransactions.map { ($0.ethTxHash!, ($0.status!, $0.dateUpdatedAt, $0)) }
-        )
-
+        //
+        // patch: using array instead of a dictionary (by tx hash) because of duplicate entries in the database
```
MouazAlzahabi, line 132: As we remove the duplication in the above file, do we still need this change here ? or just for defensive coding?

MouazAlzahabi (approved):

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
```diff
@@ -779,6 +779,21 @@ class CreateSafeFormUIModel {
 
         // save the tx information for monitoring purposes
         let context = App.shared.coreDataStack.viewContext
+
+        // prevent duplicates
+        let ethTxHash = txHash.storage.storage.toHexStringWithPrefix()
```
MouazAlzahabi, line 783: Can we have reusable function instead of this duplicated code?

Multisig/UI/Transaction/ExecuteTransaction/PendingTransactionMontior.swift
DmitryBespalov: out there in production, on the user device, the database still has duplicate entries. This will lead to a crash if we won't handle it here.

Issue #2723, opened 2022-10-21
[iOS][safe settings] Safe name title is missed


liliya-soroka:
**Bug description**

**Steps To Reproduce**
1. go to the safe settings page
Current result:
![image.png](https://images.zenhubusercontent.com/5ed757c29459bb36bacd4bba/e3fe62b7-6fa3-4ddf-87b1-239726c4c33c)
**Expected Result**
Safe name title should be above the field with safe name

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.18.0 (1394)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.18.0(1413)

PR #2752, opened 2022-10-26
Add safe name title

MouazAlzahabi:
Handles #2723


DmitryBespalov (approved):
Mystery why it was working before

MouazAlzahabi:
> Mystery why it was working before

it wasn't, actually

Issue #2777, opened 2022-11-07
[Keysec1] Setup Secure Enclave "Key Encryption Key" (KEK) and Sensitive Key


biafra23:
# Description

This task is about setting up an infrastructure to be able to perform cryptography in order to protect private keys.

# Before we can add signer keys to the Keychain, we need to:

1. Create a KEK (Key Encryption Key) in the secure enclave according to the [Passcode Settings](https://www.notion.so/gnosis-safe/Passcode-operations-5ccdd7d9f20940999a58eeb5c2fa6315#99ce729adcc54976aab54f93fab07ab6) (password, biometry etc.) 
2. Create a `Sensitive Key` and store the public part to be used to encrypt signer keys later
3. Encrypt private part with `KEK` and store the Data in KeyChain

# Acceptance Criteria

* Nothing user facing will be the result of this ticket

# Resources

* [Safe iOS security](https://www.notion.so/gnosis-safe/Safe-iOS-App-Security-b4d558f05f9649588c2dfb46c89d4de8) in Notion

Issue #2775, opened 2022-11-07
[keysec1] Find and use signer key for signing (and signature verification) and export


biafra23:
# Description

This task will allow using the signing key for blockchain operations

# What To Do

Implement `findKeyByAddress()`
1. Retrieve the encrypted data for signer key
2. Retrieve encrypted Sensitive Key
3. Decrypt sensitive key with KEK
4. Decrypt signer key
5. Return signer key ( we need to do this for backups, but we could implement signing without handing out the key)

# Acceptance Criteria

* import a key
* check it can be retrieved

# Resources

* [Safe iOS security](https://www.notion.so/gnosis-safe/Safe-iOS-App-Security-b4d558f05f9649588c2dfb46c89d4de8) in Notion

Issue #2774, opened 2022-11-07
[Keysec1] Import signer key


biafra23:
# Description

As a safe owner, 
I'd like to use my signer key in the app,
So that I can create, confirm, and execute wallet transactions.

# What To Do

Implement `importKey()`
1. Use the sensitive public key to encrypt signer private key
2. Store the encrypted Data under its address so it can be retrieved by `findKey()`

# Acceptance Criteria

* Test the key can be found after import

# Resources

* [Safe iOS security](https://www.notion.so/gnosis-safe/Safe-iOS-App-Security-b4d558f05f9649588c2dfb46c89d4de8) in Notion

PR #2792, opened 2022-11-17
Setup EncryptionStore, import and find Ethereum keys

biafra23:
Handles #2777 #2774 #2775

- Initially setup secure enclave key (KEK) and use a randomly generated application password
- Encrypt `Sensitive Key` 
- Store the `Sensitive Key` (public part) unencrypted
- Store  encrypted `Sensitive Key` (private part)
- Also store the initial application password
- Setup automatic testing for KeychainStorage and EncryptedSensitiveStore
- import(ethPrivateKey: EthPrivateKey)
- find(address: Address, password: String?) -> EthPrivateKey
- unit tests

Things to clarify during review:
- When throw? And what Error-Class
- What to return if nothing found? nil?


MultisigTests/Features/Security/EncryptedSensitiveStoreTests.swift
```diff
@@ -0,0 +1,66 @@
...
+
+        // When
+        try encryptedStore.import(ethPrivateKey: randomKey)
+
+        // Then
+        let ethPrivateKey = try encryptedStore.find(address: "0xE86935943315293154c7AD63296b4e1adAc76364", password: nil)
+        XCTAssertEqual(ethPrivateKey, randomKey)
+    }
+
+// TODO
```
elgatovital: could be deleted

MultisigTests/Features/Security/EncryptedSensitiveStoreTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    var keychainStorage: KeychainStorage! = nil
+
+    public override func setUp() {
+        super.setUp()
+        keychainStorage = KeychainStorage()
+        encryptedStore = SensitiveEncryptedStore(keychainStorage)
+    }
+
+    func testInitialSetup() {
+        // Given
```
elgatovital: comments don't really make sense here

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+    init(_ keychainStorage: KeychainStorage) {
+        self.keychainStorage = keychainStorage
+    }
+
+    init() {
+        keychainStorage = KeychainStorage()
+    }
+
+    // This is called, when using the new key security is activated after updating the app. Even if the user did not activate it.
+    func initialSetup() throws {
```
DmitryBespalov: need better name that describes purpose

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+
+    }
+
+    func delete(address: Address) {
+        // delete encrypted blob by address
+    }
+
+    // Copied from AuthenticationController
+    private func derivePasscode(from plaintext: String, useOldSalt: Bool = false) -> String {
+        let salt = "Safe Multisig Passcode Salt"
```
DmitryBespalov: needs to handle old salt and old parameters if used elsewhere

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+
+    private func persistRandomPassword() throws -> String {
+        let randomPasscode = createRandomPassword()
+        let derivedPasscode = derivePasscode(from: randomPasscode)
+        try keychainStorage.storePasscode(derivedPasscode: derivedPasscode) // TODO check error?
+        return derivedPasscode
+    }
+
+    private func createRandomPassword() -> String {
+        Data.randomBytes(length: 32)!.toHexString()
```
DmitryBespalov: use SecRandomCopyBytes directly

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+    func initialSetup() throws {
+        let derivedPasscode = try persistRandomPassword()
+        let sensitiveKey = try keychainStorage.createKeyPair()
+        try persistSensitivePublicKey(sensitiveKey: sensitiveKey)
+        try persistSensitivePrivateKey(derivedPasscode: derivedPasscode, sensitiveKey: sensitiveKey)
+    }
+
+    private func persistRandomPassword() throws -> String {
+        let randomPasscode = createRandomPassword()
+        let derivedPasscode = derivePasscode(from: randomPasscode)
```
DmitryBespalov: deriving password is not needed with random password

Multisig/Features/Security/SensitiveEncryptedStore.swift
DmitryBespalov: Also, please inline it.

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+    init() {
+        keychainStorage = KeychainStorage()
+    }
+
+    // This is called, when using the new key security is activated after updating the app. Even if the user did not activate it.
+    func initialSetup() throws {
+        let derivedPasscode = try persistRandomPassword()
+        let sensitiveKey = try keychainStorage.createKeyPair()
+        try persistSensitivePublicKey(sensitiveKey: sensitiveKey)
+        try persistSensitivePrivateKey(derivedPasscode: derivedPasscode, sensitiveKey: sensitiveKey)
```
DmitryBespalov: what happens when this is done in the background when device is locked?

DmitryBespalov: how to check that updating / setup was successful on the next startup? How do we know that previous attempt failed?

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+    static let derivedPasswordTag = "global.safe.password.as.data"
+    static let sensitiveKekTag = "global.safe.sensitive.KEK"
+
+    // used to store encrypted data as a password in keychain
+    static let defaultService: String = "encrypted_data"
+
+    // store passcode. Either if it random (user didn't give a password) or the user asked us to remember it
+    func storePasscode(derivedPasscode: String) throws {
+        // Store encrypted Password data
+        try storePasswordData(passwordData: derivedPasscode.data(using: .utf8)!)
```
DmitryBespalov: we could make this the main interface, without "derived passcode"

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+        case let status:
+            throw convertStatusToError(status: status)
+        }
+    }
+
+    func deleteData(_ account: String) throws {
+        let query = SQuery.generic(id: account).queryData()
+
+        let status = SecItemDelete(query)
+        switch status {
```
DmitryBespalov: please try to use `if ` and see if it makes it more readable

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+        } catch {
+            return nil
+        }
+    }
+
+    private func storePasswordData(passwordData: Data) throws {
+        // delete existing sensitive key data
+        try deleteData(KeychainStorage.derivedPasswordTag)
+        // Create query
+        let addQuery = SQuery.generic(id: KeychainStorage.derivedPasswordTag, encryptedData: passwordData).queryData()
```
DmitryBespalov: why is query used here instead of `SItem`?

DmitryBespalov: `SQuery` <> `SItem` either needs to merge and become one with 2 functions: create search query and create attributes for item

Multisig/Features/Security/SQuery.swift
```diff
@@ -0,0 +1,141 @@
+//
+// Created by Dirk Jäckel on 17.11.22.
+// Copyright (c) 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import LocalAuthentication
+
+enum SQuery {
```
DmitryBespalov: Rename to `item search query`

Multisig/Features/Security/SQuery.swift
```diff
@@ -0,0 +1,141 @@
...
+    case generic(id: String, service: String, data: Data)
+    case enclaveKey(tag: String = KeychainStorage.sensitiveKekTag)
+    case ecKey(_ tag: String?)
+
+    // applies access flags and password to any type of item
+    func attributes(access: SecAccessControlCreateFlags? = nil, password: Data? = nil) throws -> NSDictionary {
+        var result: NSMutableDictionary = [:]
+        switch self {
+        case let .generic(id, service, data):
+            break
```
DmitryBespalov: use this for creating an item

Multisig/Features/Security/SQuery.swift
```diff
@@ -0,0 +1,141 @@
...
+        }
+        self.init()
+        let success = setCredential(appPassword, type: .applicationPassword)
+        guard success else {
+            return nil
+        }
+    }
+}
+
+enum SItem {
```
DmitryBespalov: rename to `KeychainItem`

Multisig/Features/Security/SQuery.swift
```diff
@@ -0,0 +1,141 @@
...
+// Copyright (c) 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import LocalAuthentication
+
+enum SQuery {
+    case generic(id: String, service: String = KeychainStorage.defaultService, encryptedData: Data? = nil)
+    case ecPrivateKey(tag: String = KeychainStorage.sensitiveKekTag, password: Data? = nil)
+    case ecPubKey(tag: String = KeychainStorage.sensitivePublicKeyTag, pubKeyData: SecKey? = nil)
```
DmitryBespalov: document the abbreviation: `ecPubKey` = Elliptic Curve Public Key, etc.

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+        // safe to Keychain (as type password?) using SecItemAdd() and sensitiveEncryptedPrivateKeyTag
+        let status = SecItemAdd(addQuery, nil) // TODO consider passing error ref instead of nil
```
DmitryBespalov: please add comment why we pass `nil` - because we don't need to return newly added item

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+        // setCredential() returns false on the Simulator.
+        // This means on the simulator the created SE key is not protected by the application password.
+        let result = authenticationContext.setCredential(passwordData, type: .applicationPassword)
+
+        return (result, authenticationContext)
+    }
+
+    // used to create a public-private key pair (asymmetric) NOT in secure enclave -> Sensitive Key
+    func createKeyPair() throws -> SecKey {
+        let attributes = try SItem.ecKey(nil).attributes()
```
DmitryBespalov: the tag is not used. Please create a new `case` for a public key or a key pair

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+        var error: Unmanaged<CFError>?
+        guard let keyPair = SecKeyCreateRandomKey(attributes, &error) else {
```
DmitryBespalov: could we use `SItem` as an argument so that it specifies what type of thing we want to create, and KeychainStorage is creating it and storing it.

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,240 @@
...
+        guard status == errSecSuccess else {
+            throw GSError.GenericPasscodeError(reason: "Cannot store public key")
+        } // Should be a new and more specific error type
+    }
+
+    func retrieveSensitivePublicKey() throws -> SecKey? {
+        try findKey(query: SQuery.ecPubKey())
+    }
+
+    private func createSEKey(flags: SecAccessControlCreateFlags, applicationPassword: String) throws -> SecKey {
```
DmitryBespalov: I would merge all those `create` methods into one and pass `SItem` (or items) as an argument, including flags. Or make the `create` method accept a dictionary, and then depending on attribute of `kSecItemClass` to choose between `SecItemAdd` or `SecKeyCreateRandomKey`

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        let derivedPasscode = derivePasscode(from: randomPasscode)
+        try keychainStorage.storePasscode(derivedPasscode: derivedPasscode) // TODO check error?
+        return derivedPasscode
+    }
+
+    private func createRandomPassword() -> String {
+        Data.randomBytes(length: 32)!.toHexString()
+    }
+
+    private func persistSensitivePrivateKey(derivedPasscode: String, sensitiveKey: SecKey) throws { // create SE key (KEK) with a hard coded tag for example: "sensitive_KEK"
```
DmitryBespalov: maybe just use `passcode`

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        let sensitiveKEK = try keychainStorage.createSecureEnclaveKey(
```
DmitryBespalov: move the setup of KEKs to the `initialSetup` function, otherwise it is hidden.

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        guard let sensitiveKeyData = SecKeyCopyExternalRepresentation(sensitiveKey, &error) else {
+            throw error!.takeRetainedValue() as Error
+        }
+        LogService.shared.info("sensitiveKeyData: \((sensitiveKeyData as Data).toHexString())")
+
+        // Copy public KEK Key to encrypt sensitive key)
+        let sensitivePublicKek = SecKeyCopyPublicKey(sensitiveKEK)
+        // encrypt private part of sensitive_key
+        // encrypt data using: SecKeyCreateEncryptedData using sensitiveKEK
+        guard let encryptedSensitiveKey = SecKeyCreateEncryptedData(sensitivePublicKek!, .eciesEncryptionStandardX963SHA256AESGCM, sensitiveKeyData, &error) as? Data else {
```
DmitryBespalov: can we extract the encryption function into a separate operation, so that we can use it for encrypting key and encrypting data?

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        let sensitiveKEK = try keychainStorage.createSecureEnclaveKey(
+                useBiometry: false,
+                canChangeBiometry: true,
+                applicationPassword: derivedPasscode
+        )
+
+        // Convert SecKey -> Data SecKeyCopyExternalRepresentation -> CFData -> Data
+        // Copy private part of sensitive key
+        var error: Unmanaged<CFError>?
+        guard let sensitiveKeyData = SecKeyCopyExternalRepresentation(sensitiveKey, &error) else {
```
DmitryBespalov: can we use `find()` api to 1) fetch the public KEK, and then use `encrypt()` api to encrypt?

DmitryBespalov: Also, store the public KEK

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        let sensitivePublicKey = SecKeyCopyPublicKey(sensitiveKey)
+        // safe it via keychainStorage.storeSensitivePublicKey()
+        if let key = sensitivePublicKey {
+            try keychainStorage.storeSensitivePublicKey(publicKey: key)
+        } else {
+            throw GSError.GenericPasscodeError(reason: "Cannot copy public key")
+        }
+    }
+
+    func `import`(ethPrivateKey: EthPrivateKey) throws {
```
DmitryBespalov: make argument binary `Data`

DmitryBespalov: and another parameter for identifier

DmitryBespalov: I would prepend some identifier to the `address` so that this sensitive store items have different ids from other data items

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        // 4. store encrypted blob in the keychain
+        let address = privateKey.address
+        try keychainStorage.storeData(encryptedData: encryptedSigningKey, account: address.checksummed)
+    }
+
+    /// Find private signer key.
+    /// - parameter address: find ky for this address
+    /// - parameter password: application password. Can be nil, then the sored password is used
+    /// - returns: String with hex encoded bytes of the private key
+    func find(address: Address, password: String? = nil) throws -> EthPrivateKey? {
```
DmitryBespalov: `address → id`
`return type →binary data`

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        // find encrypted sensitive key
+        let encryptedSensitiveKeyData = try keychainStorage.retrieveEncryptedData(account: KeychainStorage.sensitiveEncryptedPrivateKeyTag)!
+
+        // decrypt encrypted sensitive key
+        var error: Unmanaged<CFError>?
+        guard let decryptedSensitiveKeyData = SecKeyCreateDecryptedData(sensitiveKEK, .eciesEncryptionStandardX963SHA256AESGCM, encryptedSensitiveKeyData as CFData, &error) else {
+            throw error!.takeRetainedValue() as Error
+        }
+        // Data -> key
+        LogService.shared.debug("decryptedSensitiveKeyData: \((decryptedSensitiveKeyData as Data).toHexString())")
```
DmitryBespalov: needs to use query

Multisig/Features/Security/SensitiveEncryptedStore.swift
```diff
@@ -0,0 +1,183 @@
...
+        // find sensitiveKEK
+        let s = password != nil ? password : keychainStorage.retrievePasscode()
+        let sensitiveKEK = try keychainStorage.findKey(query: SQuery.ecPrivateKey(tag: KeychainStorage.sensitiveKekTag, password: s?.data(using: .utf8)))!
+
+        // find encrypted sensitive key
+        let encryptedSensitiveKeyData = try keychainStorage.retrieveEncryptedData(account: KeychainStorage.sensitiveEncryptedPrivateKeyTag)!
+
+        // decrypt encrypted sensitive key
+        var error: Unmanaged<CFError>?
+        guard let decryptedSensitiveKeyData = SecKeyCreateDecryptedData(sensitiveKEK, .eciesEncryptionStandardX963SHA256AESGCM, encryptedSensitiveKeyData as CFData, &error) else {
```
DmitryBespalov: needs to be extracted in separate function

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: What about `initializeKeyStore()`?

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: Using `SecRandomCopyByte()` needs some boilerplate. So I'd rather leave the method: 
```
    private func createRandomPassword() -> String? {
        var bytes: [UInt8] = .init(repeating: 0, count: 32)
        let result = SecRandomCopyBytes(kSecRandomDefault, bytes.count, &bytes)
        guard result == errSecSuccess else {
            return nil
        }
        return bytes.toHexString()
    }
```

Multisig/Features/Security/KeychainStorage.swift
biafra23: I think merging them makes most sense. Otherwise we need to preface every call to store with a delete. Delete takes the query and store takes the item. See https://github.com/safe-global/safe-ios/pull/2792/commits/eb280ab2d2be7c272baa16640993ad8e0bb34675

biafra23: Done. See: 26f28f89e705e549ad4a149ba09a9baf9d16760d

Multisig/Features/Security/KeychainStorage.swift
biafra23: I think it could be replaced with this:
```
        if status != errSecSuccess && status != errSecItemNotFound {
            throw convertStatusToError(status: status)
        }
```
Which is more readabla, nor?

biafra23: Or  with:
```
guard status == errSecSuccess || status == errSecItemNotFound else {
            throw convertStatusToError(status: status)
   }
```

Which is even more readable because it is not using a negation.

Multisig/Features/Security/SQuery.swift
biafra23: I have merged  SQuery and SItem to ItemSearchQuery. And added two methods to create the respective dictionaries

Multisig/Features/Security/KeychainStorage.swift
biafra23: Done

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: Isn't the service id enough? 
```
static let defaultService: String = "global.safe.sensitive.encrypted.data"
```
It is used for the password and the encrypted signer key. If we move it to `SensitiveEncryptedStore`, we should be fine, nor?

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: We only use the public KEK to encrypt the sensitive key private key. That only happens when re-creating the KEK (password change, access change, etc) immediately **after** creating it. If we store this public KEK we will only retrieve it, when it is outdated. Because we just created a new KEK.

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: I deleted `derivePasscode()` because we're not using it. When we derive the user given asscode, we might be able to just use the original implementation. Which takes care of the old salt.

biafra23: If we really need to be backwards compatible (I don't think we do), then we cannot change the number of iterations. And then we can use the original function from AuthenticationController.

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: I added `DataID` to the import and find() functions

Multisig/Features/Security/SensitiveEncryptedStore.swift
biafra23: initializeKeyStore() fails, when it tries to persist the password. 
```
        let status = SecItemAdd(try item.createAttributesForItem(), nil)
``` 
fails when the phone is locked.

biafra23: Doing this in the background is not a problem.

biafra23: I tried this by putting the following into MainTabbarViewController and locking the phone after starting the app: 

```
        let seconds = 15.0
        DispatchQueue.main.asyncAfter(deadline: .now() + seconds) {
            let encryptedStore: EncryptedStore = SensitiveEncryptedStore()
            let randomKey = Data(ethHex: "da18066dda40499e6ef67a392eda0fd90acf804448a765db9fa9b6e7dd15c322") as EthPrivateKey
            try! encryptedStore.initializeKeyStore()
            try! encryptedStore.import(id: DataID(id: "0xE86935943315293154c7AD63296b4e1adAc76364", protectionClass: .sensitive), ethPrivateKey: randomKey)
            let ethPrivateKey = try! encryptedStore.find(dataID: DataID(id: "0xE86935943315293154c7AD63296b4e1adAc76364", protectionClass: .sensitive), password: nil)
            print("Foo")
        }
```

Multisig/Features/Security/SensitiveEncryptedStore.swift
DmitryBespalov: OK, thank you for explanation, that is an important piece of information.

Multisig/Features/Security/ItemSearchQuery.swift
```diff
@@ -0,0 +1,153 @@
+//
+// Created by Dirk Jäckel on 17.11.22.
+// Copyright (c) 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import LocalAuthentication
+
+enum ItemSearchQuery {
```
DmitryBespalov: I think the name is still not good.

What about `KeychainItem`?

Multisig/Features/Security/ItemSearchQuery.swift
```diff
@@ -0,0 +1,153 @@
...
+    // Encrypted blob. Can be a password or a cec secret key
+    case generic(id: String, service: String, data: Data? = nil)
+    // Key stays in the Secure Enclave
+    case enclaveKey(tag: String = KeychainStorage.sensitiveKekTag, password: Data? = nil, access: SecAccessControlCreateFlags? = nil)
+    // Elliptic Curve Public Key
+    case ecPubKey(tag: String = KeychainStorage.sensitivePublicKeyTag, publicKey: SecKey? = nil)
+    // Elliptic Curve Key pair
+    case ecKeyPair
+
+    func createSearchQuery() -> NSDictionary {
```
DmitryBespalov: Swift's way of naming this would be just `searchQuery()`.

Multisig/Features/Security/ItemSearchQuery.swift
```diff
@@ -0,0 +1,153 @@
...
+                kSecReturnRef: true,
+            ]
+        case .ecKeyPair:
+            result = [:]
+        }
+
+        return result
+    }
+
+    func createAttributesForItem() throws -> NSDictionary {
```
DmitryBespalov: This name seems ambiguous, how about `creationAttributes()`?

Multisig/Features/Security/ItemSearchQuery.swift
```diff
@@ -0,0 +1,153 @@
...
+        )
+        else {
+            throw accessError!.takeRetainedValue() as Error
+        }
+        return access
+    }
+
+}
+
+func ==(left: ItemSearchQuery, right: ItemSearchQuery) -> Bool {
```
DmitryBespalov: the body of this function doesn't make sense to me - why is any ecKeyPair is the same?

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,170 @@
...
+//
+
+import Foundation
+import CommonCrypto
+import LocalAuthentication
+
+class KeychainStorage {
+
+    static let sensitivePublicKeyTag = "global.safe.sensitive.public.key"
+    static let sensitivePublicKekTag = "global.safe.sensitive.public.kek"
```
DmitryBespalov: these two variables are identical except one "k" character. Can we rename it so that they're more distinct?
 
For example: `sensitivePublicKEKTag` and `sensitivePublicKeyTag`

Multisig/Features/Security/KeychainStorage.swift
```diff
@@ -0,0 +1,170 @@
...
+    static let sensitiveEncryptedPrivateKeyTag = "global.safe.sensitive.private.key.as.encrypted.data"
+    static let derivedPasswordTag = "global.safe.password.as.data"
+    static let sensitiveKekTag = "global.safe.sensitive.KEK"
```
DmitryBespalov: is this a sensitive private kek tag? Maybe good to make it consistent with the public key variable

Multisig/Features/Security/ItemSearchQuery.swift
biafra23: I replaced its invocation with a simple switch statement wit a default branch.

DmitryBespalov:
after the changes, the code is still not in the place where KeychainStore is only responsible for keychain access, the Sensitive encrypted store is only responsible for encrypted/decrypted storage, and wallet-related and settings-related functionality would be in yet another layer above

Multisig/Features/Security/SensitiveStore.swift
```diff
@@ -0,0 +1,178 @@
...
+//  Multisig
+//
+//  Created by Dirk Jäckel on 30.11.22.
+//  Copyright © 2022 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+
+class SensitiveStore: EncryptedStore {
+    var initialized = false
```
DmitryBespalov: should it be `private var`?

Multisig/Features/Security/SensitiveStore.swift
```diff
@@ -0,0 +1,178 @@
...
+    private let store: KeychainItemStore
+    static let sensitivePublicKeyTag = "global.safe.sensitivePublicKeyTag"
+    static let sensitivePublicKEKTag = "global.safe.sensitivePublicKEKTag"
+    static let sensitivePrivateKEKTag = "global.safe.sensitivePrivateKEKTag"
+    static let sensitiveEncryptedPrivateKeyTag = "global.safe.sensitive.private.key.as.encrypted.data"
+    static let derivedPasswordTag = "global.safe.password.as.data"
```
DmitryBespalov, line 18: should all these tags be also private?

Multisig/Features/Security/SensitiveStore.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    init(_ store: KeychainItemStore) {
+        self.store = store
+    }
+
+    func isInitialized() -> Bool {
```
DmitryBespalov: instead of this, you could declare the `initialized` the following way:

```
private (set) var isInitialized: Bool
```

This way, you can set it only inside the class, but it is available to read for outside.

Multisig/Features/Security/SensitiveStore.swift
```diff
@@ -0,0 +1,178 @@
...
+        let kekItem = KeychainItem.enclaveKey(
+                tag: SensitiveStore.sensitivePrivateKEKTag,
+                password: passwordData,
+                access: [.applicationPassword]
+        )
+        let keyEncryptionKey: SecKey = try store.create(kekItem) as! SecKey
+
+        // 3. create key pair
+        let privateKeyItem = KeychainItem.ecKeyPair // is it a pair though? also, why no tag? because it is not stored directly.
+        let sensitiveKey: SecKey = try store.create(privateKeyItem) as! SecKey
```
DmitryBespalov, line 118: if this `try` fails, then the system is in the state where `password` is created. Do we need to care about this or it will be overridden on the next initialization?

Multisig/Features/Security/SensitiveStore.swift
DmitryBespalov: do we need this variable, at all? if the `initialize()` fails, it will throw error anyway

MultisigTests/Features/Security/KeychainItemStoreTests.swift
```diff
@@ -0,0 +1,214 @@
...
+
+    public override func setUp() {
+        super.setUp()
+        // Given
+        kciStore = KeychainItemStore(KeychainStore())
+    }
+
+    public override func tearDown() {
+        super.tearDown()
+        // Is it possible to always have a clean/empty keychain?
```
DmitryBespalov: no, it's not

MultisigTests/Features/Security/KeychainItemStoreTests.swift
```diff
@@ -0,0 +1,214 @@
+//
+// Created by Dirk Jäckel on 14.11.22.
+// Copyright (c) 2022 Gnosis Ltd. All rights reserved.
+//
+
+import XCTest
+@testable import Multisig
+import LocalAuthentication
+
+class KeychainItemStoreTests: XCTestCase {
```
DmitryBespalov, line 10: should we have a test when the same item is stored twice? i.e. updated?

Multisig/Features/Security/SensitiveStore.swift
biafra23: They are used when creating  KeychainItems.

Multisig/Features/Security/SensitiveStore.swift
biafra23: This var does not work for what I wanted to do. It was supposed to check whether the Sensitive Store has been initialized. Across app starts. To do that we would need to check whether there is, for example, a sensitive key already in the key chain.

Multisig/Features/Security/SensitiveStore.swift
DmitryBespalov: ok, but are these constants are only used within this class?

Multisig/Features/Security/SensitiveStore.swift
biafra23: IsInitialized() is false until the pub key of the sensitive key is stored. Last thing in this block. So if it crashes here, the password will be overwritten in the next attempt to initialize().

DmitryBespalov (approved):

Issue #2770, opened 2022-11-02
[Refactor] Make Connect WalletConnect & Ledger keys as flow


MouazAlzahabi:
We want to make add wallet connect and ledger keys using the flow structure, same as we have in other key types

DmitryBespalov:
@MouazAlzahabi I think this is done already?

MouazAlzahabi:
@DmitryBespalov Yes it's done already, I guess it has been returned by QA because of reported issues like #2773

PR #2799, opened 2022-11-22
Gh 2770/make connect wallet connect & ledger keys as flow

MouazAlzahabi:
Handles #2770


elgatovital (approved):
Tests are missing

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyFlow.swift
```diff
@@ -0,0 +1,118 @@
...
+    override func didIntro() {
+        selectDevice()
+    }
+
+    func selectDevice() {
+        let vc = flowFactory.selectDevice { [unowned self] deviceId, controller  in
+            bluetoothController = controller
+            addressPicker(deviceUUID: deviceId)
+        } onClose: {
+
```
DmitryBespalov, line 35: is it supposed to be empty on close?

Multisig/UI/Settings/OwnerKeyManagement/ImportOwnerKey/ImportKeyFlow.swift
```diff
@@ -19,49 +19,49 @@ import UIKit
...
         show(pickerVC)
     }
 
     override func doImport() -> Bool {
-        assert(privateKey != nil)
-        assert(keyName != nil)
-        let success = OwnerKeyController.importKey(privateKey!, name: keyName!, isDrivedFromSeedPhrase: keySource == .seed)
-        return success
+        assert(parameters?.name != nil)
+        return OwnerKeyController.importKey(parameters!.privateKey!, name: parameters!.name!, isDrivedFromSeedPhrase: parameters!.source == .seed)
```
DmitryBespalov: previous pattern of `let success = ... ; return success` allows to see the return value when debugging

DmitryBespalov: is there a way to see it with the debugger but without such a variable? 🤔

DmitryBespalov: There is! https://stackoverflow.com/questions/20902019/inspect-the-return-value-of-a-function-in-lldb

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
```diff
@@ -120,9 +105,9 @@ class AddKeyFlow: UIFlow {
     }
 
     func keyAdded() {
-        assert(keyInfo != nil)
-        assert(keyName != nil)
-        let vc = factory.keyAdded(address: keyInfo!.address, name: keyName!, keyType: keyType) { [unowned self] in
+        assert(keyParameters != nil)
+        assert(keyParameters.name != nil)
+        let vc = factory.keyAdded(address: keyParameters.address, name: keyParameters.name!, type: keyParameters.type) { [unowned self] in
```
DmitryBespalov: line too long

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
```diff
@@ -21,24 +21,19 @@ import UIKit
 /// 7. Create Passcode screen
 /// 8. (optional) override didCreatePasscode()
```
DmitryBespalov: comment needs better explanation of the steps

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
```diff
@@ -169,9 +154,9 @@ class AddKeyFlowFactory {
         enterNameVC.screenTitle = "Enter Key Name"
         enterNameVC.trackingEvent = parameters.keyNameTrackingEvent
         enterNameVC.placeholder = "Enter name"
-        enterNameVC.name = parameters.keyName
+        enterNameVC.name = parameters.name
         enterNameVC.address = parameters.address
-        enterNameVC.badgeName = parameters.badgeName
+        enterNameVC.badgeName = parameters.type.imageName
```
DmitryBespalov: implement `badgeName` on the `AddKeyParameters` as a get-only property

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyFlow.swift
DmitryBespalov: add comment `// not needed so that ledger connection can be re-established without closing the flow`

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyFlow.swift
```diff
@@ -0,0 +1,118 @@
...
+                .init(image: UIImage(named: "ico-onboarding-key"),
+                      title: "How secure is that?",
+                      body: "Your key will remain on your Ledger wallet. We do not store it in the app.")
+        ]
+        introVC.viewTrackingEvent = .ledgerOwnerOnboarding
+        introVC.navigationItem.title = "Connect Ledger Nano X"
+        return introVC
+    }
+
+    func selectDevice(completion: @escaping (UUID, BaseBluetoothController) -> Void, onClose: (() -> Void)?) -> SelectLedgerDeviceViewController {
```
DmitryBespalov: let's remove the `onClose` which is not used

Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/WalletConnectKeyFlow.swift
```diff
@@ -0,0 +1,99 @@
...
+            .init(image: UIImage(named: "ico-onboarding-import-key-3"),
+                    title: "Is my wallet supported?",
+                    body: "You wallet needs to support the WalletConnect protocol.")
+        ]
+        introVC.viewTrackingEvent = .connectOwnerOnboarding
+        introVC.navigationItem.title = "Connect Owner Key"
+        return introVC
+    }
+
+    func selectWallet(completion: @escaping (_ connection: WebConnection, _ wallet: WCAppRegistryEntry?) -> Void) -> SelectWalletViewController {
```
DmitryBespalov: let's switch the argument order in completion closure, and then we can pass the closure directly to the view controller.

DmitryBespalov: actually, we can remove this method, as it doesn't do anything useful

Multisig/UI/Settings/OwnerKeyManagement/LedgerOwnerKey/LedgerKeyFlow.swift
```diff
@@ -0,0 +1,118 @@
...
+class AddLedgerKeyParameters: AddKeyParameters {
+    var index: Int
+    var derivationPath: String
+    var deviceUUID: UUID
+
+    init(address: Address, keyName: String?, index: Int, derivationPath: String, deviceUUID: UUID) {
+        self.index = index
+        self.derivationPath = derivationPath
+        self.deviceUUID = deviceUUID
+        super.init(address: address, name: keyName, type: KeyType.ledgerNanoX, keyNameTrackingEvent: .ledgerEnterKeyName)
```
DmitryBespalov: Let's replace this tracking event with a `enter name` tracking event.

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
DmitryBespalov: not fixed after changes

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
DmitryBespalov: @MouazAlzahabi this comment also was resolved, but  I don't see the code change

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
DmitryBespalov: sorry, was checking wrong commit

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/GenerateKeyFlow.swift
```diff
@@ -35,37 +34,40 @@ class GenerateKeyFlow: AddKeyFlow {
...
 
-    override func didCreatePasscode() {
-        backup()
-    }
-
     func backup() {
-        assert(privateKey?.mnemonic != nil)
-        backupFlow = BackupFlow(mnemonic: privateKey!.mnemonic!) { [unowned self] _ in
+        assert(parameters?.privateKey?.mnemonic != nil)
+        backupFlow = BackupFlow(mnemonic: parameters!.privateKey!.mnemonic!) { [unowned self] _ in
```
DmitryBespalov: so many forced unwraps here and in other places

MouazAlzahabi: all are protected with the assert before right?, how do you suggest to handle it ?

DmitryBespalov: well... what options do we have and  can we try out and see what works better?
* guard let
* making struct fields forced unwrapped

DmitryBespalov: @MouazAlzahabi is it only in this place or in other functions and flows the same thing happens?

MouazAlzahabi: So as parameters are optional and some of its properties are optional then you might find force unwrapping in multiple places, but all are protected with check before

DmitryBespalov: what I'm trying to say that this problem is not only in this place but in multiple places, so the fix should be applied consistently. Please check out if the guard-let or forced unwrapping... basically so many ! in the code itself making it hard to reason / read, because then I question: why so many forced unwraps? is it intended?

MouazAlzahabi: @DmitryBespalov I did improvement for other places wdyt?

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/AddKeyFlow.swift
```diff
@@ -54,45 +52,35 @@ class AddKeyFlow: UIFlow {
...
     func enterName() {
-        assert(privateKey != nil)
-        let parameters = AddKeyParameters(
-            address: privateKey!.address,
-            keyName: nil,
-            badgeName: keyType.imageName
-        )
-        let nameVC = factory.enterName(parameters: parameters) { [unowned self] name in
-            keyName = name
+        assert(keyParameters != nil)
```
DmitryBespalov, line 60: I found more 😊  : replace assert with guard here and in other places in the file

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/GenerateKeyFlow.swift
```diff
@@ -35,37 +34,45 @@ class GenerateKeyFlow: AddKeyFlow {
...
+        backupFlow = BackupFlow(mnemonic: mnemonic) { [unowned self] _ in
             backupFlow = nil
             addKeyAsOwner()
         }
         push(flow: backupFlow)
     }
 
     func addKeyAsOwner() {
-        assert(keyInfo != nil)
+        assert(parameters?.address != nil)
```
DmitryBespalov: Found some more here

Multisig/UI/Settings/OwnerKeyManagement/GenerateOwnerKey/GenerateKeyFlow.swift
```diff
@@ -98,15 +105,16 @@ class GenerateKeyFlow: AddKeyFlow {
     }
 
     func addOwner() {
-        addOwnerFlow = AddOwnerFlow(newOwner: keyInfo!.address, safe: safe!) { [unowned self] skippedTxDetails in
+        assert(parameters != nil)
```
DmitryBespalov: and here and in other function below

DmitryBespalov (approved):
Look, you've deleted more lines than you're adding! Great!

Issue #2797, opened 2022-11-22
the "Update now" screen should appear on each app call


DmitryBespalov:
As part of preparation to maintenance update for December 1, we need to test if the "deprecated soon" banner works in iOS 3.18.0 and Android 3.5.0.

What to test:
* Enable banner for staging builds in Firebase
* Wait 15 minutes
* Open app, go to some screens
* Close and wait for 15 minutes
* Open app again
Banner should show

Original documentation:
https://github.com/safe-global/safe-ios/issues/1084

liliya-soroka:
Android:
1. nothing happens when "Update now" is clicked

ios:
1. doesn't work for me - need to check together with @MouazAlzahabi

DmitryBespalov:
Android: Update now" works only for signed version (not for staging version)

liliya-soroka:
the "Update now" screen should appear on each app call ( call from the background is included), not only the cold app start
ios app version 3.19.0 (1446)
cc @MouazAlzahabi

PR #2800, opened 2022-11-23
Fix check update status

MouazAlzahabi:
Handles #2797


elgatovital (approved):

liliya-soroka:
the "Update now" screen should appear on each app call ( call from the background is included), not only the cold app start
ios app version 3.19.0 (1446)
cc @MouazAlzahabi

PR #2801, opened 2022-11-24
Resolve confilcts

elgatovital:
merged main into release/3.18.1


MouazAlzahabi:
I believe we can do that much simpler way, no need to merge all changes into that pr because that way there will be no point of having separated pr for the release

elgatovital:
> I believe we can do that much simpler way, no need to merge all changes into that pr because that way there will be no point of having separated pr for the release

This means you can push directly to release branches omitting branch protection rules

Issue #2796, opened 2022-11-21
Migrate to collectibles v2


DmitryBespalov:
Reason: v1 is going to be deprecated and will limit number of returned collectibles.

DmitryBespalov:
duplicate

PR #2803, opened 2022-11-28
gh-2796 Paginated collectibles list

elgatovital:
Handles #2796

- Use new endpoint for getting collectibles data
- Paginated collectibles list


Multisig/UI/Assets/CollectiblesViewController/FlatCollectiblesListViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+        self.items = collectibleListItems
+    }
+
+    mutating func append(from list: Self) {
+        self.next = list.next
+        add(list.items)
+    }
+
+    mutating func add(_ models: [CollectibleListItem] = []) {
```
DmitryBespalov, line 37: What are the requirements for this method? Also, what does it guarantee to do?

Multisig/UI/Assets/CollectiblesViewController/CollectibleHeaderTableViewCell.xib
```diff
@@ -0,0 +1,61 @@
...
+        <deployment identifier="iOS"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="20020"/>
+        <capability name="Named colors" minToolsVersion="9.0"/>
+        <capability name="Safe area layout guides" minToolsVersion="9.0"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
+    </dependencies>
+    <objects>
+        <placeholder placeholderIdentifier="IBFilesOwner" id="-1" userLabel="File's Owner"/>
+        <placeholder placeholderIdentifier="IBFirstResponder" id="-2" customClass="UIResponder"/>
+        <tableViewCell contentMode="scaleToFill" selectionStyle="default" indentationWidth="10" rowHeight="51" id="KGk-i7-Jjw" customClass="CollectibleHeaderTableViewCell" customModule="Multisig" customModuleProvider="target">
```
DmitryBespalov, line 14: Does it gets selected when user taps on it? Plase make the "selection" to "none" for this table view cell, otherwise if user click on it, it could be highlihgted.

Multisig/UI/Transaction/TransactionListViewController/Cells/LoadingFooterView.xib
```diff
@@ -22,7 +22,7 @@
                             <rect key="frame" x="197" y="20" width="20" height="22"/>
                         </activityIndicatorView>
                     </subviews>
-                    <color key="backgroundColor" name="backgroundPrimary"/>
+                    <color key="backgroundColor" name="backgroundContent"/>
```
DmitryBespalov, line 25: Will this affect some other tables in other screens? (similar comment to other footers). Why is the color changed? i.e. was it wrong?

Multisig/UI/Assets/CollectiblesViewController/FlatCollectiblesListViewModel.swift
elgatovital: same as in FlatTransactionsListViewModel. It adds another page of items to the list.

Multisig/UI/Assets/CollectiblesViewController/CollectibleHeaderTableViewCell.xib
elgatovital: it is already being set to none in code. see CollectiblesViewController:230

Multisig/UI/Transaction/TransactionListViewController/Cells/LoadingFooterView.xib
elgatovital: yes, content background (lists) is different from primary background.

Multisig/UI/Assets/CollectiblesViewController/FlatCollectiblesListViewModel.swift
DmitryBespalov: I understood that, but why is all the guard/let logic?

DmitryBespalov: does it expect the models content be in some configurations?

Multisig/UI/Assets/CollectiblesViewController/CollectibleHeaderTableViewCell.xib
DmitryBespalov: Thanks for checking this!

Multisig/UI/Transaction/TransactionListViewController/Cells/LoadingFooterView.xib
DmitryBespalov: ok, so before it was showing a visible horizontal band?

elgatovital: exactly

Multisig/UI/Assets/CollectiblesViewController/FlatCollectiblesListViewModel.swift
elgatovital: header separators are inserted manually. If next page starts with items from group that started on the previous page, we have to make sure that first header is removed.

DmitryBespalov: Can you please add this to a method comment? Otherwise it is not clear what is the intent of the code, i.e. if it works as expected or not

On line 41 it uses "models[1]", is that correct? Should not it be "models[0]"?

Multisig/UI/Assets/CollectiblesViewController/FlatCollectiblesListViewModel.swift
```diff
@@ -41,6 +41,7 @@ struct FlatCollectiblesListViewModel {
         guard case let .collectible(collectibleItem) = models[1], let currentPageFirstNft = collectibleItem.collectible.address else {
             return
         }
+        // drop first header if page starts with collectibles that are from the same group as last collectibles of the previous page
```
DmitryBespalov: Vitaly, come on, this just tells what the code does, but not why it is done this way.

The better comment would be:
```
// The 'header separator' items are inserted manually after downloading the collectibles. This results in duplicated headers if the same section is split between different pages of response. We have to remove such headers.
```

Multisig/UI/Assets/CollectiblesViewController/FlatCollectiblesListViewModel.swift
elgatovital: added comments, everything is correct as is. models[0] always contains header item at this point.

DmitryBespalov: maybe I would add explanation: 

```
// The models[0] always contains a header because it is a very first section on the page, and the header was added manually before.
```

DmitryBespalov (approved):
thankyouverymuch!

Issue #2695, opened 2022-09-28
[SAFE] Handle delegating to the safe itself


DmitryBespalov:
# Context
The delegation to self does not work, i.e. transaction would fail. We need to handle this case now.

# 1. In the "Custom address or ENS" screen, change text

from:

```
You can select yourself or any other person to be a delegate.
```

to:

```
You can select yourself or any other person to be a delegate. Your Safe has the full voting power even after delegation.
```

# 2. Show error when the safe itself is selected as delegate
When user enters the safe address as a delegate, show the error message under the address field and disable "Select & Continue" button.

```
The Safe address can't be a delegate of itself. Please choose another address or select a guardian.

Your Safe always has the full voting power even after delegating to another address.
```

PR #2804, opened 2022-11-30
Fix issue

MouazAlzahabi:
Handles #2695


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -353,6 +353,15 @@ enum GSError {
         let code = 1112
         let loggable = false
     }
+
+    struct DelagateToSameSafe: DetailedLocalizedError {
```
elgatovital, line 357: where is this error used?

elgatovital (approved):

DmitryBespalov (approved):
👍

PR #2805, opened 2022-12-05
Fix assert

elgatovital:


DmitryBespalov (dismissed):

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
```diff
@@ -112,10 +112,11 @@ extension SECP256K1.UnmarshaledSignature {
             // recover V by deducting (chainId * 2 + 35) according to EIP-155.
             let vRecovered = vInt - (chainIdInt * 2 + 35)
             v = try! UInt8(vRecovered % 256)
+
+            assert(v == 0 || v == 1)
         } else {
             v = data[64]
         }
-        assert(v == 0 || v == 1)
```
DmitryBespalov, line 118: wait, what? 

🤔 

At this point, it has to be 0, 1, or 27 or 28

https://eips.ethereum.org/EIPS/eip-155

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
```diff
@@ -112,10 +112,11 @@ extension SECP256K1.UnmarshaledSignature {
             // recover V by deducting (chainId * 2 + 35) according to EIP-155.
             let vRecovered = vInt - (chainIdInt * 2 + 35)
```
DmitryBespalov, line 113: Another comment after I was digging into EIP-155 and chainlist.org (https://github.com/DefiLlama/chainlist):

the chain Id must actually be `Int64`, not `UInt64`, if to conform to the chainlist repository use.

DmitryBespalov: But perhaps it's a mistake in the chainlist repository

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
elgatovital: we hash 9 values according to eip-155, so v can be 0 or 1 at this point

DmitryBespalov: then why the assert fails?

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
DmitryBespalov: actually, they moved from Kotlin to JS and reduced number of chain ids, so it's not clear now if it's Int64 or UInt64. Then I'd keep this as is.

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
DmitryBespalov: Correction: the `v` can be: `0, 1, 2, 3, 27, 28, 29, 30`

DmitryBespalov: ... and I'm wrong again. 

If the data is not overflow, the v still can be chainId * 2 + 35 for non-legacy transactions. For legacy transactions, it can be 27, 28, 29, 30. For raw ends it is 0, 1, 2, 3

DmitryBespalov: To recap:
- ECDSA `v` can be: `0, 1, 2, 3`
- Legacy ethereum (pre-eip-155) adds `27` to `v` - the same as bitcoin does ([see internal thread for the rabbit hole](https://5afe.slack.com/archives/C03BHKS5K6D/p1669128877489559?thread_ts=1664985580.849769&cid=C03BHKS5K6D) or [stack exchange](https://bitcoin.stackexchange.com/questions/5082/why-do-compact-signatures-need-to-start-with-a-byte-27-35)), so `v` can be in range `27...30`
- EIP-155 transaction use `{0,1} + chain_id * 2 + 35` to compute `v`. This overflows 1 byte integer width when `chain_id > 109`. Therefore, the `v` can be in the range from `31 ... `


I believe that this whole method should be rewritten

elgatovital: wrong place place for this assert

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
```diff
@@ -101,21 +101,35 @@ extension SECP256K1.UnmarshaledSignature {
...
         } else {
-            v = data[64]
+            vBytes = [UInt8]([data[64]])
+            let vInt = UInt8(vBytes)
+            if isLegacyTx {
+                // Legacy ethereum (pre-eip-155) adds 27 to v
+                v = vInt - 27
+            } else {
+                // v still can be chainId * 2 + 35 for non-legacy transactions
+                if vInt >= 37 {
```
DmitryBespalov: this is incorrect, because chainId can be = 0 and so the eip-155 v value  will be >= 35

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
```diff
@@ -101,21 +101,35 @@ extension SECP256K1.UnmarshaledSignature {
...
+                    let vRecovered = UInt64(vBytes) - (chainIdInt * 2 + 35)
+                    v = try! UInt8(vRecovered % 256)
+                } else {
+                    v = vInt
```
DmitryBespalov, line 127: According to the current code, this branch will execute for values of vInt `>= 0 || <37`, so the assert at line 132 can fail.

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
elgatovital: this is highly unlikely though

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
elgatovital: I would say it is desired for the assert to fail in case of wrong values.

DmitryBespalov: my point is that the values are not wrong

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
DmitryBespalov: that is incorrect

Multisig/UI/Settings/OwnerKeyManagement/KeystoneOwnerKey/KeystoneSignFlow.swift
```diff
@@ -101,21 +101,35 @@ extension SECP256K1.UnmarshaledSignature {
...
+                }
+            }
         }
-        assert(v == 0 || v == 1)
+        
+        assert(v == 0 || v == 1 || v == 27 || v == 28)
```
DmitryBespalov: the correct values for `v` are `0, 1, 2, 3, 27, 28, 29, 30`

https://bitcoin.stackexchange.com/a/112489

DmitryBespalov: am I wrong about `2, 3, 29, 30`?

DmitryBespalov: According to the source code here

https://github.com/Boilertalk/secp256k1/blob/d5407179912e8c1f825a212a474aaa86b10f1352/src/ecdsa_impl.h#L292

the `recid` can have values of `0, 1, 2, and 2^1=3`, therefore v can have values `27, 28, 29, 30`. qed :D

biafra23 (approved):
LGTM 🚀

Issue #2847, opened 2023-01-16
[KeySec1] Delete private key


DmitryBespalov:
Use Wallet to delete a private key

liliya-soroka:
reported issues:
https://github.com/safe-global/safe-ios/issues/2877
https://github.com/safe-global/safe-ios/issues/2873

liliya-soroka:
Verified
app version 3.19.0(1568)

PR #2864, opened 2023-01-18
gh-2847 Delete private key

elgatovital:
Handles #2847

### DIFF
--- a/Multisig/Logic/Models/PrivateKey/PrivateKey.swift
+++ b/Multisig/Logic/Models/PrivateKey/PrivateKey.swift
@@ -113,12 +113,26 @@ extension PrivateKey {
         }
     }
 
-    static func remove(id: KeyID) throws {
-        do {
-            try App.shared.keychainService.removeData(forKey: id)
-        } catch {
-            throw GSError.KeychainError(reason: error.localizedDescription)
+    //@Deprecated: legacy code
+    //TODO: extract legacy code
+    static func remove(id: KeyID, address: Address) throws {
+        if App.configuration.toggles.securityCenter {
+            //TODO: rewrite as App.securityCenter
+            //TODO: make invocation async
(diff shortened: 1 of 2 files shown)


Multisig/Logic/Models/PrivateKey/PrivateKey.swift
```diff
@@ -113,14 +113,28 @@ extension PrivateKey {
...
-            try App.shared.keychainService.removeData(forKey: id)
-        } catch {
-            throw GSError.KeychainError(reason: error.localizedDescription)
+    //@Deprecated: legacy code
+    //TODO: extract legacy code
+    static func remove(id: KeyID, address: Address) throws {
+        if App.configuration.toggles.securityCenter {
+            //TODO: rewrite as App.securityCenter
+            //TODO: make invocation async
+            App.shared.securityCenter.remove(address: address) { result in
```
DmitryBespalov, line 122: why is address used instead of keyID?

elgatovital: ProtectedKeystore uses address for removal

DmitryBespalov: ok for now. From my perspective, the store should use whatever ID it receives, I could store the sensitive information for something not address related...

MouazAlzahabi: I would also be for using ID instead of address

Multisig/Logic/Models/PrivateKey/PrivateKey.swift
```diff
@@ -113,14 +113,28 @@ extension PrivateKey {
...
+                try! result.get()
```
DmitryBespalov, line 123: another todo: handle errors

DmitryBespalov (approved):

Issue #2831, opened 2022-12-20
[KeySec1] Delegate Keys - Creating, removing


DmitryBespalov:
- [ ] Creating a delegate key when adding new owner key
- [ ] Removing a delegate key from key details
- [ ] Removing a delegate key when the owner key is removed

liliya-soroka:
blocked by https://github.com/safe-global/safe-ios/issues/2890

liliya-soroka:
Checked:
Imported key
Ledger Nano X
Keystome device 

In progress

liliya-soroka:
Verified
app version 3.20.0(1713)

PR #2871, opened 2023-01-23
gh-2831 Adjust delegate keys lookup and deletion

elgatovital:
Handles #2831

- Adjust delegate keys lookup and deletion

### DIFF
--- a/Multisig/Features/Security/SecurityCenter.swift
+++ b/Multisig/Features/Security/SecurityCenter.swift
@@ -74,7 +74,7 @@ class SecurityCenter {
     }
 
     // import data potentially overriding existing value
-    func `import`(id: DataID, ethPrivateKey: EthPrivateKey, protectionClass: ProtectionClass = .sensitive, completion: @escaping (Result<Bool?, Error>) -> ()) {
+    func `import`(id: DataID, ethPrivateKey: EthPrivateKey, protectionClass: ProtectionClass = .sensitive, completion: @escaping (Result<Bool, Error>) -> ()) {
         switch(protectionClass) {
         case .sensitive:
             perfomSecuredAccess { [unowned self] result in
--- a/Multisig/Features/Security/SecurityCenter.swift
+++ b/Multisig/Features/Security/SecurityCenter.swift
@@ -104,16 +104,12 @@ class SecurityCenter {
         switch(protectionClass) {
         case .sensitive:
             perfomSecuredAccess { [unowned self] result in
-                switch result {
-                case .success:
-                    do {
-                        try sensitiveStore.delete(address: address)
-                        completion(.success(true))
-                    } catch let error {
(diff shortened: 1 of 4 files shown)


Multisig/Features/Security/SecurityCenter.swift
```diff
@@ -74,66 +74,106 @@ class SecurityCenter {
...
-    func remove(address: Address, completion: @escaping (Result<Bool?, Error>) -> ()) {
-        perfomSecuredAccess { [unowned self] result in
-            switch result {
-            case .success:
-                do {
-                    try sensitiveStore.delete(address: address)
-                    completion(.success(true))
-                } catch let error {
-                    completion(.failure(GSError.KeychainError(reason: error.localizedDescription)))
+    func remove(address: Address, protectionClass: ProtectionClass = .sensitive, completion: @escaping (Result<Bool?, Error>) -> ()) {
```
DmitryBespalov: unrelated comment: 

```(Result<Bool?, Error>) -> ()``` means it returns `Result`. Result is an enum with 2 cases: `success` and `failure`. In this declaration, for success case, it returns boolean that can be nil. I understand when the `success(true)` means success, but what do `success(false)` or `success(nil)` mean?

If `success` case's Bool is always true, then it can be replaced with `Void` type (which is equivalent to empty tuple `()`). Then, to indicate success, you'd do `success(())`. I do agree that is less readable.

Another alternative is to return `Error?` and if it's nil, that would mean success.

File: Multisig/Features/Security/SecurityCenter.swift [hunk 2]
```diff
@@ -74,66 +74,106 @@ class SecurityCenter {
...
+                case .success:
+                    do {
+                        try sensitiveStore.delete(address: address)
+                        completion(.success(true))
+                    } catch let error {
+                        completion(.failure(GSError.KeychainError(reason: error.localizedDescription)))
+                    }
+                case .failure(let error):
+                    completion(.failure(error))
                 }
```
DmitryBespalov: Could we simplify it this way?
```suggestion
               do {
                    _ = try result.get()
                    try sensitiveStore.delete(address: address)
                    completion(.success(true))
                } catch {
                    completion(.failure(GSError.KeychainError(reason: error.localizedDescription)))
                }
```

DmitryBespalov:
App lock/unlock needs the changes from this PR

Multisig/Features/Security/SecurityCenter.swift
elgatovital: changed to non nullable success type for now. It could be interpreted as following for example: success(true) operation performed successfully, success(false) operation canceled by user without any exception.

DmitryBespalov: please document it, because the code doesn't tell the intention

DmitryBespalov (approved):

Issue #2855, opened 2023-01-16
[KeySec1] Security Settings Screen - UI - Stage 2


DmitryBespalov:
# Security Settings Screen - UI

<img src="https://user-images.githubusercontent.com/284056/208631946-aba6ae2b-892e-46c8-bb53-fb3e815ba9e9.png" width=300 /> <img src="https://user-images.githubusercontent.com/284056/208631947-8e5e6a2a-1113-477a-88f3-1bd8ab893005.png" width=300 />

## What to do
This is only about building UI. Handling the functionality is done in other tickets.

### Text Descriptions depending on the Lock Method

|  | Unlocking the app | Making transactions |
| --- | --- | --- |
| Template | Only &lt;TEXT&gt; will be required to unlock the app. | &lt;TEXT&gt; will be used for making transactions. |
| Lock method: |  |  |
| Passcode | Passcode | Passcode |
| Face ID | Face ID | Face ID |
| Touch ID | Touch ID | Touch ID |
| Device Passcode | Device Passcode | Device Passcode |
| Passcode & Face ID | Face ID | Both Passcode & Face ID |
| Passcode & Touch ID | Touch ID | Both Passcode & Touch ID |
| Passcode & Device Passcode | Device Passcode | Both Passcode & Device Passcode ID |

### Text Descriptions depending on the Biometry type and availability

|  | Enable security lock |
| --- | --- |
| Template | Require passcode &lt;TEXT&gt; for unlocking the app, making transactions and using signer accounts |
| Biometry Setting |  |
| Available; Face ID | “or Face ID “ |
| Available; Touch ID | “or Touch ID “ |
| Not available | “” (empty) |

## Resources
- [Figma](https://www.figma.com/file/lywEf6byZNrQbFdY7x44f3/iOS-Security-Settings?node-id=0%3A1&t=PmHGFKKgQF2DAGRz-11)

liliya-soroka:
Verified
app version 3.19.0 (1561)

PR #2882, opened 2023-01-25
Security settings UI

MouazAlzahabi:
Handles #2855 

- Known issue: 
Change lock method is not working because ProtectedKeyStore.changePassword method not working properly

### DIFF
--- a/Multisig/UI/Settings/Passcode/SecuritySettingsViewController.swift
+++ b/Multisig/UI/Settings/Passcode/SecuritySettingsViewController.swift
@@ -217,23 +217,13 @@ class SecuritySettingsViewController: UITableViewController {
 
     private func toggleBiometricsSecurityLock(newLockMethod: LockMethod) {
         guard newLockMethod != App.shared.securityCenter.lockMethod else { return }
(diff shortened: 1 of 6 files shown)


Multisig/UI/Settings/Passcode/SecuritySettingsViewController.swift
```diff
@@ -225,6 +210,33 @@ class PasscodeSettingsViewController: UITableViewController {
...
+
+    private func toggleBiometricsSecurityLock(newLockMethod: LockMethod) {
+        guard newLockMethod != App.shared.securityCenter.lockMethod else { return }
+        withPasscodeAuthentication(for: "Login with biometrics") { [unowned self] success, nav, finish in
+            let completion = { [unowned self] in
+                finish()
+                reloadData()
+            }
+
+            if success {
```
DmitryBespalov: I thought that this call is not needed for the new system, because the securityCenter will request passcode by itself, so I think this can be deleted.

DmitryBespalov (commented):
1 comment to resolve

DmitryBespalov (approved):

Issue #2825, opened 2022-12-20
[KeySec1] Security Settings Screen - Toggles


DmitryBespalov:
Changing any toggle requires user authentication with the present authentication method. Then, the app changes the configuration of the appropriate security item (sensitive KEK or unlock challenge).

| Unlocking the app | Making transactions | Action |
| --- | --- | --- |
| ✅ | ✅ | Starting state after the security lock is enabled. |
| ✅ | ❌ | Change the settings for private keys only, i.e. configure the Sensitive KEK as if the security lock is disabled |
| ❌ | ✅ | Change the settings for app unlock: do not use app unlock, i.e. as if the security lock is disabled |
| ❌ | ❌ | Disable the security lock all together |

liliya-soroka:
approved without Passcode&FaceID method

PR #2893, opened 2023-01-31
add authentication when toggle security options

MouazAlzahabi:
Handles #2825

### DIFF
--- a/Multisig/Features/Security/SecurityCenter.swift
+++ b/Multisig/Features/Security/SecurityCenter.swift
@@ -183,7 +183,6 @@ class SecurityCenter {
         if toggleWillBeOff {
             requestPasswordV2(for: [protectionClass]) { [unowned self] plaintextPasscode in
                 let currentDerivedPassword = plaintextPasscode.map { derivedKey(from: $0) }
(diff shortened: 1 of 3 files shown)


Multisig/Features/Security/SecurityCenter.swift
```diff
@@ -182,19 +182,26 @@ class SecurityCenter {
         let toggleWillBeOff: Bool = enabledOptions.contains(passcodeOption)
         if toggleWillBeOff {
             requestPasswordV2(for: [protectionClass]) { [unowned self] plaintextPasscode in
+                let currentDerivedPassword = plaintextPasscode.map { derivedKey(from: $0) }
+                try protectedKeyStore.authenticate(password: currentDerivedPassword)
```
DmitryBespalov: I thought that authentication here is not needed, because the other call will fail

DmitryBespalov (approved):

Multisig/Features/Security/SecurityCenter.swift
MouazAlzahabi: Actually I added it because I thought change password doesn't check for correct passcode, then I discovered that it's not working properly, I will remove it

PR #2894, opened 2023-02-01
Fix disable usage options

MouazAlzahabi:
Handles #2825

### DIFF
--- a/Multisig/Features/Security/KeychainItem.swift
+++ b/Multisig/Features/Security/KeychainItem.swift
@@ -58,7 +58,6 @@ enum KeychainItem {
             ]
 
         case let .enclaveKey(tag, service, password, _):
-
             result = [
                 kSecClass: kSecClassKey,
                 kSecAttrKeyClass: kSecAttrKeyClassPrivate,
@@ -69,6 +68,9 @@ enum KeychainItem {
 
             if let context = LAContext(password: password) {
                 result[kSecUseAuthenticationContext] = context
+            }  else {
+                // Without this line, keychain item can be accessible with worng passcode if it has been initated without passcode
+                return [:]
             }
         case let .ecPubKey(tag, service, _):
             result = [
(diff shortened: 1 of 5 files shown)


Multisig/Features/Security/KeychainItem.swift
```diff
@@ -110,6 +110,9 @@ enum KeychainItem {
             ]
             if let context = LAContext(password: password) {
                 privateKeyAttrs[kSecUseAuthenticationContext] = context
+            } else {
+                // In case password is worng
```
elgatovital: I would also add short explanation here regarding case when device password is not set (failure to create LAContext)

Multisig/Features/Security/SecurityCenter.swift
```diff
@@ -148,9 +148,20 @@ class SecurityCenter {
 
             AppSettings.securityLockMethod = newMethod
 
-            try changeStoreSettings(currentPlaintextPassword: oldPasscode, newPlaintextPassword: changedPasscode, store: sensitiveStore)
-            try changeStoreSettings(currentPlaintextPassword: oldPasscode, newPlaintextPassword: changedPasscode, store: dataStore)
-            NotificationCenter.default.post(name: .securitySettingsChanged, object: self)
+            // Change settings should be done only for the enabled store
+            // Change settings for disabled store will through an exception
```
elgatovital: typo in comment: will throw an exception

Multisig/Features/Security/KeychainItem.swift
```diff
@@ -68,9 +68,6 @@ enum KeychainItem {
 
             if let context = LAContext(password: password) {
                 result[kSecUseAuthenticationContext] = context
-            }  else {
```
elgatovital: what if keychain item has been initialised when device password was not set? Wouldn't this edge-case lead to problems?

DmitryBespalov (approved):

Multisig/Features/Security/KeychainItem.swift
MouazAlzahabi: This we will do in next step

Issue #2900, opened 2023-02-02
[iOS] FaceID check appears twice when switching from FaceID to Passcode


liliya-soroka:
**Bug description**
FaceID check appears twice when FaceID method is switched to Passcode
**Steps To Reproduce**
Case 1:
1. enable the feature toggle and enable security with the FaceID method
2. open security settings and change FaceID to Passcode
3. enter valid new passcode twice
Current result: FaceID check appears twice
**Expected Result**
Make sure that the FaceID check appears only once

Case 2:
1. enable the feature toggle and enable security with the FaceID method
2. open security settings and change FaceID to Passcode
3. enter valid new passcode twice
4. but invalid faceID
5. the user is asked to enter device passcode
6. enter correct device passcode

Current result: FaceID check appears again
**Expected Result**
Make sure that FaceID check appears only once

**Device & App version**
 - Device/OS: iPhone 11/15.0
 - App version: 3.19.0(1561)
 
**Environment**
staging
**Additional info**
(Optional)

biafra23:
~~The reason seems to be that we need to unlock two secure enclave keys. Maybe we can use one secure enclave key for both data and sensitive store.~~

biafra23:
~~I found a solution. We can reuse the first biometrics authentication LAContext for the second KEK.~~

biafra23:
Looks like with the current code the bug described cannot be reproduced. 
1. But unlocking the app when restarting sometimes asks three times for the TouchId.
2. When switching to faceId & passcode, the app asks for the passcode on startup (wrong, should ask for faceId only) and refuses to accept the correct passcode (wrong)
3. When switching to faceId the imported keys are lost
4. When usinf faceid app start takes forever on ipados 16.2
5. App asks for pasccode on app start even if app unlock is disabled

biafra23:
The bug described only happens after fresh install. When switching between touchId and passcode later, touchId is only requested once. 

After a fresh install the `ProtectedKeyStore.changePassword()` method is called twice. That's why the touchId is requested twice.

liliya-soroka:
verified
app version 3.20.0(1719)

PR #2914, opened 2023-02-15
gh-2900 Execute onPasscodeEnter()

biafra23:
Handles 

- Execute onPasscodeEnter to disable or enable the security lock
- This allows to enable/disable the lock
- It allows to switch between FaceId and password

### DIFF
--- a/Multisig/UI/Settings/Passcode/EnterPasscodeViewController.swift
+++ b/Multisig/UI/Settings/Passcode/EnterPasscodeViewController.swift
@@ -10,6 +10,8 @@ import Foundation
 import UIKit
 
 class EnterPasscodeViewController: PasscodeViewController {
+
+    // Deprecated with the new security center
     var passcodeCompletion: (_ success: Bool, _ reset: Bool, _ passcode: String?) -> Void = { _, _, _ in }
 
     var onPasscodeEnter: (_ password: String?) throws -> Void = { _ in }
@@ -72,7 +74,7 @@ class EnterPasscodeViewController: PasscodeViewController {
             do {
                 try onPasscodeEnter(text)
             } catch {
-                LogService.shared.error(error.localizedDescription)
+                onError(error)
             }
         } else {
             wrongAttemptsCount += 1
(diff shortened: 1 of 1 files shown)


Multisig/UI/Settings/Passcode/EnterPasscodeViewController.swift
```diff
@@ -69,6 +69,11 @@ class EnterPasscodeViewController: PasscodeViewController {
 
         if isCorrect {
             passcodeCompletion(true, false, text)
+            do {
```
MouazAlzahabi, line 74: I think this is wrong, look for didEnterEnoughSymbolsV2 method below, that method should be called instead of this in case securitycenter feature is enabled, so the question is why this method is being called instead of didEnterEnoughSymbolsV2

elgatovital (approved):

MouazAlzahabi (changes requested):

Multisig/UI/Settings/Passcode/EnterPasscodeViewController.swift
elgatovital: we have removed didEnterEnoughSymbols couple of pr's ago and doing the security version check in didEnoughSymbols now. Having 2 functions for didEnterEnoughSymbols seemed redundant.

biafra23: I could not find `didEnterEnoughSymbolsV2` in this class or anywhere in our code.

Multisig/UI/Settings/Passcode/EnterPasscodeViewController.swift
```diff
@@ -69,6 +69,11 @@ class EnterPasscodeViewController: PasscodeViewController {
...
+                try onPasscodeEnter(text)
+            } catch {
+                LogService.shared.error(error.localizedDescription)
```
MouazAlzahabi: here I think we need to call onError callback instead

biafra23: See latest commit.

MouazAlzahabi (approved):

Issue #2830, opened 2022-12-20
[KeySec1] Delegate Keys - Signing with the key, Push notifications


DmitryBespalov:
- [ ] Delegate key is used for signing (for de-registering)
- [ ] Delegate key is used for signing for push notification registration

liliya-soroka:
Verified
app version 3.20.0(1713)

PR #2924, opened 2023-02-21
gh-2830 Register delegate keys for push notification

elgatovital:
Handles #2830

- Retrieve and register delegate keys for push notification using new security v2 mechanism

### DIFF
--- a/Multisig/App/RemoteNotificationHandler.swift
+++ b/Multisig/App/RemoteNotificationHandler.swift
@@ -54,7 +54,9 @@ class RemoteNotificationHandler {
         logDebug("Push token updated")
         self.token = token
         if authorizationStatus != nil {
-            registerAll()
+            Task {
+                await registerAll()
+            }
         }
     }
 
@@ -63,7 +65,9 @@ class RemoteNotificationHandler {
         if authorizationStatus == nil {
             requestUserPermissionAndRegister()
         } else {
-            registerAll()
+            Task {
+                await registerAll()
+            }
         }
     }
 
@@ -74,7 +78,9 @@ class RemoteNotificationHandler {
     /// For add / remove signing key
     func signingKeyUpdated() {
         logDebug("Signing key updated")
-        registerAll()
+        Task {
+            await registerAll()
+        }
     }
 
     func received(notification userInfo: [AnyHashable: Any]) {
@@ -105,7 +111,9 @@ class RemoteNotificationHandler {
                     self.setStatus(settings.authorizationStatus)
 
                     if settings.authorizationStatus.hasPermission {
-                        self.registerAll()
+                        Task {
+                            await self.registerAll()
+                        }
                     }
                 }
             }
@@ -138,7 +146,9 @@ class RemoteNotificationHandler {
 
                 // At the time when permission granted, the token will be
                 // already set, so we need to register all stored safes
-                self.registerAll()
+                Task {
+                    await self.registerAll()
+                }
             }
             self.updateAuthorizationStatus()
         }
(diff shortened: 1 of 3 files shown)


Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -393,6 +393,34 @@ extension KeyInfo {
         if let key = (try? privateKey()) { return key }
         return try? delegatePrivateKey()
     }
+
+    func pushNotificationSigningKey(completion: @escaping (Result<PrivateKey?, Error>) -> ()) {
```
biafra23: Looks like we're not using this anywhere except int the async wrapper version. Can we get rid of this by rewriting the async method or at least annotate it with 
```
@available(*, deprecated, renamed: "pushNotificationSigningKey()")
```
So it is not used anywhere? Or we could make it private?

Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -393,6 +393,34 @@ extension KeyInfo {
...
+                } else {
+                    completion(.success(privateKey))
+                }
+            case .failure(let error):
+                completion(.failure(error))
+            }
+        }
+    }
+
+    func pushNotificationSigningKey() async throws -> PrivateKey? {
```
MouazAlzahabi, line 397: So this method is like a wrapper for the previous one, I don't see the why we need it actually, I would just keep the previous one and use it directly

Multisig/App/RemoteNotificationHandler.swift
```diff
@@ -138,7 +146,9 @@ class RemoteNotificationHandler {
 
                 // At the time when permission granted, the token will be
                 // already set, so we need to register all stored safes
-                self.registerAll()
+                Task {
```
MouazAlzahabi: Why we need a new task here  and on the other places?

Multisig/App/RemoteNotificationHandler.swift
```diff
@@ -163,6 +173,34 @@ class RemoteNotificationHandler {
...
+        // get keys for signing
+        var privateKeys: [PrivateKey] = []
+        let keyInfos = try KeyInfo.all()
+        for keyInfo in keyInfos {
+            let privateKey = try await keyInfo.pushNotificationSigningKey()!
+            privateKeys.append(privateKey)
+        }
+
+        let hashPreimage = [
+            "gnosis-safe",
```
MouazAlzahabi, line 180: I wonder if this should stay gnosis-safe or we can rebrand to "safe" only

Multisig/Logic/Models/KeyInfo.swift
```diff
@@ -393,6 +393,34 @@ extension KeyInfo {
         if let key = (try? privateKey()) { return key }
         return try? delegatePrivateKey()
     }
+
+    func pushNotificationSigningKey(completion: @escaping (Result<PrivateKey?, Error>) -> ()) {
+        privateKey { [unowned self] result in
+            switch result {
+            case .success(let privateKey):
```
MouazAlzahabi: I think this is wrong, 

pushNotificationSigningKey should return the delegate key for all key types now right? not the private key

Multisig/App/RemoteNotificationHandler.swift
```diff
@@ -225,13 +264,24 @@ class RemoteNotificationHandler {
...
                 // app coming to foreground event.
-                let signResult = try Self.sign(safes: safes,
-                                               deviceID: deviceID,
-                                               token: pushToken,
-                                               timestamp: timestamp)
-                
-                safeRegistration.append(SafeRegistration(chainId: chainSafes.chain.id!, safes: safes,
-                                                         signatures: signResult.signatures))
+
+                if AppConfiguration.FeatureToggles.securityCenter {
```
biafra23: When testing this PR on my phone the app crashes here (in AppConfiguration.FeatureToggles.securityCenter). Regardless whether security center is enabled or not.

Multisig/App/RemoteNotificationHandler.swift
elgatovital: because we are getting delegate keys async now

Multisig/Logic/Models/KeyInfo.swift
elgatovital: getting delegate key from keystore is async operation

Multisig/App/RemoteNotificationHandler.swift
elgatovital: backend has to be adjusted also. Otherwise signature check will fail. I would not touch this now.

Multisig/Logic/Models/KeyInfo.swift
elgatovital: yes, I will see that this function is not exposed

Multisig/App/RemoteNotificationHandler.swift
MouazAlzahabi: but we don't need to wait for the result, because we don't do anything even if the registration fails

elgatovital: we are not waiting for the result. Reason for this task here is that we need means to invoke async function. Collecting delegate keys happens in async function now, which is called from other async function. And task here is for invoking async function from sync context.

Multisig/App/RemoteNotificationHandler.swift
elgatovital: unable to reproduce

biafra23: Will have another look at the latest version.

biafra23: I can still reproduce the crash as follows:
- delete dev app
- install dev app 
- hit explore demo to install demo safe
- allow app to receive push notifications

App will crash
App will crash on restart after hitting "skip" on the update screen

biafra23 (approved):

Issue #2919, opened 2023-02-16
[iOS] WalletConnect V2: Core data storage


MouazAlzahabi:
**Attention:** Ticket cannot be tested on its own but needs to be tested for the overall implementation

**Two things need to be considered**
- Based on the new structure of connection the structure of storage needs to be adjusted 
- There are common classes used for connection to web and wallet connect support --> does it have an effect if we use the same structure since we need to make sure that connection to web is not affected --> if yes, we might need to create a new structure for wallet connect

**What entities need to be considered**
- Session
- Pairing
- Request
- Connection

**To be decided**
- Create new entities for pairings and sessions and request  or extend current entities

**Open questions:**
- [ ] Do we need migration or not?
- [ ] Can we ignore the previous version?
- [ ] Is there an it effect if we use the same data structure?

liliya-soroka:
Verified
mobile pairing is checked. It's not affected, works as expected

app version 3.19.0 (1621)

Issue #2917, opened 2023-02-16
[iOS] WalletConnect V2: View dApps


MouazAlzahabi:
**What needs to be done**
- We have the dapps connection/sessions list and get those from the WC library in WC V2 
- In the WC V1 we have the list of active sessions in our Safe internal DB
- We have to consolidate the list info from the WC library V2 and our internal DB --> consolidate and merge and have an unified list
- No pagination needed

**Acceptance criteria**
- Merged list of active connections/sessions of WC V1 connections and WC V2 connections is created and displayed to the user

**Comment**
No info if V1 or V2 is supported by the dapp is displayed to the user as of now. We can think of an additional scope item to display this later

liliya-soroka:
Verified
app version 3.19.0 ( 1621)

Issue #2916, opened 2023-02-16
[iOS] WalletConnect V2: Disconnect dApp


MouazAlzahabi:
Disconnect dApp that has been connected using WalletConnect V2 ((disconnect in WalletConnect V2 works with a command that takes the current session and disconnect via web3 wallet SDK)

Disconnect means in V1: killing session  and removing from storage
In Wallet Connect V2: pairing is introduced --> it needs to be decided if we define disconnect as killing session and removing from storage (or both); question is if it is for session only or for the pair

**Are changes in the UI expected?**
- Hypothesis: No changes are expected (follow-up see open questions below)

**Assumption**
No special handling for V2 is required - only different syntax needs to be considered 

This ticket covers disconnecting by the dapp as well

**Acceptance criteria**
- When disconnecting from mobile app connection is removed
- When disconnecting from the dapp the connection is removed

**Open question**
- [ ] Is the assumption correct?
- [ ] Check with web if there were any changes introduced

liliya-soroka:
Verified
app version 3.19.0(1621)
the issue that happens with a disconnect from time to time was reported separately  - https://github.com/safe-global/safe-ios/issues/2965

Issue #2915, opened 2023-02-16
[iOS] WalletConnect V2: Connect to dApp


MouazAlzahabi:
**Attention**: this ticket cannot be tested in itself but needs other tickets in WC V2 update to be implemented

Implement connect to dApp using Wallet Connect V2

**What needs to be done**
- Verify the scan link and check if V1 or V2 --> parse new parameter if V2, if V1 nothing required (details on the parameter to be parsed are documented here https://eips.ethereum.org/EIPS/eip-1328)
- Do the required steps to make the connection and use the library depending on V1 or V2

- **Parameter to parse for V1:**
- key - symmetric key used for encryption
- bridge - url of the bridge server for relaying messages

- **Parameters to parse for V2:**
- symKey - symmetric key used for encrypting messages over relay
- methods - jsonrpc methods supported for pairing topic
- relay-protocol - transport protocol for relaying messages
- relay-data - (optional) transport data for relaying messages

**Acceptance criteria**
- V1 is still working
- https://react-app.walletconnect.com/ - example for V2 dapps that we can test with; it needs to be tested if those are working properly 

**Open question**
- [ ] Duplicated references in V1 and V2: is it enough to include dependencies from V2 only?

liliya-soroka:
Verified: connection to the dapps with v1 and v2 works 
The issue with connection that happens time to time is reported separately - https://github.com/safe-global/safe-ios/issues/2967

app version 3.19.0(1621)

PR #2926, opened 2023-02-22
Connect to dapps

MouazAlzahabi:
Handles #2915 #2916 #2917 #2919


Multisig.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -199,6 +199,15 @@
           "version": "1.30.2"
         }
       },
+      {
+        "package": "Starscream",
+        "repositoryURL": "https://github.com/daltoniam/Starscream.git",
```
elgatovital, line 204: do we really need it? Every 3rd party library brings additional risk. This one had it's latest release in 2019. We might get stuck at some point due to the lack of maintenance for it.

biafra23 (approved):
LGTM 🚀

elgatovital (approved):

Multisig/UI/Dapps/DappsViewController.swift
```diff
@@ -26,7 +27,7 @@ class DappsViewController: UIViewController, UITableViewDataSource, UITableViewD
         case dapp(String)
 
         enum WalletConnect: SectionItem {
-            case activeSession(WCSession)
+            case activeSession(Any)
```
biafra23, line 30: Would it make sense to have a protocol that both session types V1 & V2 implement? And use that protocol instead of Any here?

MouazAlzahabi: I like that, will try to do

MouazAlzahabi: I will do in future pr

Multisig.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
MouazAlzahabi: I will do in next pr

Issue #2918, opened 2023-02-16
[iOS] WalletConnect V2: Handling incoming requests


MouazAlzahabi:
Signing request is different but the transaction itself is not that different.
Here is the documentation for sign request itself https://docs.walletconnect.com/2.0/swift/sign/wallet-usage

**What needs to be done**
- Check if other methods need to be supported and include them
- Parse request that we receive from the dapp in V2

Methods supported currently are (in our current code-base): 
eth_sendTransaction

**Flow of actions**
Current V1 implementation: Incoming transaction signing request --> User signs it --> Submit to client gateway --> Send the transaction hash back to the dapp

**Acceptance criteria**
- Whenever incoming request is received either we handle it, i.e. transaction signing request is handled for V1 and V2 and display proper information as it should look like and if other request is coming showing the user an error message (i.e. that we can not handle the request as it is currently implemented)

liliya-soroka:
Verified
app version 3.19.0 (1621)
the issue with incoming txs that happens time to time ,was reported separately - https://github.com/safe-global/safe-ios/issues/2967

PR #2950, opened 2023-03-06
Handle incoming requests

biafra23:
Handles #2918

- Handle `eth_sendTransaction`
- Implements flow: Open screen to review tx -> sign/reject -> select key for signing -> send tx hash to dapp and send tx to CGW
- all other requests are forwarded to Infura (and fail if they require a a signature).


Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -183,15 +199,98 @@ class WalletConnectManager {
...
+                }
+
+                let sceneDelegate = UIApplication.shared.connectedScenes.first!.delegate as! SceneDelegate
+                sceneDelegate.present(ViewControllerFactory.modal(viewController: confirmationController))
+            }
+        } else {
+            DispatchQueue.global(qos: .background).async {
+                do {
+                    let rpcURL = safe.chain!.authenticatedRpcUrl
+                    let result = try AnyCodable(any: App.shared.nodeService.rawCall(payload: request.asJSONEncodedString(),
```
biafra23: This is the same behaviour as the WC1 code. It doesn't make much sense for  `eth_signTransaction`, `personal_sign`. `eth_sign` (standard) and `eth_signTypedData`,  but these are currently impossible because the contract doesn't have the key for the address it is deployed to. This might change for `eth_sig`/`eth_singTypedData` with the advent of 1271.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -15,90 +15,98 @@ import WalletConnectUtils
...
+                    "https://app.safe.global/favicons/logo_120x120.png"])
 
-    private init() { }
+    private init() {
+    }
 
     func config() {
         Networking.configure(projectId: App.configuration.walletConnect.walletConnectProjectId,
-                             socketFactory: SocketFactory())
+                socketFactory: NativeSocketFactory())
```
biafra23: I would like to use Native WebSocket implementation for debugging wire content. And if it works reliably get rid of Starscream in a future PR

Multisig/App/SceneDelegate.swift
```diff
@@ -369,16 +369,18 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             return
         }
 
-        showWindow(makeEnterPasscodeWindow(showsCloseButton: true,
-                                           onPasscodeEnter: { [weak self] pwd in
-            try task(pwd)
-            self?.showMainContentWindow()
-        }, onError: { [weak self] error in
-            if let str = error as? String, str == "Cancelled" {
+        Task { @MainActor in
```
MouazAlzahabi: What does this solve ?

Multisig/Logic/Models/Safe.swift
```diff
@@ -92,6 +92,10 @@ extension Safe {
         return try? context.fetch(fr).first
     }
 
+    static func by(topic: String) -> Safe? {
+        Safe.all.first {$0.walletConnectSessiontopics.contains(topic)}
```
MouazAlzahabi: I believe this should have precondition to be executed on main thread, otherwise it might throw an error, that's why I made the original implementation with the precondition

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
MouazAlzahabi: Is this correct indentation?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -15,90 +15,98 @@ import WalletConnectUtils
...
     private var publishers = [AnyCancellable]()
     private var dappConnectedTrackingEvent: TrackingEvent?
 
     private let metadata = AppMetadata(
-        name: Bundle.main.displayName,
-        description: "The most trusted platform to manage digital assets on Ethereum",
-        url: App.configuration.services.webAppURL.absoluteString,
-        icons: ["https://app.safe.global/favicons/mstile-150x150.png",
-                "https://app.safe.global/favicons/logo_120x120.png"])
+            name: Bundle.main.displayName,
```
MouazAlzahabi: Is this correct indentation?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -15,90 +15,98 @@ import WalletConnectUtils
...
 
     func setUpAuthSubscribing() {
         Sign.instance.socketConnectionStatusPublisher
-            .receive(on: DispatchQueue.main)
-            .sink { [weak self] status in
-                if status == .connected {
-                    NotificationCenter.default.post(name: .wcDidConnectSafeServer, object: self)
-                } else {
-                    NotificationCenter.default.post(name: .wcDidDisconnectSafeServer, object: self)
+                .receive(on: DispatchQueue.main)
```
MouazAlzahabi: I don't have Xcode now to view, but not sure if the indentation is correct in this change and the below changes

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -15,90 +15,98 @@ import WalletConnectUtils
...
+                .store(in: &publishers)
     }
 
     func canConnect(url: String) -> Bool {
         WalletConnectURI(string: url) != nil
     }
 
     func pairClient(url: String, trackingEvent: TrackingEvent?) {
-        guard let uri = WalletConnectURI(string: url) else { return }
+        guard let uri = WalletConnectURI(string: url) else {
```
MouazAlzahabi: Based  our coding style, this return should be on same line, so as the original line

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -128,9 +136,9 @@ class WalletConnectManager {
         Task {
             do {
                 try await Sign.instance.respond(
-                    topic: request.topic,
-                    requestId: request.id,
-                    response: .error(.init(code: 0, message: ""))
+                        topic: request.topic,
```
MouazAlzahabi: Also problem with Indentation

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -140,14 +148,22 @@ class WalletConnectManager {
 
     func approveSession(proposal: Session.Proposal) {
         Task {
-            guard let safe = try? Safe.getSelected() else { return }
+            guard let safe = try? Safe.getSelected() else {
```
MouazAlzahabi: "return" line should be on same line

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -183,15 +199,98 @@ class WalletConnectManager {
...
     }
+
+    private func handle(request: Request) {
+        dispatchPrecondition(condition: .notOnQueue(.main))
+
+        guard let session = getSessions(topics: [request.topic]).first else {
+            reject(request: request)
+            return
+        }
+
```
MouazAlzahabi, line 203: Extra line

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -183,15 +199,98 @@ class WalletConnectManager {
...
+
+        guard let safe = Safe.by(topic: request.topic) else {
```
MouazAlzahabi: If you "Safe.by(topic)" to be performed on main thread then this one should be performed on main thread, that's why I made the original implementation as it was.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -183,15 +199,98 @@ class WalletConnectManager {
...
+            return
+        }
+
+        if request.method == "eth_sendTransaction" {
+
+            // make transformation of incoming request into internal data types
+            // and fetch information about safe from the request
+
+            guard let safeInfo = try? App.shared.clientGatewayService.syncSafeInfo(
+                    safeAddress: safe.addressValue, chainId: safe.chain!.id!)
```
MouazAlzahabi: Wrong Indentation

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -183,15 +199,98 @@ class WalletConnectManager {
...
+                // present confirmation controller
+
+                let confirmationController = WCIncomingTransactionRequestViewController(
+                        transaction: transaction,
+                        safe: safe,
+                        dAppName: session.peer.name,
+                        dAppIconURL: URL(string: session.peer.icons.first ?? ""))
+
+                confirmationController.onReject = { [unowned self] in
+
```
MouazAlzahabi, line 247: Extra line

MouazAlzahabi (changes requested):

Multisig/Logic/Models/Safe.swift
biafra23: Why would this throw an error when not executed on the main thread?

Multisig/App/SceneDelegate.swift
biafra23: For some reason this was executed on a background thread. I just checked. Apparently I can remove this now, as it is no longer executed on a background thread.

Multisig/Logic/Models/Safe.swift
biafra23: `private func handle(request: Request)` has the precondition to not execute on the main thread and then it uses `Safe.by()` to get the safe for the rest of the function. And that needs to be retrieved on the main thread?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: I restored `DispatchQueue.main.async { [unowned self] in` in the latest commit. Now the app crashes in `App.shared.clientGatewayService.syncSafeInfo()` when an eth_sendTransaction is received.

biafra23: I put `App.shared.clientGatewayService.syncSafeInfo()` on a background thread. And now everything seems to work again. But I am not sure it is the right was to switch treads like this.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: I didn't add that.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: It is now indented as it was in your commit. If that's not correct, then I need to know how it should be indented.

Multisig/App/SceneDelegate.swift
```diff
@@ -54,8 +54,8 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         }
 
         NotificationCenter.default.addObserver(self, selector: #selector(handlePasscodeRequired),
-                                       name: .passcodeRequired,
-                                       object: nil)
+                                               name: .passcodeRequired,
```
MouazAlzahabi, line 57: Wrong indentation

Multisig/App/SceneDelegate.swift
```diff
@@ -299,17 +299,14 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     func onAppUpdateCompletion() {
         if !AppSettings.termsAccepted {
             showWindow(makeTermsWindow())
-        // TODO: Enable when implemented new security center
+            // TODO: Enable when implemented new security center
```
MouazAlzahabi, line 302: Wrong indentation

Multisig/UI/WalletConnectV2/NativeSocketFactory.swift
MouazAlzahabi, line 1: So if we won't use the native socket because of stability issues as WC team mentioned, I assume we don't need this class anymore right ?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -183,9 +188,88 @@ class WalletConnectManager {
...
+                    let sceneDelegate = UIApplication.shared.connectedScenes.first!.delegate as! SceneDelegate
+                    sceneDelegate.present(ViewControllerFactory.modal(viewController: confirmationController))
+                }
+            } else {
+                DispatchQueue.global(qos: .background).async {
+                    do {
+                        let rpcURL = safe.chain!.authenticatedRpcUrl
+                        let result = try AnyCodable(any: App.shared.nodeService.rawCall(payload: request.asJSONEncodedString(),
+                                                                                        rpcURL: rpcURL))
+                        //let response = try Response(url: request.url, jsonString: result)
```
MouazAlzahabi, line 261: Do we need this commented line?

Multisig/UI/WalletConnectV2/NativeSocketFactory.swift
biafra23: It is really helpful for wire debugging. To see what exactly is sent and received.

biafra23: I would keep the class but deactivate it. At least while we're still developing.

MouazAlzahabi (approved):

elgatovital (approved):

Issue #2936, opened 2023-03-01
[iOS] Relaying V1: Submitting transaction to our BE for relaying


the-real-chrizzo:
Background
In our "normal" setup we are interacting with the blockchain from mobile without BE services. To enable relaying we need to change that and connect to the relaying v1 endpoint

**What needs to be done**
Submitting transactions to the blockchain via the BE endpoints (relaying V1 https://github.com/5afe/safe-gelato-relay-service/blob/main/src/routes/relay/relay.controller.ts) 

**Acceptance criteria**
Interacting with BE service and subsequent execution is working as expected

liliya-soroka:
Verified
app version 3.19.0 (1658)

PR #2952, opened 2023-03-07
gh-2936 Relay service integration

elgatovital:
Handles #2936

- Gelato relay service
- Remaining relays request
- Relay request

### DIFF
--- a/Multisig/Cross-layer/Configuration/Config.Example.xcconfig
+++ b/Multisig/Cross-layer/Configuration/Config.Example.xcconfig
@@ -50,6 +50,11 @@ _CLAIMING_DATA_URL_DEV        = https:/$()/safe-claiming-app-data.staging.5afe.d
 _CLAIMING_DATA_URL_STAGING    = https:/$()/safe-claiming-app-data.staging.5afe.dev/
 _CLAIMING_DATA_URL_PROD       = https:/$()/safe-claiming-app-data.gnosis-safe.io/
 
+// Gelato relay service
+_RELAY_URL_DEV                = https:/$()/
+_RELAY_URL_STAGING            = https:/$()/
+_RELAY_URL_PROD               = https:/$()/
+
 // App Groups
 _APP_GROUP_ID_PROD     = group.io.gnosis.multisig.prod.mainnet
 _APP_GROUP_ID_STAGING  = group.io.gnosis.multisig.staging.mainnet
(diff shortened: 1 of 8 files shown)


Multisig/Cross-layer/Configuration/Config.Example.xcconfig
```diff
@@ -50,6 +50,11 @@ _CLAIMING_DATA_URL_DEV        = https:/$()/safe-claiming-app-data.staging.5afe.d
 _CLAIMING_DATA_URL_STAGING    = https:/$()/safe-claiming-app-data.staging.5afe.dev/
 _CLAIMING_DATA_URL_PROD       = https:/$()/safe-claiming-app-data.gnosis-safe.io/
 
+// Gelato relay service
+_RELAY_URL_DEV                = https:/$()/
```
biafra23, line 54: Can you provide an endpoint URL for testing? Or do you use an instance running on your laptop?

elgatovital: I will adjust URLs after service is deployed.

Multisig/Data/Services/Safe Relay Service/SafeGelatoRelayService.swift
```diff
@@ -0,0 +1,42 @@
...
+    func execute<T: JSONRequest>(request: T) throws -> T.ResponseType {
+        try httpClient.execute(request: request)
+    }
+
+    func asyncExecute<T: JSONRequest>(request: T, completion: @escaping (Result<T.ResponseType, Error>) -> Void) -> URLSessionTask? {
+        httpClient.asyncExecute(request: request, completion: completion)
+    }
+
+    // Returns mocked response in completion
+    func asyncExecuteMock<T: JSONRequest>(request: T, completion: @escaping (Result<T.ResponseType, Error>) -> Void) -> URLSessionTask? {
```
biafra23, line 39: Why do we need this?

biafra23: Can we use it to have a unit test that returns a predefined response? For example to test that the number in `expiresAt` is converted to a Date?

biafra23 (approved):
From what I can tell without running it, it looks good.

Multisig/Data/Services/Safe Relay Service/SafeGelatoRelayService.swift
elgatovital: this was the idea behind it. Unfortunately we are not using MockJSONHttpClient in our tests yet.

Issue #2937, opened 2023-03-01
[iOS] Update links for terms and policy


the-real-chrizzo:
**What needs to be done**
Remove old links for terms and privacy policy and update to new ones in 
- App settings
- App store

New links are:
- Privacy policy: https://app.safe.global/privacy 
- Terms: https://app.safe.global/terms

More background can be found here https://www.notion.so/gnosis-safe/Terms-policy-update-f3827a6f724340deae9554d24afe8c29

**Acceptance criteria**
Privacy policy and terms links are updated in app settings and app store

elgatovital:
https://app.safe.global/app -> https://app.safe.global/privacy

liliya-soroka:
@the-real-chrizzo , 
1. could you check terms link , please?
I have checked the web app. Seems that https://app.safe.global/terms is a terms link
2. Also, Licence link is missing on the App settings->About app . Safe.global 404 Not found is opened

liliya-soroka:
@elgatovital , 404 not found is displayed instead of Policy and Terms 
1. Clean app start
2. click Policy or Terms before the onboarding is skipped on the first app start
app version 3.19.0 (1617)

liliya-soroka:
Verified
app version 3.6.3(3825)rc

liliya-soroka:
verified
app version 3.19.0 (1624)

PR #2960, opened 2023-03-14
gh-2937 Fix links

elgatovital:
Handles #2937

### DIFF
--- a/Multisig/Cross-layer/Configuration/Config.Example.xcconfig
+++ b/Multisig/Cross-layer/Configuration/Config.Example.xcconfig
@@ -124,9 +124,9 @@ CREATE_SAFE_URL = https:/$()/help.safe.global/en/articles/3876461-create-a-safe
 CONFIRMATIONS_URL = https:/$()/help.safe.global/en/articles/3952319-signature-policies
 
 // Legal URLs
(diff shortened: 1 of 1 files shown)


MouazAlzahabi (approved):

Multisig/Cross-layer/Configuration/Config.Example.xcconfig
```diff
@@ -124,9 +124,9 @@ CREATE_SAFE_URL = https:/$()/help.safe.global/en/articles/3876461-create-a-safe
 CONFIRMATIONS_URL = https:/$()/help.safe.global/en/articles/3952319-signature-policies
 
 // Legal URLs
-TERMS_URL       = https:/$()/app.safe.global/terms/
-PRIVACY_URL     = https:/$()/app.safe.global/privacy/
-LICENSES_URL    = https:/$()/safe.global/licenses/
+TERMS_URL       = https:/$()/app.safe.global/terms
```
MouazAlzahabi, line 127: @elgatovital In the ticket description I see different link, is that wrong description of it's the one intended to be used ?

elgatovital: https://app.safe.global/terms is the correct link

Issue #2920, opened 2023-02-16
[iOS] WalletConnect V2: Session life cycle controller


MouazAlzahabi:
**Comments in advance** 
- Pairing and multi-chain support is introduced in WC V2 --> this might affect session life cycle controller
- New WC library introduced in WC V2 is taking care of session life cycle 
- Reference in DB is needed though
- Session lifetime is 7 days in V2 and is automatically dropped after that time (session lifetime = 7 days is the default value in the WC library)
- No session expiry in V1 in place
- Unpairing is happening when disconnection/deleting the connection
- When disconnecting the session is removed
- In V1 we check if session are still working at the app start --> and act accordingly 

**What needs to be done**
 - [x] Start implementation with new controller handling everything related to V2 and being able to switch between V1 and V2
 - [x] Check if session life cycle is working as expected, especially taking into consideration the multi-chain support
 - [x] If possible reusing UI without any changes
 - [ ] Communicate to the dapp: prevent dapp to attempt to switch to other chain (because our UI is not supporting this) 
 -- What Dapps do this? How do I test this?
 - [ ] In V2 we should be check also at the app start if session are still working and act accordingly 
 -- Trying to find out if the SDK handles this already

**Acceptance Criteria**
- New controller handling for everything related to V2 and switch between V1 and V2 is implemented
- Successful prevention of dapp to attempt to switch to other chains
- Possibility to check on active sessions at app start

biafra23:
Notes/Questions to investigate when implementing this: 
- Is there a way to reduce the session time for debugging?
-

liliya-soroka:
Verified
app version 3.19.0 (1658)

PR #2964, opened 2023-03-16
Gh 2920/wc2 session lifecycle

biafra23:
Handles #2920

- Rejects sessions for other networks than the selected Safes network.
- After deleting a session (triggered by user or Dapp) delete unused Pairings (pairings that do nor show up in any sessions pairingTopic)

### DIFF
--- a/Multisig/UI/WalletConnectV2/DefaultSignerFactory.swift
+++ b/Multisig/UI/WalletConnectV2/DefaultSignerFactory.swift
@@ -19,26 +19,18 @@ public struct DefaultSignerFactory: SignerFactory {
 }
 
 public struct Web3Signer: EthereumSigner {
-
+    enum Web3Signer: Error {
+      case shouldNotBeCalledError
+    }
     public func sign(message: Data, with key: Data) throws -> EthereumSignature {
-        let privateKey = try EthereumPrivateKey(privateKey: [UInt8](key))
-        let signature = try privateKey.sign(message: message.bytes)
-        return EthereumSignature(v: UInt8(signature.v), r: signature.r, s: signature.s)
+        throw Web3Signer.shouldNotBeCalledError
     }
 
     public func recoverPubKey(signature: EthereumSignature, message: Data) throws -> Data {
-        let publicKey = try EthereumPublicKey(
-            message: message.bytes,
-            v: EthereumQuantity(quantity: BigUInt(signature.v)),
-            r: EthereumQuantity(signature.r),
-            s: EthereumQuantity(signature.s)
-        )
-        return Data(publicKey.rawPublicKey)
+        throw Web3Signer.shouldNotBeCalledError
     }
 
     public func keccak256(_ data: Data) -> Data {
-        let digest = SHA3(variant: .keccak256)
-        let hash = digest.calculate(for: [UInt8](data))
-        return Data(hash)
+        return Data()
     }
 }
--- a/Multisig/UI/WalletConnectV2/DefaultSignerFactory.swift
+++ b/Multisig/UI/WalletConnectV2/DefaultSignerFactory.swift
@@ -1,0 +1,44 @@
+//
+//  DefaultSignerFactory.swift
+//  Multisig
+//
+//  Created by Dirk Jäckel on 16.03.23.
+//  Copyright © 2023 Gnosis Ltd. All rights reserved.
+//
+
+import Foundation
+import CryptoSwift
+import Web3
+import Auth
+
+public struct DefaultSignerFactory: SignerFactory {
+
+    public func createEthereumSigner() -> EthereumSigner {
+        return Web3Signer()
+    }
+}
+
+public struct Web3Signer: EthereumSigner {
+
+    public func sign(message: Data, with key: Data) throws -> EthereumSignature {
+        let privateKey = try EthereumPrivateKey(privateKey: [UInt8](key))
+        let signature = try privateKey.sign(message: message.bytes)
+        return EthereumSignature(v: UInt8(signature.v), r: signature.r, s: signature.s)
+    }
+
+    public func recoverPubKey(signature: EthereumSignature, message: Data) throws -> Data {
+        let publicKey = try EthereumPublicKey(
+            message: message.bytes,
+            v: EthereumQuantity(quantity: BigUInt(signature.v)),
+            r: EthereumQuantity(signature.r),
+            s: EthereumQuantity(signature.s)
+        )
+        return Data(publicKey.rawPublicKey)
+    }
+
+    public func keccak256(_ data: Data) -> Data {
+        let digest = SHA3(variant: .keccak256)
+        let hash = digest.calculate(for: [UInt8](data))
+        return Data(hash)
+    }
+}
--- a/Multisig/UI/WalletConnectV2/WalletConnectManager.swift
+++ b/Multisig/UI/WalletConnectV2/WalletConnectManager.swift
@@ -86,7 +86,7 @@ class WalletConnectManager {
         
         Web3Wallet.instance.sessionDeletePublisher
             .receive(on: DispatchQueue.main)
-            .sink { [unowned self] (topic, reason) in
+            .sink { [unowned self] (topic, _) in
                 deleteStoredSession(topic: topic)
                 NotificationCenter.default.post(name: .wcDidDisconnectSafeServer, object: self)
             }.store(in: &publishers)
--- a/Multisig/UI/WalletConnectV2/WalletConnectManager.swift
+++ b/Multisig/UI/WalletConnectV2/WalletConnectManager.swift
@@ -124,6 +124,9 @@ class WalletConnectManager {
                 try await Web3Wallet.instance.respond(topic: request.topic, requestId: request.id, response: .response(response))
             } catch {
                 print("DAPP: Respond Error: \(error.localizedDescription)")
+                Task { @MainActor in
+                    App.shared.snackbar.show(message: error.localizedDescription)
+                }
             }
         }
     }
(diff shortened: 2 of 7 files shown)


Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -19,28 +19,28 @@ import UIKit
...
+    
     private init() { }
-
+    
     func config() {
         Networking.configure(projectId: App.configuration.walletConnect.walletConnectProjectId,
-                             socketFactory: NativeSocketFactory())
+                             socketFactory: SocketFactory())
         Pair.configure(metadata: metadata)
+        Web3Wallet.configure(metadata: metadata, signerFactory: DefaultSignerFactory())
```
MouazAlzahabi: I think this will replace our signing logic right?, otherwise we need DefaultSignerFactory to handle all key types

Multisig/UI/WalletConnectV2/DefaultSignerFactory.swift
```diff
@@ -0,0 +1,44 @@
...
+import Auth
+
+public struct DefaultSignerFactory: SignerFactory {
+
+    public func createEthereumSigner() -> EthereumSigner {
+        return Web3Signer()
+    }
+}
+
+public struct Web3Signer: EthereumSigner {
```
MouazAlzahabi: I think this code is similar to WalletConnect example code, but it's differ from our signing logic right ?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -49,20 +49,20 @@ class WalletConnectManager {
                     NotificationCenter.default.post(name: .wcDidDisconnectSafeServer, object: self)
                 }
             }.store(in: &publishers)
-
-        Sign.instance.sessionProposalPublisher
+
```
MouazAlzahabi, line 53: Why we have this extra spaces ?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +73,184 @@ class WalletConnectManager {
...
                 NotificationCenter.default.post(name: .wcDidConnectSafeServer, object: self)
             }.store(in: &publishers)
-
-        Sign.instance.sessionDeletePublisher
+        
+        Web3Wallet.instance.sessionDeletePublisher
             .receive(on: DispatchQueue.main)
-            .sink { [unowned self] response in
-                deleteStoredSession(topic: response.0)
+            .sink { [unowned self] (topic, reason) in
```
MouazAlzahabi: Do we need reason here ?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +73,184 @@ class WalletConnectManager {
...
-                try await Sign.instance.pair(uri: uri)
+                // Clean up pairings to prevent pairingAlreadyExist error
+                disconnectUnusedPairings()
+                try await Web3Wallet.instance.pair(uri: uri)
                 NotificationCenter.default.post(name: .wcConnectingSafeServer, object: self)
             } catch {
                 LogService.shared.error("DAPP: Failed to register to remote notifications \(error)")
+                Task { @MainActor in
+                    App.shared.snackbar.show(message: "DAPP: Failed to register to remote notifications: \(error)")
+                }
```
MouazAlzahabi, line 121: Can we define new GSError cases to handle error messages in this file ?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +73,184 @@ class WalletConnectManager {
...
             proposal.requiredNamespaces.forEach {
                 let caip2Namespace = $0.key
                 let proposalNamespace = $0.value
-                let chains = proposalNamespace.chains
-
-                let accounts = Set(chains.compactMap {
+                guard let chains = proposalNamespace.chains else { return }
+                
+                let selectedSafeChain = chains.filter { chain in
+                    chain.namespace == "eip155" && chain.reference == safe.chain?.id
```
MouazAlzahabi: Can we please just extract this magic value to constant variable "eip155"?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +73,184 @@ class WalletConnectManager {
...
-
+    
     func deleteStoredSession(topic: String) {
         precondition(Thread.isMainThread)
         Safe.removeSession(topic: topic)
+        disconnectUnusedPairings()
     }
-
+    
+    private func disconnectUnusedPairings() {
```
MouazAlzahabi, line 238: Sorry but I don't think this should be the way to go for this, we need to discover why exactly this happens, please check the code of WalletConnect library, it checks for the validity of session and pairing each time on fetching,

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +73,184 @@ class WalletConnectManager {
...
         }
     }
-
+    
     func disconnect(session: Session) {
         Task {
             do {
                 NotificationCenter.default.post(name: .wcDidDisconnectSafeServer, object: self)
-                try await Sign.instance.disconnect(topic: session.topic)
+                try await Web3Wallet.instance.disconnect(topic: session.topic)
```
MouazAlzahabi, line 218: Maybe if we disconnect pairing instead of disconnect session it might solve the problem ?, not sure about that

MouazAlzahabi (changes requested):

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: I am actually not sure what this is really used for since we do not use Auth v2 at all. I had to introduce this because it is necessary for initialization of Web3Wallet.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: nope

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: These are inserted when executing `Editor -> Structure -> Re-Indent `in XCode

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: Do we have a good place for this? Or can I just define it in this file?

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: All unused pairings are disconnected in `disconnectUnusedPairings()` below.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: What do you mean by: "we need to discover why exactly this happens"?

When the session is connected remotely (from the Dapp side) we only get the topic of the session after the session is already disconnected and deleted. So we do not have the pairingTopic. We agreed that we want to disconnect the pairing when a session is disconnected. I do not see any other way to achieve this than to iterate over the sessions to find the unused pairings.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +73,184 @@ class WalletConnectManager {
...
         pairClient(uri: uri)
     }
-
+    
     func pairClient(uri: WalletConnectURI) {
         Task {
             do {
-                try await Sign.instance.pair(uri: uri)
+                // Clean up pairings to prevent pairingAlreadyExist error
+                disconnectUnusedPairings()
```
biafra23, line 109: This is the place where it shouldn't be necessary to clean out old pairings. I am also not sure whether it prevents `pairingAlreadyExist` errors.  I cannot reliably reproduce this error.

Multisig/UI/WalletConnectV2/DefaultSignerFactory.swift
biafra23: I will replace this code with stubs that throw exceptions so it is clear it is not used for anything.

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
MouazAlzahabi: Can you please check if we can replace this call with something else not using 'signerFactory'

MouazAlzahabi (approved):

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
```diff
@@ -73,140 +74,186 @@ class WalletConnectManager {
...
         Task {
             do {
                 NotificationCenter.default.post(name: .wcDidDisconnectSafeServer, object: self)
-                try await Sign.instance.disconnect(topic: session.topic)
+                try await Web3Wallet.instance.disconnect(topic: session.topic)
             } catch {
-                print("DAPP: disconnectting Session error: \(error)")
+                print("DAPP: disconnecting Session error: \(error)")
+                Task { @MainActor in
+                    App.shared.snackbar.show(error: GSError.error(description: "Disconnecting Session error", error: error))
```
MouazAlzahabi, line 222: Please Create new GSError struct in 'GSError' file and use it instead

MouazAlzahabi (approved):

Multisig/UI/WalletConnectV2/WalletConnectManager.swift
biafra23: I added for new DetailedLocalizedError. Two generic errors and two specific 
1. pairing failed because it already existed
2. session creation failed because a non-safe chain was in the required namespaces

PR #2976, opened 2023-03-22
Use dev project_id everywhere

biafra23:
- Prod project_id doesn't work for several network. Using dev id everywhere


elgatovital (approved):

elgatovital:
do we need separate env for wallet connect? If so please add the task to add correct project_id for prod.

PR #3014, opened 2023-04-03
Fetch relay limit from the service

elgatovital:
- Fetch relay limit from the service instead of using hardcoded value


MouazAlzahabi (approved):

Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionViewController.swift
```diff
@@ -388,20 +387,21 @@ class ReviewExecutionViewController: ContainerViewController, PasscodeProtecting
     func getRemainingRelays() {
         switch(contentVC.model?.executionOptions.relayerState) {
         case .loading, .filled:
-            let task = controller.getRemainingRelays { [weak self] remaining in
+            let task = controller.getRemainingRelays { [weak self] remaining, limit in
```
biafra23, line 390: Looks like we're not using limit in this block. Can we replace it with _ ?

biafra23 (commented):
LGTM

Issue #2988, opened 2023-03-24
[iOS] App crash during add delegate for the installed wallet via Wallet connect 


liliya-soroka:
**Bug description**
App crash during add delegate for the installed wallet via Wallet connect
https://console.firebase.google.com/u/0/project/safe-firebase-mainnet/crashlytics/app/ios:io.gnosis.multisig.prod.mainnet/issues/878848f8f8f1f1495679b499cbf8404d?time=last-seven-days&versions=3.18.2%20(1597)&types=crash&sessionEventKey=e8343d94afad4383b12f95dd8f3d582b_1790928461355024152
**Steps To Reproduce**
1. connect key from the wallet installed on the same device via walletconnect
2. try to enable delegate
Current result: the app is crashed for some users
For me enable push notification doesn't work without the crashing 
**Expected Result**
It should be possible to enable delegate for the keys connected via wc
**Screenshots**
<img width="1119" alt="image" src="https://user-images.githubusercontent.com/338622/227451164-eecee153-907b-42d7-953e-79793487f52f.png">

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.18.2/3.19.0(1642)
 
**Environment**
prod
**Additional info**
(Optional)

biafra23:
Just to clarify this is a bug in the old v1 code, right? It happens when an owner key is added via WalletConnect? I assume the current released version does not have this problem?

biafra23:
> connect key from the wallet installed on the same device via walletconnect

Do you mean `Connect a key from a wallet (like MetaMask) installed on the same device via WalletConnect`?

biafra23:
>  For me enable push notification doesn't work without the crashing

Do you mean it is not crashing for you and the push notifications are not activated? Or is it crashing sometimes and only with the crash push notifications are working?

biafra23:
I also cannot reproduce the crash. 

I get an app crash when I show a QR: 

Steps to reproduce:
1. Select a Görli Safe
2. Settings -> Owner Keys -> + -> Connect Key -> Next -> Show QR code
3. Scan with Metamask on Android
4. Hit Connect button for a key on Görli
5. Enter a name for the Key in iOS app
6. Observe a crash of the DEV app. (build from main)

**Current released Prod app 3.18.2(1597)** doesn't crash there. But also doesn't import the key.
 
App (main branch) crashes in Line 49 of WalletConenctKeyFlow.swift: 
```
[...]

PR #3016, opened 2023-04-03
gh-2988 App crashes when using QR code to connect key via Walletconnect

biafra23:
Handles #3017  

- Wallet can be nil for importKey

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/WalletConnectKeyFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/WalletConnectKeyFlow.swift
@@ -44,14 +44,13 @@ class WalletConnectKeyFlow: AddKeyFlow {
 
     override func doImport() -> Bool {
         guard let connection = parameters?.connection,
-                let wallet = parameters?.wallet,
                 let name = parameters?.name else {
             assertionFailure("Missing key arguments")
             return false
         }
 
         guard OwnerKeyController.importKey(connection: connection,
-                                           wallet: wallet,
+                                           wallet: parameters?.wallet,
                                            name: name) else {
             WebConnectionController.shared.userDidDisconnect(connection)
             return false


Multisig/UI/Settings/OwnerKeyManagement/WalletConnectOwnerKey/WalletConnectKeyFlow.swift
```diff
@@ -44,14 +44,13 @@ class WalletConnectKeyFlow: AddKeyFlow {
 
     override func doImport() -> Bool {
         guard let connection = parameters?.connection,
-                let wallet = parameters?.wallet,
```
MouazAlzahabi, line 47: For data integrity I think we need different validation rules, so if we don't have wallet selected we need to make sure the url is valid and can be used for creating connection is that fully covered by the other checks ?

biafra23: Maybe I am wrong but it looks to me this worked in 3.18.1
How do I check the URL is valid? Do we have a test for that?

biafra23: I can try and find out what happens with a slightly broken URL

biafra23: This is what happens on the side that creates the QR-code. Do we really need to check we created a valid URL?

biafra23: Is it safe to assume that we need either a non-nil cdConnection or a non-nil wallet [here](https://github.com/safe-global/safe-ios/blob/0b05a77e4a0d722bf78a7a0cbfdcaa9f3e3ab054/Multisig/Logic/Models/KeyInfo.swift#L211) :
```
        if let cdConnection = CDWCConnection.connection(by: connection.connectionURL.absoluteString) {
            item.addToConnections(cdConnection)
        }

        if let wallet = wallet, let cdRegistryEntry = CDWCAppRegistryEntry.entry(by: wallet.id) {
            item.wallet = cdRegistryEntry
        }
```

biafra23: I don't think it is possible to get here (`doImport()`) without a connection. The crashing app showed a QR code, that is scanned by a wallet. The crash happens, after the other wallet connects to the crashing app when we enter the name for the imported key.

Issue #3018, opened 2023-04-04
[iOS] Relaying V1/V1.1: Enabling non-sponsered transaction behaviour for chains not sponsoring relaying


the-real-chrizzo:
As a user
I want to pay as usual
on chains that are not supporting "sponsored" transactions

**Background**
Due to changes in the code base in light of relaying, the "old" transaction execution flow is not available any more. This ticket intends to enable transaction execution on chain not sponsoring relaying transactions.

**What needs to be implemented**
- Based on the feature toggle, EOA/Signer key payment method interface should be displayed for not sponsored transactions; Figma designs are here https://www.figma.com/file/M3J3IJ5YFxRKaT95F2uBAt/Relaying?node-id=2702-38747&t=AxoNgg4CRtWjL2Am-4
- If feature toggle is enabled show the relaying flow should be available

Acceptance criteria
- Implemented as specified

francovenica:
LGTM

I tried to execute tx in other networks besides goerli and GNO that have the relayer available.
I used ETH, Polygon and Avax networks
I used owners created in the app, owner with imported PK and owners imported through WC (MM phone app).

Issue #3004, opened 2023-03-29
[iOS] [Relay v1] Choose how to pay should be available only on networks with relay support


liliya-soroka:
**Bug description**
Choose how to pay should be available only on networks with relay support
**Steps To Reproduce**
1. open safe on polygon
2. try to execute tx
3. click " with an owner key" on Execute screen
Current result: The option to choose between relay and EOA is displayed
**Expected Result**
Make sure that choose pop-up appears only for the networks that support relay
**Screenshots**
<img width="383" alt="image" src="https://user-images.githubusercontent.com/338622/228585302-6c636f8d-5485-44ae-b6c5-2239f6c37229.png">

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.19.0 (1564)
 
**Environment**
staging
**Additional info**
(Optional)

francovenica:
Tested on the Safe Stg and prod builds 3.19.0 (1671)

Looks good
Only Goerli and GNO chains give you the option to change between the relayer and the owner key.
For the other netwoks the option "with the owner key" is the only avaliable
The other networks tested: ETH, Poly, BNB, Avalanche, optimism, Arbitrum

Issue #3003, opened 2023-03-29
[iOS][relay v1] Show banner only on networks that supports relay 


liliya-soroka:
**Bug description**
Show banner only on networks that support relay 
**Steps To Reproduce**
1. open polygon safe on the assets page
Current result: There is a relay banner 
**Expected Result** make sure that the banner is displayed on the networks that support the relay 

**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/16.0 
 - App version: 3.19.0 (1564)
 
**Environment**
staging
**Additional info**
(Optional)

PR #3019, opened 2023-04-04
Disable relaying feature on not supported safes

MouazAlzahabi:
Handles #3004 #3003 #3018

### DIFF
--- a/Multisig/Logic/Models/Chain.swift
+++ b/Multisig/Logic/Models/Chain.swift
@@ -414,7 +414,7 @@ extension Chain {
         } ?? []
     }
 
-    func isSupport(feature: Feature) -> Bool {
+    func isSupported(feature: Feature) -> Bool {
         AppConfiguration.FeatureToggles.relay && enabledFeatures.contains(feature)
     }
 }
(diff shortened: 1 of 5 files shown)


Multisig/Logic/Models/Chain.swift
```diff
@@ -415,6 +415,6 @@ extension Chain {
     }
 
     func isSupport(feature: Feature) -> Bool {
```
elgatovital: isSupported would be more natural to read

biafra23: `isSupported(feature: Feature) -> Bool` or `supports(feature: Feature) -> Bool`

biafra23 (approved):
I added one line to this PR:

In ReviewExecutionViewController
`choosePaymentVC.relaysLimit = controller.relaysLimit`

Otherwise the sheet has the wrong number.

image

francovenica:
Testing in build Safe Stg and Prod builds, both 3.19.0 (1671) 

Looks good.
Currently only the Goerli and GNO the feature is enabled and usable
I've checked ETH, Polygon, BNB, Optimism, Arb, Avalanche and Optimism and the feature is not enabled there.

Issue #3007, opened 2023-03-31
[iOS] Relay v1 -Execute and switch payment method screens have grey background at the bottom


liliya-soroka:
**Bug description**
Execute and switch payment method screens have a grey background at the bottom instead of white one
**Steps To Reproduce**
1. open execute tx screen
2. click switch payment method
Current result: there is grey background at the bottom

**Expected Result**
Make sure that the background is white at the bottom
https://www.figma.com/file/CtKM4gMqfb6CDdU6a9rt0V/Relaying?node-id=1047-15065&t=07ONY2ZWTeudq0g6-4
**Screenshots**
<img width="377" alt="image" src="https://user-images.githubusercontent.com/338622/229107006-a63317d8-8a5a-4aeb-a80f-4ade25cb1731.png">
<img width="372" alt="image" src="https://user-images.githubusercontent.com/338622/229107231-e89ecfda-f00e-4fcd-99a2-afc1d2704de5.png">

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.19.0 (1658)
 
**Environment**
prod , gnosis chain
**Additional info**
(Optional)

PR #3020, opened 2023-04-04
Gh 3007/execute and switch payment method screens have grey background at the bottom

MouazAlzahabi:
Handles #3007

### DIFF
--- a/Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionCellBuilder.swift
+++ b/Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionCellBuilder.swift
@@ -55,7 +55,7 @@ class ReviewExecutionCellBuilder: TransactionDetailCellBuilder {
         paymentGroupCell.tableView.registerCell(DisclosureWithContentCell.self)
         paymentGroupCell.tableView.registerCell(SecondaryDetailDisclosureCell.self)
         paymentGroupCell.tableView.registerCell(PaymentMethodCell.self)
-
+        paymentGroupCell.tableView.separatorStyle = .none
         let estimatedFeeCell = buildEstimatedGasFee(model.feeState, tableView: paymentGroupCell.tableView)
 
         if case let .filled(relayerInfo) = model.relayerState,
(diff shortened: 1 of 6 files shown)


Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionCellBuilder.swift
```diff
@@ -55,11 +55,13 @@ class ReviewExecutionCellBuilder: TransactionDetailCellBuilder {
...
         paymentGroupCell.tableView.registerCell(PaymentMethodCell.self)
-
+        paymentGroupCell.tableView.separatorStyle = .none
         let estimatedFeeCell = buildEstimatedGasFee(model.feeState, tableView: paymentGroupCell.tableView)
 
         if case let .filled(relayerInfo) = model.relayerState,
-           relayerInfo.remainingRelays > ReviewExecutionViewController.MIN_RELAY_TXS_LEFT && !userSelectedSigner {
+           relayerInfo.remainingRelays > ReviewExecutionViewController.MIN_RELAY_TXS_LEFT &&
+            !userSelectedSigner &&
+            safe.chain!.isSupport(feature: .relay) {
```
elgatovital: relayerInfo should not be .filled at this point. No need to check if relayer feature is supported here. It should have been checked before.

Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionViewController.swift
```diff
@@ -400,7 +400,7 @@ class ReviewExecutionViewController: ContainerViewController, PasscodeProtecting
     }
 
     func didLoadPaymentData() {
-        if controller.relaysRemaining > ReviewExecutionViewController.MIN_RELAY_TXS_LEFT && !self.userSelectedSigner {
+        if controller.relaysRemaining > ReviewExecutionViewController.MIN_RELAY_TXS_LEFT && !self.userSelectedSigner && safe.chain!.isSupport(feature: .relay) {
```
elgatovital: not related to this pr, but what is the point of having MIN_RELAY_TX_LEFT constant that is zero?

biafra23: For debugging. You can set this to 4 and you only need to relay on tx to check the other flow. If it is not a constant you have to change zeroes in several places to achieve the same effect.

MouazAlzahabi: I'm also not in favor of pushing debug code, but this has been done in different pr

biafra23: This is not debugging code. It makes sense to have this exatracted into one constant. Instead of using "0" in several places.

elgatovital (approved):

francovenica:
Looks good. The gray square is no longer present. 
Checked in both light and dark themes

Issue #3005, opened 2023-03-29
[iOS] [Relay v1] Disable "Submit" button for execution the same way as for the tx signing


liliya-soroka:
**Bug description**
[Relay v1] Disable "Submit" button for execution the same way as for the tx signing
**Steps To Reproduce**
1. try to execute tx using relay + passcode
2.  click submit button
3. enter passcode
Current result: the user is back to the execute screen with the available Submit button and can stay there for a long time.
**Expected Result**
Could we implement the same flow for the execute as for the signing when the button are locked with loading icon?

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.19.0(1564)
 
**Environment**
staging
**Additional info**
(Optional)

PR #3021, opened 2023-04-05
gh-3005 Disable submit button while tx is being relayed

elgatovital:
Handles #3005


MouazAlzahabi (approved):

francovenica:
This is still not working for me.
Tested in both Stg and Prod builds 3.19 (1671)

After executing a tx with the relayer and click on "see the tx details" I'm taken to the tx where I can click "Reject" or "Submit" again.
In the tx review you can see the warning says the tx might fail (since is trying to be executed), but the idea is that the submit button cannot be even clicked
Also the reject button works normally, so I could try to reject a tx that is in process of being executed

francovenica:
I was mistaken in the button that was supposed to be checked. The button is the submit button that takes you to the passcode screen.
I've check that indeed, once you input the passcode and are returned to the previous page, the submit button is disabled and a loading icon shows on top of it.

LGTM

Issue #2998, opened 2023-03-29
[iOS] [Relay v1] Tx  state after execution with relay is not the same as for the execution with owner


liliya-soroka:
**Bug description**
Tx details state after execution with relay is not the same as for the execution with owner.
"Pending" status is missing
**Steps To Reproduce**
1. execute tx using Relay and check the tx state in queue list or on the tx details
Current result: Execute status is displayed and Execute button is avalable
**Expected Result**
As soon as the execution process is started "pending " state should be used for the tx on queue list and tx details, as for the execution with owner key
Execute button should not be available while the tx is pending state
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version:13.9.0(1564)
 
**Environment**
staging
**Additional info**
(Optional)

francovenica:
Tested in build 3.19 (1692)
iPhone 13pro

I tired in goerli and GNO chains, both tx with sponsored and non-sponsored tx and the tx still gets stuck in pending. I have to manually go to the history tab to see the tx executed

francovenica:
Mouaz help me to debug this. We suspect that the issue is when a tx is opened in a "modal" view, which is how you see it if you click in the "View tx details" after  you are done executing. 
If you access the tx by first going to the queue tab and clicking the tx it will open in a "full screen" view, and in that view the tx status updates just fine.

francovenica:
Tested in Prod build (1697)

It works fine now. The tx updates fine while in "modal view". I checked for both Sponsored and non-Sponsored tx.
I tested in other networks and it works fine as well

PR #3025, opened 2023-04-05
Fix pending transaction status

MouazAlzahabi:
Handles #2998


elgatovital (approved):

francovenica:
What is working:
The status changes to "pending" when executing with the relayer
The Execute and reject button are removed while the tx is in pending now

Issue:
The Pending status never goes away. I've checked the blockexplorer and even tho the tx was executed several minutes ago the Pending status persists
I've tried with relayer and without it.

Issue #3027, opened 2023-04-06
[iOS] Relay V1/V1.1 Update Review Screen


the-real-chrizzo:
Two things not to be updated in the review screen:

1. For now we would like to remove the chevron in the "Estimated Fee" part of the review screen for relayed transaction (Figma: https://www.figma.com/file/M3J3IJ5YFxRKaT95F2uBAt/Relaying?node-id=2599-39706&t=RQmgokMNlOGFdcWy-4. So for the user that means that the user cannot see the "Fee Details" screen (not be shown for relaying option https://www.figma.com/file/M3J3IJ5YFxRKaT95F2uBAt/Relaying?node-id=2599-39886&t=RQmgokMNlOGFdcWy-4).
For the execution with owner keys nothing changes.

2. The estimated fees should be struck through as indicated in the designs and the colour of the struck through fees should be changed according to the designs

**What needs to be done**
Implement 1 and 2

**Acceptance criteria**
1 and 2 implemented

francovenica:
LGTM.

The fee is shown grayed out and with a strikethrough line when the Sponsored tx are selected
The user cannot enter into the Estimation fee's form while the Sponsored tx is selected
The user can enter if the EOA wallet is selected and edit the values, but that edition does not affect the values that are used for a sponsored tx

PR #3029, opened 2023-04-06
gh-3027 Disable fee details editing for sponsored payment option

elgatovital:
Handles #3027

- Disable fee details editing for sponsored payment
- Strike through sponsored amount

### DIFF
--- a/Multisig/UI/Transaction/ExecuteTransaction/AmountAndValuePiece.swift
+++ b/Multisig/UI/Transaction/ExecuteTransaction/AmountAndValuePiece.swift
@@ -21,8 +21,19 @@ class AmountAndValuePiece: UINibView {
         setTextAlignment(.right)
     }
 
-    func setAmount(_ value: String?) {
-        amountLabel.text = value
+    func setAmount(_ value: String?, sponsored: Bool = false) {
+        if value != nil && sponsored {
+            amountLabel.attributedText = NSMutableAttributedString(
+                string: value!,
+                attributes: [
+                    NSAttributedString.Key.foregroundColor: UIColor.labelTertiary,
+                    NSAttributedString.Key.strikethroughStyle: NSUnderlineStyle.single.rawValue,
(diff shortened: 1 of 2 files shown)


Multisig/UI/Transaction/ExecuteTransaction/AmountAndValuePiece.swift
```diff
@@ -21,8 +21,19 @@ class AmountAndValuePiece: UINibView {
         setTextAlignment(.right)
     }
 
-    func setAmount(_ value: String?) {
-        amountLabel.text = value
+    func setAmount(_ value: String?, sponsored: Bool = false) {
```
MouazAlzahabi, line 24: Maybe it would be better if we use more generalized name rather than sponsored, something that covers all possible causes why we strikethrough that label

Multisig/UI/Transaction/ExecuteTransaction/AmountAndValuePiece.swift
```diff
@@ -21,8 +21,19 @@ class AmountAndValuePiece: UINibView {
...
+        if value != nil && sponsored {
+            amountLabel.attributedText = NSMutableAttributedString(
+                string: value!,
+                attributes: [
+                    NSAttributedString.Key.foregroundColor: UIColor.labelTertiary,
```
MouazAlzahabi: Please refer to file ConnectKeystoneFlow.swift line 40 and see how we used our custom styles to create attributedString and then append whatever new attributes you want

MouazAlzahabi (changes requested):

Multisig/UI/Transaction/ExecuteTransaction/AmountAndValuePiece.swift
elgatovital: it's over optimisation at this point. We can revisit naming later, but now `sponsored` suits quite well here for now.

MouazAlzahabi (approved):

Issue #3032, opened 2023-04-09
[iOS] Create Button incorrectly enabled/disabled


biafra23:
**Bug description**

1. When there is no owner key imported, the Create Safe button stay disabled, if Sponsored by Gnosis is selected (Create button should be enabled)
2. When signer key has no balance but is selected, sponsoring is used anyway (Create button should be disabled)

**Steps To Reproduce**

**Expected Result**

**Screenshots**

**Device & App version**
 - Device/OS: [e.g. iPhone8/iOS13.4.1]
 - App version:
 
**Environment**

**Additional info**
(Optional)

francovenica:
Issue:

Prod build 3.19.0 (1692)

With no owner keys in the app, the create button is still enabled in networks that don't support Sponsored tx (like ETH). The owner key "by default" is set as 0x000..000. Trying to create the safe ask for the nueric code but nothing happens after that.

Is expected for the button "Create" to be disabled if no manually imported/created owner keys exist in the app

francovenica:
Tested on the prod build 3.19 (1696)

Ok the button is grayed out when you are in a network that do not support sponsored tx.

LGTM

francovenica:
Build: Prod 3.19 (1697)
@biafra23 

I have an issue I missed:

The first time you enter to create a safe, since it is on ETH, the default method selected for payment is by owner key. When you change to GNO or goerli, the selection changes to the sponsored apps, but the create button still remains grayed out. I think is because in the selection page (where you choose how to pay for the safe creation) the owner key option is highlighted 

![create safe](https://user-images.githubusercontent.com/4503659/231515417-f7e261c7-735d-4759-811d-13af0b68b55b.gif)

francovenica:
Tested in Prod build 1698, on iOS iPhone13

I've seen that changing to the network that supports Sponsored not only shows it as the payment method, but also changes the selection to that option, but now we have the opposite problem:

Going from a network that supports Sponsored tx to a network that don't changes the payment method to the owner key, but I assume that the selection remains in the Sponsored tx, since the create button remains enabled even when there is no owner key in the app yet.

![selected button](https://user-images.githubusercontent.com/4503659/231540071-0f83cb30-185d-43dd-aca5-a0af58b0c043.gif)

francovenica:
issue fixed. LGTM

PR #3033, opened 2023-04-09
Gh 3032/create safe button

biafra23:
Handles #3032 

- Allow selectedKey and deployerBalance to be nil or less than estimated if relayer selected
- Update uiModel when userSelectedPaymentMethod selected
- Update cells when owner added

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
@@ -119,7 +119,7 @@ class CreateSafeFormUIModel {
     }
 
     var isCreateEnabled: Bool {
-        state == .ready &&
+        (state == .ready || (state == .keyNotFound && userSelectedPaymentMethod == .Relayer)) &&
         name != nil &&
         !name!.isEmpty &&
         chain != nil &&
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
@@ -128,7 +128,8 @@ class CreateSafeFormUIModel {
         threshold <= owners.count &&
         (selectedKey != nil || userSelectedPaymentMethod == .Relayer) &&
         transaction != nil &&
-        (deployerBalance != nil && deployerBalance! >= transaction.requiredBalance || userSelectedPaymentMethod == .Relayer) &&
+        (deployerBalance != nil || userSelectedPaymentMethod == .Relayer) &&
+        (deployerBalance! >= transaction.requiredBalance || userSelectedPaymentMethod == .Relayer) &&
         error == nil
     }
 
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
@@ -120,13 +120,15 @@ class CreateSafeFormUIModel {
 
     var isCreateEnabled: Bool {
         state == .ready &&
-        name != nil && !name!.isEmpty &&
+        name != nil &&
+        !name!.isEmpty &&
         chain != nil &&
         !owners.isEmpty &&
-        threshold > 0 && threshold <= owners.count &&
-        selectedKey != nil &&
+        threshold > 0 &&
+        threshold <= owners.count &&
+        (selectedKey != nil || userSelectedPaymentMethod == .Relayer) &&
         transaction != nil &&
-        deployerBalance != nil && (deployerBalance! >= transaction.requiredBalance || userSelectedPaymentMethod == .Relayer) &&
+        (deployerBalance != nil && deployerBalance! >= transaction.requiredBalance || userSelectedPaymentMethod == .Relayer) &&
         error == nil
     }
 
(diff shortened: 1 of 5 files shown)


Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
```diff
@@ -119,14 +119,17 @@ class CreateSafeFormUIModel {
     }
 
     var isCreateEnabled: Bool {
-        state == .ready &&
-        name != nil && !name!.isEmpty &&
+        (state == .ready || (state == .keyNotFound && userSelectedPaymentMethod == .Relayer)) &&
```
elgatovital: why to look for a signer key when execution is done via relayer? `state` related conditions are confusing.

biafra23: The search happens before the user has a chance to override the selected payment method.

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
```diff
@@ -274,6 +274,9 @@ class CreateSafeViewController: UIViewController, UITableViewDelegate, UITableVi
             self.uiModel.setChain(chain)
             self.chain = self.uiModel.chain
             self.initExecutionBuilder()
+            if self.chain.isSupported(feature: .relayingMobile) {
```
MouazAlzahabi, line 277: Not sure about this but maybe we need to introduce new state for the controller called "estimate relaying" same as "estimate" in which we get the relays left on parameter change

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
```diff
@@ -697,13 +701,14 @@ class CreateSafeViewController: UIViewController, UITableViewDelegate, UITableVi
         choosePaymentVC.chooseRelay = { [unowned self] in
             LogService.shared.debug("User selected Relay")
             executionOptionsCellBuilder.userSelectedSigner = false
-
+            uiModel.userSelectedPaymentMethod = .Relayer
```
MouazAlzahabi: Very minor comment but naming of enum cases should starts with lower case letters

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
```diff
@@ -119,14 +119,17 @@ class CreateSafeFormUIModel {
...
+        name != nil &&
+        !name!.isEmpty &&
         chain != nil &&
         !owners.isEmpty &&
-        threshold > 0 && threshold <= owners.count &&
-        selectedKey != nil &&
+        threshold > 0 &&
+        threshold <= owners.count &&
+        (selectedKey != nil || userSelectedPaymentMethod == .Relayer) &&
```
MouazAlzahabi: I think we can rewrite those conditions in order to avoid many checks for "userSelectedPaymentMethod == .Relayer"

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
MouazAlzahabi: Do we really need to make sure of this state == .keyNotFound?

MouazAlzahabi (changes requested):

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
biafra23: Without it the create button is not enabled when the user selects relayer execution. We could just ignore any state that is not `.ready` if `userSelectedPaymentMethod == .Relayer`. But I found that too risky.

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
biafra23: We probably could. But wouldn't that make it less readable? I find it more readable that the user selection overrides certain conditions.

biafra23:
> Do we really need to make sure of this state == .keyNotFound?

We might not want any state (except .`ready`) to be overridden by `userSelectedPaymentMethod == .Relayer`

Are we certain that any of these states can be overridden by the user selection? 

    case initial
    case setup
    case changed
    case estimating
    case searchingKey
    case keyNotFound
    case sending
    case error
    case final

For example .error should probably not be overridden by the user choice.

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
biafra23: Since we do not have unit tests for this class, I do not feel confident to rewrite this part right now.

MouazAlzahabi: Agree, let's maybe create separated ticket for it please and proceed for now

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
MouazAlzahabi: Actually I think it would make it more readable

MouazAlzahabi (approved):

PR #3034, opened 2023-04-10
create relayed transaction monitor

MouazAlzahabi:
Handles #2998 

- Monitoring relayed transactions status


elgatovital:
haven't we decided, that we don't need it for v1?

biafra23:
This works as expected for safe creation. 

But for tx execution I see no difference. Shouldn't the tx in queue disappear when executed by the relay? Without pulling to refresh? For me it stays in the queue as pending until I pull to refresh.

MouazAlzahabi:
@elgatovital unfortunately state management didn't worked properly without it
@biafra23 works fine for me, could be certain steps to reproduce ?

biafra23:
When creating a safe, the etherscan url does not point to the pending tx.

biafra23 (approved):
LGTM: 🚀

PR #3035, opened 2023-04-11
Change feature flag to RELAYING_MOBILE

biafra23:
- Allow to disable and enable RELAYING separately for mobile and web.

### DIFF
--- a/Multisig/Logic/Models/Chain.swift
+++ b/Multisig/Logic/Models/Chain.swift
@@ -400,7 +400,7 @@ extension Chain {
         case eip1271 = "EIP1271"
         case eip1559 = "EIP1559"
         case erc721 = "ERC721"
-        case relay = "RELAYING"
+        case relayingMobile = "RELAYING_MOBILE"
         case safeApps = "SAFE_APPS"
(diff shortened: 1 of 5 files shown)


biafra23:
Merging this needs to be coordinated with the flag being introduced on staging and production

MouazAlzahabi (approved):

Multisig/Logic/Models/Chain.swift
```diff
@@ -400,7 +400,7 @@ extension Chain {
         case eip1271 = "EIP1271"
         case eip1559 = "EIP1559"
         case erc721 = "ERC721"
-        case relay = "RELAYING"
+        case relayingMobile = "RELAYING_MOBILE"
```
fmrsabino, line 403: nitpick: maybe keep the previous name? 🤔 From the mobile perspective we don't need to know that this feature is "mobile specific".

fmrsabino (approved):
🚀

Multisig/Logic/Models/Chain.swift
biafra23: I want the enum to have the same name. So if I am looking for it with curl, I am not confused.

PR #3038, opened 2023-04-11
Fix the crash

MouazAlzahabi:
Handles #2988

### DIFF
--- a/Multisig/Features/Connect to Wallet/Data/CDWCAppRegistryEntry.swift
+++ b/Multisig/Features/Connect to Wallet/Data/CDWCAppRegistryEntry.swift
@@ -26,11 +26,16 @@ extension CDWCAppRegistryEntry {
         }
     }
 
-    static func entries(name: String?, role: Int16) throws -> [CDWCAppRegistryEntry] {
+    static func entries(name: String?, url: String?, role: Int16) throws -> [CDWCAppRegistryEntry] {
         do {
             let context = App.shared.coreDataStack.viewContext
             let fr = CDWCAppRegistryEntry.fetchRequest().by(name: name, role: role)
-            let entries = try context.fetch(fr)
+            var entries = try context.fetch(fr)
+            if let homePageURL = url {
+                let fr = CDWCAppRegistryEntry.fetchRequest().by(url: homePageURL, role: role)
+                let urlEntries = try context.fetch(fr)
+                entries.append(contentsOf: urlEntries)
+            }
             return entries
         } catch {
             throw GSError.DatabaseError(reason: error.localizedDescription)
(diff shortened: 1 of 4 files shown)


biafra23:
When I add a key via QR code from Metamask on a different device it wants to connect to a local Rainbowwallet when I try to subscribe push notifications. 

https://user-images.githubusercontent.com/246473/231182526-22919f56-8ddf-4226-96ec-f83760bc3d26.mov

Multisig/Features/Connect to Wallet/Data/CDWCAppRegistryEntry.swift
```diff
@@ -26,11 +26,16 @@ extension CDWCAppRegistryEntry {
         }
     }
 
-    static func entries(name: String?, role: Int16) throws -> [CDWCAppRegistryEntry] {
+    static func entries(name: String?, url: String?, role: Int16) throws -> [CDWCAppRegistryEntry] {
```
elgatovital: I would rename argument: url -> homePageURL for the declaration to be more specific

Multisig/Features/Connect to Wallet/UI/SelectWalletViewController.swift
```diff
@@ -147,8 +147,13 @@ extension SelectWalletViewController: UITableViewDelegate, UITableViewDataSource
...
         let walletConnectionVC = StartWalletConnectionViewController(wallet: wallet, chain: chain)
         walletConnectionVC.onSuccess = { [weak self] connection in
+            var connectedWallet = wallet
             self?.connection = connection
-            self?.completion(wallet, connection)
+            if connectedWallet == nil , let url = connection.remotePeer?.url.absoluteString {
+                connectedWallet = self?.walletsSource.wallets(url: url).first
+            }
+
+            self?.completion(connectedWallet, connection)
```
biafra23, line 156: Does this save a local-wallet for this connection (in my case Rainbow-Wallet) when the cennection was made via QR code?

Multisig/Features/Connect to Wallet/Data/CDWCAppRegistryEntry.swift
```diff
@@ -91,4 +96,14 @@ extension NSFetchRequest where ResultType == CDWCAppRegistryEntry {
         }
         return self
     }
+
+    func by(url: String?, role: Int16) -> Self {
```
elgatovital: here also we are interested only in a very specific url, so it might be better to rename param to homePageUrl

Multisig/Features/Connect to Wallet/Data/CDWCAppRegistryEntry.swift
```diff
@@ -26,11 +26,16 @@ extension CDWCAppRegistryEntry {
...
     }
 
-    static func entries(name: String?, role: Int16) throws -> [CDWCAppRegistryEntry] {
+    static func entries(name: String?, url: String?, role: Int16) throws -> [CDWCAppRegistryEntry] {
         do {
             let context = App.shared.coreDataStack.viewContext
             let fr = CDWCAppRegistryEntry.fetchRequest().by(name: name, role: role)
-            let entries = try context.fetch(fr)
+            var entries = try context.fetch(fr)
+            if let homePageURL = url {
```
elgatovital: should we do this only if entries is empty at this stage?

Multisig/Features/Connect to Wallet/Data/CDWCAppRegistryEntry.swift
```diff
@@ -26,11 +26,16 @@ extension CDWCAppRegistryEntry {
...
+                let fr = CDWCAppRegistryEntry.fetchRequest().by(url: homePageURL, role: role)
```
elgatovital: some comments explaining the necessity of this extra step would be very helpful

elgatovital (approved):

Issue #3036, opened 2023-04-11
[iOS] [RELAYING] Sponsored by is not disabled when relays exhausted in safe creation or review tx


biafra23:
**Bug description**
When relay limit is reached, the relay option in the payment options sheet should be disabled

**Steps To Reproduce**

- create 5 relayed tx for one safe
- click on "With an owner key" in the Payment details section

**Expected Result**
- Spnosored by button should be disdabled

**Screenshots**

**Device & App version**
safe creation | review tx
------------- | ----------
![image (9)](https://user-images.githubusercontent.com/246473/231114889-68febc53-0855-4ef2-b6c2-6fda30cf7683.png) |  ![File (26)](https://user-images.githubusercontent.com/246473/231115934-a696db3e-a2d5-4094-8ca8-4cf701b2bf0c.jpg)

 - Device/OS: [e.g. iPhone8/iOS13.4.1]
 - App version: 
 
**Environment**

**Additional info**
(Optional)

francovenica:
LGTM.

After using all the sponsored attempts the selection is defaulted to the owner key and you cannot pick the sponsored tx option

PR #3039, opened 2023-04-11
gh-3036 Preselect right payment when no relays available

elgatovital:
Handles #3036


MouazAlzahabi:
@elgatovital @biafra23  do I understand the issue correctly? 
I thought it's about disabling the option of relaying if the limit is reached, how this solve the issue ?

elgatovital:
>  do I understand the issue correctly? I thought it's about disabling the option of relaying if the limit is reached, how this solve the issue ?

@MouazAlzahabi option is disabled in payment selection dialog. There is a check that prevents sponsored payment selection when no more relays are available. Issue was mainly that wrong payment method was initially preselected thus initialising dialog incorrectly.

MouazAlzahabi (approved):

biafra23 (approved):
Works great! LFG!

PR #3047, opened 2023-04-12
gh-3032 - Change user selection when switching to a network that supports relaying

biafra23:
Handles https://github.com/safe-global/safe-ios/issues/3032#issuecomment-1505539899

-
-
-

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
@@ -122,7 +122,10 @@ class CreateSafeFormUIModel {
 
     // TODO: Refactor this and introduce new state to reflect relaying state
     var isCreateEnabled: Bool {
-        name != nil &&
+
+        LogService.shared.error("isCreateEnabled() state: \(state)")
+
+        return name != nil &&
         !name!.isEmpty &&
         chain != nil &&
         !owners.isEmpty &&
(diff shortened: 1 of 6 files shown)


biafra23:
Will create this change for main, when the other changes are there

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeViewController.swift
```diff
@@ -273,7 +273,7 @@ class CreateSafeViewController: UIViewController, UITableViewDelegate, UITableVi
             self.chain = self.uiModel.chain
             self.initExecutionBuilder()
             if self.chain.isSupported(feature: .relayingMobile) {
-                self.getRemainingRelays()
+                self.getRemainingRelays(enableRelay: true)
```
MouazAlzahabi: How can "getRemainingRelays" be called if the relay is not enabled? I think it's not the proper fix

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
```diff
@@ -34,7 +34,7 @@ class CreateSafeFormUIModel {
     var sectionHeaders: [CreateSafeFormSectionHeader] = []
     var state: CreateSafeFormUIState = .initial
     var futureSafeAddress: Address?
-    var userSelectedPaymentMethod = Transaction.PaymentMethod.signerAccount
+    var userSelectedPaymentMethod: Transaction.PaymentMethod? = nil
```
MouazAlzahabi, line 37: why we need this to be nil?

biafra23: Because the user hasn't decided yet. So no override of an other default

MouazAlzahabi (approved):

Issue #3063, opened 2023-04-17
[iOS] Relaying tracking


elgatovital:
Add missing relaying events

see
https://docs.google.com/spreadsheets/d/1mj9iQIhpM-Pak7lQMb1dmxhQn1QdJoJyhbc730dZNzA/edit#gid=1949479240

liliya-soroka:
Verified

PR #3074, opened 2023-04-19
gh-3063 Add missing relay tracking

elgatovital:
Handles #3063

- Add tracking events for success screen
- Events indicating submission status to relay service
- Distinguish btw screen and user action events

### DIFF
--- a/Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionViewController.swift
+++ b/Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionViewController.swift
@@ -633,9 +633,11 @@ class ReviewExecutionViewController: ContainerViewController, PasscodeProtecting
                 switch result {
                 case .failure(let error):
                     self.submitButton.state = .normal
+                    Tracker.trackEvent(.relayUserFailure)
                     self.didSubmitFailed(error)
 
                 case .success:
+                    Tracker.trackEvent(.relayUserSuccess)
                     self.didSubmitSuccess()
                 }
             })
@@ -646,6 +648,9 @@ class ReviewExecutionViewController: ContainerViewController, PasscodeProtecting
                 switch result {
                 case .failure(let error):
                     self.submitButton.state = .normal
+                    Tracker.trackEvent(.executeFailure, parameters: [
+                        "chain_id": self.controller.chainId
+                    ])
                     self.didSubmitFailed(error)
 
                 case .success:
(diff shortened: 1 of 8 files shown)


Multisig/UI/Transaction/ExecuteTransaction/SuccessViewController.swift
```diff
@@ -21,26 +21,29 @@ class SuccessViewController: UIViewController {
...
+        secondaryAction: String?
     ) {
         self.init(nibName: nil, bundle: nil)
         self.titleText = titleText
         self.bodyText = bodyText
         self.primaryAction = primaryAction
         self.secondaryAction = secondaryAction
+    }
+
+    func setTrackingData(trackingEvent: TrackingEvent?, trackingParams: [String: Any]? = nil) {
```
MouazAlzahabi, line 46: Why this has been moved to new method, we can still keep it in the initializer and I would keep it there

elgatovital: controller can be created without tracking info as it is optional and not essential. Also this way decision about tracking parameters can be delayed.

elgatovital: additionally with this method event and it's parameters are bundled together now.

Multisig/UI/Transaction/ExecuteTransaction/ReviewExecutionViewController.swift
```diff
@@ -633,9 +633,11 @@ class ReviewExecutionViewController: ContainerViewController, PasscodeProtecting
                 switch result {
                 case .failure(let error):
                     self.submitButton.state = .normal
+                    Tracker.trackEvent(.relayUserFailure)
                     self.didSubmitFailed(error)
 
                 case .success:
+                    Tracker.trackEvent(.relayUserSuccess)
```
biafra23, line 640: What is the difference between `.relayUserSuccess` and `.successTxRelay`? Do we need to send both?

biafra23: Apparently `.successTxRelay` is a screen event and `.relayUserSuccess` is a user event. 

Do we need the user event?

elgatovital: we don't show success screen after safe creation. Screen event `.successTxRelay` is used when success screen is shown, `.relayUserSuccess` is for successful submission to the relay service. For safe creation we have only user events (also when signed key is used). But for other transactions, asset transfer, and claiming we show the success screen with corresponding tracking events. To be consistent I have added screen event also for relaying.

biafra23 (approved):
LGTM 👍🏻

Multisig/UI/Transaction/ExecuteTransaction/SuccessViewController.swift
MouazAlzahabi: I would actually still prefer to find a way to do all in initializer, better than two calls every time, as you can see in the places we currently use, but if you want to keep it like this then fine

Issue #3068, opened 2023-04-18
[iOS] [safe creation-relay]Default key with proper amount flow is broken for the networks without relay + keys balance is not displayed any more


liliya-soroka:
**Bug description**
[iOS] [safe creation-relay]Default key with proper amount flow is broken for the networks without relay + keys balance is not displayed any more.
After relay implementation, the next features for the networks without relay were broken during safe creation flow:
1. Previously, the key with suitable for the safe creation balance was defined  and selected by default
2. Previously, the list of "Deployer accounts" contained all added to the app keys with the balance on the selected network to select the right one for the safe creation
**Steps To Reproduce**
1. import and generate a few keys with diff balance on the network without Relay ( any network except Gnosis chain)
2. start "safe creation" flow
3. add safe name
4. select proper network without relay
5. add owners
6. check the default key 
Current result: By default, the first key is used for the deployment even if there is no enough funds, as result "create " button is disabled without any error

7. click on the key to select another one
Current result: The keys list is opened , but there is no keys balance to select a right one 

**Expected Result**
1. The most suitable ( enough funds) key should be used for the safe creation by default
2. make sure that key list contains balance for each key
**Screenshots**
3.19.1(1700)
<img width="386" alt="image" src="https://user-images.githubusercontent.com/338622/232738520-f14773b4-edb8-43bc-a9be-cbe89c7ddc6d.png">

3.18.2
<img width="382" alt="image" src="https://user-images.githubusercontent.com/338622/232738607-06a0f260-e6c5-4ba7-b851-3e1a166f39a9.png">

**Device & App version**
 - Device/OS: iPhone 11/16.0
 - App version: 3.19.1(1700)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.20.0(1721)

PR #3075, opened 2023-04-19
Fix load balance

MouazAlzahabi:
Handles #3068

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
+++ b/Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
@@ -190,7 +190,12 @@ class CreateSafeFormUIModel {
 
     func setChain(_ scgChain: SCGModels.Chain) {
         let newChain = Chain.createOrUpdate(scgChain)
+        if chain.id != newChain.id {
+            selectedKey = nil
+            deployerBalance = nil
+        }
         chain = newChain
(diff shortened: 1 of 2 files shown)


Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeFormUIModel.swift
```diff
@@ -190,7 +190,12 @@ class CreateSafeFormUIModel {
 
     func setChain(_ scgChain: SCGModels.Chain) {
         let newChain = Chain.createOrUpdate(scgChain)
+        if chain.id != newChain.id {
```
elgatovital, line 193: how can this happen?

MouazAlzahabi: If you select same chain again in the UI

elgatovital: why would it have different id then if it's the same chain?

MouazAlzahabi: Here I'm trying to prevent the change if the user selected the same chain again and again 

>

biafra23 (approved):
LGTM 🚀

elgatovital (approved):

Issue #2828, opened 2022-12-20
[KeySec1] App Unlock and Unlock Challenge Configuration


DmitryBespalov:
![Image](https://user-images.githubusercontent.com/284056/208657967-38df008a-b76d-4c14-9d1f-61914795bc0e.png)

When the app started, if the Face ID or passcode is needed from user, system asks for it.

If this is enough to unlock the data store, then the app is unlocked

### App Unlock Screen

When the lock method is “Face ID”  then the app should use the following screen:

<img src="https://user-images.githubusercontent.com/284056/208633573-7312969a-b814-4625-a27c-8e0de8cebd12.png" width=300 />

The screen is shown with the “face id” prompt. If the user cancels or not passes the Face ID, then there’s an option to trigger the authentication once again using the “Unlock” button.

When the lock method is “Passcode”, then the app should use the “Enter passcode” screen (without biometrics), same as before:

<img src="https://user-images.githubusercontent.com/284056/208633589-a868f59d-14df-4930-a9f3-feda1f00bc44.png" width=300 />

DmitryBespalov:
From Dirk: perhaps we don't need the challenge, because decoding the data key will be enough to check that the user has correct passcode.

biafra23:
I think we cannot have a custom biometric screen. I think the phone will always display the system dialog. We do have a custom screen underneath already in the current implementation. 


![Image](https://user-images.githubusercontent.com/246473/213398553-68b9e525-8a5e-42e5-aa86-24d2a2ae0260.png)

DmitryBespalov:
I think that the description for this task is confusing, because I never meant to have a custom biometrics screen

biafra23:
@DmitryBespalov  I just realized my misunderstanding. The screen is just for the retry.

liliya-soroka:
The expected design is not applied 

Current result:
![Image](https://user-images.githubusercontent.com/338622/218760049-c6eeb578-8789-4bd8-8f3e-ff2e58bb777b.png)

TanyaEfremova:
Here is the [light mode design](https://www.figma.com/file/teDjWgGgqKGjFp7R3pvTBD/iOS-Master-File?node-id=2706%3A51660&t=s04juReU4A8K8W91-1) and assets



[safe-wallet-logo.pdf](https://github.com/safe-global/safe-ios/files/11287987/safe-wallet-logo.pdf)

liliya-soroka:
[...]

PR #3076, opened 2023-04-24
gh-2828 Unlock screen layout

elgatovital:
Handles #2828

- Adjust unlock screen layout


Multisig/UI/App/Unlock/FaceIDUnlockViewController.swift
```diff
@@ -19,7 +19,7 @@ class FaceIDUnlockViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        label.setStyle(.body)
+        label.setStyle(.title1.weight(.semibold))
```
MouazAlzahabi: Can you please define new text style? I would avoid using inline customized styles because it caused me a lot of pain during rebranding, and I believe in future we will use it more frequent

MouazAlzahabi (approved):

PR #3078, opened 2023-04-24
gh-2900 Keep data store always unlocked after password change

biafra23:
Handles #2900

- When changing the password or lock mechanism, the data store should be unlocked afterwards

### DIFF
--- a/Multisig/Features/Security/ProtectedKeyStore.swift
+++ b/Multisig/Features/Security/ProtectedKeyStore.swift
@@ -179,7 +179,7 @@ class ProtectedKeyStore: EncryptedStore {
         )
         try store.create(pubKeyItem)
 
-        if locked {
+        if locked && self.protectionClass != .data {
             lock()
         }
     }
--- a/Multisig/Features/Security/ProtectedKeyStore.swift
+++ b/Multisig/Features/Security/ProtectedKeyStore.swift
@@ -10,7 +10,7 @@ import Foundation
 
 class ProtectedKeyStore: EncryptedStore {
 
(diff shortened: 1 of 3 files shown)


Multisig/Features/Security/ProtectedKeyStore.swift
```diff
@@ -179,7 +179,7 @@ class ProtectedKeyStore: EncryptedStore {
         )
         try store.create(pubKeyItem)
 
-        if locked {
+        if locked && self.protectionClass != .data {
```
biafra23: Maybe it would be better to pass an unlock override so `ProtectedKeystore` doesn't need to behave differently by store kind?

MouazAlzahabi: I think this is a case specific solution, I would go with more generic solution like passing the exception as parameter

biafra23: It is not really the exception. It is happening in 50% of the cases as long as we have two stores (data & sensitive). It is the exception in the sense that the data store is already unlocked, usually.

biafra23: Also `protectionClass` has to be public in `ProtectedStore`, or we pass this flag from a place where we know which store type it is.

biafra23: The more I think about this, I think this decision **should** be made in the ProctectedKeyStore based on the protection class. So it is not possible to do this for the sensitive key store.

elgatovital (approved):

Issue #2899, opened 2023-02-02
[iOS] Error message is displayed when check FaceID is cancelled during changing security settings


liliya-soroka:
**Bug description**
Error message is displayed when check FaceID is cancelled during changing security settings
**Steps To Reproduce**
1. enable feature toggle, enable Security with FaceID
2. turn off "Unlocking the app" toggle
3. the user is asked to check FaceID-> wrong FaceID is used
4. Click "Cancel" in the retry FaceID dialog
Current result: 
<img width="370" alt="image" src="https://user-images.githubusercontent.com/338622/216326513-9adb8771-dd24-43e5-aad7-33a70f825d6f.png">

**Expected Result**
No message is required in such case or the message should be more user-friendly 

**Device & App version**
 - Device/OS: iPhone 11/15.0
 - App version: 3.19.0 (1561)
 
**Environment**
staging 
**Additional info**
(Optional)

liliya-soroka:
Reopened:
no FaceID check on disable "Unlock app"

Steps:
1. enable security v2 and faceID as a unlock app method
2. disable "Unlock app" toggle
Current result: No FaceID check

3. enable "Unlock app" toggle
Current result: There is FaceID check

Expected result: make sure that there is "FaceID" check when the "require lock method for.." toggles state is changed


app version 3.20.0(1719)

liliya-soroka:
The issue from the comment above is still existing
app version 3.20.0(1754)

PR #3079, opened 2023-04-24
Display proper error message

MouazAlzahabi:
Handles #2899 

- Display proper message when user cancel authentication
- Some fixing for error handling

### DIFF
--- a/Multisig/Data/Services/Utils/GSError.swift
+++ b/Multisig/Data/Services/Utils/GSError.swift
@@ -534,6 +534,15 @@ enum GSError {
         let loggable = true
     }
 
+    struct SecureStoreAccessError: DetailedLocalizedError {
+        let description = ""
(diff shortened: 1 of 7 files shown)


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -534,6 +534,15 @@ enum GSError {
         let loggable = true
     }
 
+    struct SecureStoreAccessError: DetailedLocalizedError {
```
elgatovital, line 537: if it's canceled by user, then it's not an access error.

elgatovital (approved):

Multisig/Data/Services/Utils/GSError.swift
MouazAlzahabi: Let's rethink about that as part of error handling

Issue #3055, opened 2023-04-13
[iOS] Intercom: lib update + workspace migration + help desk article links


med3000:
# Description

Safe is moving Intercom workspaces from US to EU hosted.

# What To Do

- Update app ID (reach out to admin in Slack) and any article links with new workspace config - can be done in Jenkins without having to do a separate release
- Update link to help desk articles as documented in this table https://docs.google.com/spreadsheets/d/153sBQbMDzvW-OCXWl0yzDZQWCkqzCVc2hjYMsqIVXcQ/edit#gid=0 - needs to be updated with a new release
- Upgrade Intercom lib to version 14.1.0

# Acceptance criteria
- App ID updated
- New links in place
- Check push notifications still work
- Check Intercom is EU now
- Check that chat window is designed/themed properly
- Everything works as before

[Technical steps to do]

install: https://developers.intercom.com/installing-intercom/docs/intercom-for-ios

hosting region: to be updated to EU https://developers.intercom.com/installing-intercom/docs/ios-data-hosting-region-configuration

biafra23:
Build 1721 has the necessary changes and can be used for testing.

liliya-soroka:
Checked:
![Image](https://user-images.githubusercontent.com/338622/235669911-5276bad6-2abe-4e10-9f9c-3a7295bea9ed.png)


Reopened: 
Please, add the missing article - 

![Image](https://user-images.githubusercontent.com/338622/235669612-c0735883-dbf9-40cd-a205-348102e70937.png)



app version 3.20.0(1721)

PR #3080, opened 2023-04-26
Gh 3055/new intercom api key

biafra23:
Handles #3055

- Changed URLs in Config.example.xcconfig
- Added region key to Info.plist
- Credentials in the CI have been updates
- **ATTENTION** If for some reason we want to update the App store app without the changes in this PR, then we need to go back to the previous intercom credentials as well.


Multisig/Info.plist
```diff
@@ -95,6 +95,8 @@
 	<false/>
 	<key>IntercomAutoIntegratePushNotifications</key>
 	<string>NO</string>
+	<key>IntercomRegion</key>
```
MouazAlzahabi, line 98: Can you help me understand why this change required ?

MouazAlzahabi (approved):

elgatovital (approved):

Multisig/Info.plist
elgatovital: We are migrating Intercom from us to eu region

elgatovital: https://developers.intercom.com/installing-intercom/docs/ios-data-hosting-region-configuration

PR #3082, opened 2023-04-26
gh-3055 Bump Intercom lib to 14.+

biafra23:
Handles #3055

- Move all Intercom calls to IntercomConfig
- migrate calls to 14.1.0
- Make everything static

### DIFF
--- a/Multisig/Cross-layer/App.swift
+++ b/Multisig/Cross-layer/App.swift
@@ -69,7 +69,5 @@ class App {
 
     let firebaseConfig = FirebaseConfig()
 
-    let intercomConfig = IntercomConfig()
-
     private init() {}
 }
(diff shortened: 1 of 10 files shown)


Multisig/Cross-layer/App.swift
```diff
@@ -69,7 +69,5 @@ class App {
 
     let firebaseConfig = FirebaseConfig()
 
-    let intercomConfig = IntercomConfig()
```
MouazAlzahabi, line 72: Why?

biafra23: It's not necessary. All calls to the intercom api are static.

elgatovital (approved):

Issue #2876, opened 2023-01-24
[iOS] Error when the user cancels Export option opening 


liliya-soroka:
**Bug description**
Error when the user cancels Export option opening 
**Steps To Reproduce**
1. Precondition: security v1 is on, passcode+faceID is set up
2. Open generated or imported key details
3. Click Export
4. close Passcode screen ( X icon is clicked)
Current result: the passcode screen is closed but the error appears
**Expected Result**
The passcode screen should be closed without any error 
**Screenshots**
<img width="366" alt="image" src="https://user-images.githubusercontent.com/338622/214305383-9605406d-d013-452a-b4d8-1c84656d781e.png">

**Device & App version**
 - Device/OS: iPhone 11/15.0
 - App version: 3.19.0 (1541)
 
**Environment**
staging 
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.20.0(1754)

Close Export screen page was tasted with diff methods:
- passcode
-faceID
- passcode &FaceID

PR #3086, opened 2023-04-27
gh-2876 Handle local access user cancellation

elgatovital:
Handles #2876
Handles #2899

- Don't show error snackbar when user cancels authentication

### DIFF
--- a/Multisig/Data/Services/Utils/GSError.swift
+++ b/Multisig/Data/Services/Utils/GSError.swift
@@ -534,7 +534,7 @@ enum GSError {
         let loggable = true
     }
 
-    struct LACancelledByUser: DetailedLocalizedError {
+    struct CancelledByUser: DetailedLocalizedError {
         let description = ""
         let reason = "The operation is canceled by user"
         let howToFix = ""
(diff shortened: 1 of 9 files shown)


Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -534,7 +534,7 @@ enum GSError {
         let loggable = true
     }
 
-    struct SecureStoreAccessError: DetailedLocalizedError {
+    struct LACancelledByUser: DetailedLocalizedError {
```
MouazAlzahabi: Would be nice to remove the LA prefix so we use this error if we want to handle other canceled by user errors

Multisig/UI/Settings/Passcode/SecuritySettingsViewController.swift
```diff
@@ -151,7 +151,11 @@ class SecuritySettingsViewController: UITableViewController {
         if AppConfiguration.FeatureToggles.securityCenter {
             App.shared.securityCenter.toggleUsage(passcodeOption: option) { [unowned self] error in
                 if let error = error {
-                    App.shared.snackbar.show(message: "Failed to toggle usage \(error.localizedDescription)")
+                    if let userCancellation = error as? GSError.LACancelledByUser {
+                        reloadData()
+                    } else {
```
MouazAlzahabi, line 156: I think we need to reload data in the else case also

MouazAlzahabi (approved):

Issue #2817, opened 2022-12-14
[KeySec1] Migration to the new security configuration


biafra23:
Application must introduce a new persistent flag / variable that denotes the current version of the SecurityCenter. If no such variable exists, it is assumed that the current version is 1.

### Fresh app start

When the app is first started, the SecurityCenter is initialized with the default settings and the security lock disabled.

The SecurityCenterVersion = 2.

### Migration from v1 to v2

To perform the migration, the app should configure new security settings, copy the data, and only enable the new security system if and only if the migration is successful. Then it should increase the stored security center version from 1 to 2.

The passcode settings correspond to the new security settings as follows:

| Old passcode setting | New security setting |
| --- | --- |
| Passcode enabled | Security lock enabled |
| Require to open app | Require for unlocking the app |
| Require for confirmations | Require for making transactions |
| Biometry is enabled | Lock method = Face ID / Touch ID / Device Passcode |
| Biometry is not enabled | Lock method = Passcode |

After the security center is set up, the migration should copy the ethereum keys, delegate keys, and seed phrases from the Keychain items to the SecurityCenter.

After successful copying of everything, the migration is considered successful.

The items to copy:

- Passcode
- For each key info (if applicable):
    - Ethereum key
    - Delegate key
    - Seed phrase

**If an error occurs, it must be logged so that we could track it in Crashlytics. In this case, the migration must be re-attempted on the next app start.**

**In case of error, show the blocking error in the UI.**

Error: "Migration failed. The support team is notified. Please check official Discord for the updates or reach out to us directly."

We won't delete the old data where it is - in case there are problems with the new functionality. Then we could still have the access to the old data in order to restore access user data.

DmitryBespalov:
high risk item

PR #3095, opened 2023-05-06
Migration

MouazAlzahabi:
Handles #2817

### DIFF
--- a/Multisig/Features/Security/SecurityCenter.swift
+++ b/Multisig/Features/Security/SecurityCenter.swift
@@ -78,7 +78,7 @@ class SecurityCenter {
 
         //TODO: check version and perform migration
         if AppSettings.securityCenterVersion == 0 {
-            migrateFromKeychainStorageToSecurityEnclave()
+            SecurityCenter.shared.migrateFromKeychainStorageToSecurityEnclave()
         } else if AppSettings.securityCenterVersion < version {
             // perform migration if needed
         }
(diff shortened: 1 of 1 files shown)


Multisig/Features/Security/SecurityCenter.swift
```diff
@@ -94,8 +94,65 @@ class SecurityCenter {
...
-    private static func migrateFromKeychainStorageToSecurityEnclave() {
-        //TODO:
+    func migrateFromKeychainStorageToSecurityEnclave() {
+        let task = { [unowned self] passcode in
+            do {
+                let keys = try KeyInfo.all()
+                // Migrate private keys
+                keys.compactMap { try? $0.privateKey() }.forEach {
+                    SecurityCenter.shared.import(id: DataID(id: $0.id), data: $0.keyData) { result in
+                        // TODO: Rollback on error
```
elgatovital, line 104: rollback everything or just this particular key import?

Multisig/Features/Security/SecurityCenter.swift
```diff
@@ -94,8 +94,65 @@ class SecurityCenter {
...
+
+                // Migrate delegate keys
+                keys.compactMap { try? $0.delegatePrivateKey() }.forEach {
+                    SecurityCenter.shared.import(id: DataID(id: $0.id), data: $0.keyData, protectionClass: .data) { result in
+                        // TODO: Rollback on error
+                    }
+                }
+
+                // Migrate to new lock method
+                if App.shared.auth.isPasscodeSetAndAvailable {
```
elgatovital, line 116: it will be very helpful to have unit tests for migrating different protection settings

Multisig/Features/Security/SecurityCenter.swift
```diff
@@ -94,8 +94,65 @@ class SecurityCenter {
...
+                                                store: dataStore)
+                    }
+                }
+
+                // Enable feature toggle
+                AppConfiguration.FeatureToggles.securityCenter = true
+            } catch {
+                Self.reset()
+                AppSettings.securityLockEnabled = false
+                AppSettings.securityLockMethod = .passcode
```
elgatovital, line 142: what if there was no lock method set before the migration?

elgatovital:
note: Increment of AppSettings.securityCenterVersion is missing

Multisig/Features/Security/SecurityCenter.swift
MouazAlzahabi: Don't know yet

Multisig/Features/Security/SecurityCenter.swift
MouazAlzahabi: I don't have full imagination what should be in the unit test for this, if you have please share

Multisig/Features/Security/SecurityCenter.swift
MouazAlzahabi: This will have no effect then, because before migration lock method stored in AppSettings.passcodeOptions

MouazAlzahabi:
> note: Increment of AppSettings.securityCenterVersion is missing

@elgatovital See line 85 please

Multisig/Features/Security/SecurityCenter.swift
elgatovital: having different initial lock settings before migration ensuring that post-migration lock settings are correct. Also making sure that keys are migrated to a correct store. E.g. delegate keys to .data, private keys to .sensitive.

Multisig/Features/Security/SecurityCenter.swift
elgatovital: so we are setting default values here, right?

MouazAlzahabi: This work is still not final, so I'm not sure if I will change this part or not yet

DmitryBespalov:
@MouazAlzahabi  this pull request is too old

PR #3113, opened 2023-05-22
gh-3011 Rebranding

elgatovital:
Handles #3011

- use Safe{Wallet} logo
- use Safe when referring to whole project
- use Safe Account when referring to a specific Safe account
- use  Safe{Wallet} when referring to the Safe interface


Multisig/UI/App/UpdateAppViewController.swift
```diff
@@ -17,11 +17,11 @@ class UpdateAppViewController: UIViewController {
         var description: String {
             switch self {
             case .required:
-                return "Your version of the Safe app is not supported anymore since it is too old. Please update your app"
+                return "Your version of the Safe{Wallet} app is not supported anymore since it is too old. Please update your app"
```
liliya-soroka, line 20: may be "Safe {Wallet} mobile app for distinguishing between web app and mobile app".?

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
```diff
@@ -121,7 +121,7 @@ class SafeSettingsViewController: LoadableViewController, UITableViewDelegate, U
             safe = selectedSafe
             updateSections()
         } catch {
-            onError(GSError.error(description: "Failed to load safe settings", error: error))
+            onError(GSError.error(description: "Failed to load safe account settings", error: error))
```
liliya-soroka: what is the difference between "safe account" and "Safe Account"?

Multisig/UI/App/UpdateAppViewController.swift
elgatovital: It should be clear from the context that app itself is not supported.

Multisig/UI/Settings/SafeSettingsViewController/SafeSettingsViewController.swift
elgatovital: it should be "Safe Account" here. Web implementation uses "Safe Account" notation and we agreed to do the same.

liliya-soroka (approved):
In terms of updated text looks good for me

Multisig/Data/Services/Utils/GSError.swift
```diff
@@ -183,8 +183,8 @@ enum GSError {
 
     struct EntityNotFound: DetailedLocalizedError {
         let description = "HTTP 404 Not Found"
-        let reason = "Safe not found."
-        let howToFix = "Please check that the Safe exists on the blockchain"
+        let reason = "Safe Account not found."
```
MouazAlzahabi, line 186: I know we already agreed on this but, Shouldn't "Account" be lower case here?

MouazAlzahabi (approved):

Multisig/Data/Services/Utils/GSError.swift
elgatovital: no, it would be inconsistent with the rest of the app and the current state of the web implementation.

Issue #3133, opened 2023-06-02
[Easy onboarding] Recognize per chain flag in config service for simpler onboarding


biafra23:
# Description

The CGW will add a feature flag for Gnosis Chain (in prod) and (probably a test network (Görli or Sepolia) in Staging). This flag needs to be recognized by the mobile app. 

For chains containing this feature flag the app will allow to create a safe with a Web3Auth type keys. 
For chains that do not have this flag, only existing key types will be used. 

The proposed name of the feature is: WEB3AUTH_CREATE_SAFE 

# What To Do

- [x] Add the feature to staging config service to Görli chain
- [x] Parse the feature from the config service response and add it to chain info
- [x] If there is at least one chain with this flag, then the new flow is entered and the Select network screen is shown from the Figma design below. Otherwise the old flow is maintained.

Figma [Design](https://www.figma.com/file/WeIhkQJ9XrmT4VBNbRact1/Web2-style-onboarding?type=design&node-id=1067%3A13189&t=GPDQeHfAEcJSxrSj-1)
Figma [Prototype](https://www.figma.com/proto/WeIhkQJ9XrmT4VBNbRact1/Web2-style-onboarding?page-id=9%3A59&type=design&node-id=873-18913&viewport=5724%2C6262%2C0.75&scaling=scale-down&starting-point-node-id=873%3A18913&show-proto-sidebar=1)

liliya-soroka:
Verified
checked the feature flag in the config service: enable/disable the feature for the chain

app version : staging 3.22.0(1856)

PR #3135, opened 2023-06-14
GH-3133 / Add WEB3AUTH feature flag to chain info

biafra23:
Handles #3133

- [x] Add the feature to staging config service to Görli chain (Not in this PR. See: https://safe-config.staging.5afe.dev/admin/chains/feature/15/change/)
- [x] Parse the feature from the config service response and add it to chain info


Multisig/UI/UI Library/View Controllers/InstructionsViewController/InstructionsViewController.swift
```diff
@@ -42,9 +42,15 @@ class InstructionsViewController: UIViewController, UITableViewDelegate, UITable
     }
     
     @IBAction func didTapButton(_ sender: Any) {
-        let createSafeVC = CreateSafeViewController()
-        createSafeVC.onClose = onClose
-        show(createSafeVC, sender: self)
+
+        if Chain.all.filter({ $0.isSupported(feature: .web3authCreateSafe) }).count > 1 {
```
MouazAlzahabi: do we want to limit network selection as first step just for the case when web3authCreateSafe enabled ?, why not always select it first?
if so, I think the if and else cases should be switched right?

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeSelectNetworkViewController.swift
```diff
@@ -0,0 +1,15 @@
...
+//  CreateSafeSelectNetworkViewController.swift
+//  Multisig
+//
+//  Created by Dirk Jäckel on 14.06.23.
+//  Copyright © 2023 Gnosis Ltd. All rights reserved.
+//
+
+import UIKit
+
+class CreateSafeSelectNetworkViewController: UIViewController {
```
MouazAlzahabi: Maybe we can use the SelectNetworkView controller after we alter it ?

Multisig/UI/UI Library/View Controllers/InstructionsViewController/InstructionsViewController.swift
biafra23: I asked @TanyaEfremova, and the old flow should be used if there is no network with web3auth enabled.

Multisig/UI/Safe Management/Create Safe/Setup Safe/CreateSafeSelectNetworkViewController.swift
biafra23: If possible, that can be done later, right?

Multisig/UI/UI Library/View Controllers/InstructionsViewController/InstructionsViewController.swift
MouazAlzahabi: I have checked the flow, we don't present the InstructionsCiewController with the chains that supports this so we won't reach this point with chains that supports web2 style, so this is not the correct place for this check

biafra23: You are right. I misunderstood your first comment. Will fix it.

MouazAlzahabi (approved):

MouazAlzahabi (approved):

PR #3167, opened 2023-07-13
Gh 3131 - Safe creation

biafra23:
Handles #3167

- Pass key and email to flow 
- create KeyInfo in flow
- use uiModel to create and submit safe creation tx to relayer
- show progress screen
- update prgress when key created 
- update progress when safe creation tx submitted to relayer
- dismiss screen and show existing progress for safe creation
- This simplifies the flow as it removes the notification prompt and the prompt to create a app pin


Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -24,7 +25,9 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
     private var relayingTask: URLSessionTask?
     let relayerService: SafeGelatoRelayService = App.shared.relayService
     private var appleWeb3AuthLogin: AppleWeb3AuthLogin!
-
+    private let uiModel = CreateSafeFormUIModel()
```
elgatovital, line 28: it would be more consistent to retrieve this model also through factory. Also why is this reference to uiModel needed?

Multisig/UI/Safe Management/Create Safe/SafeCreatingViewController.swift
```diff
@@ -44,30 +46,25 @@ class SafeCreatingViewController: UIViewController {
...
-                    NotificationCenter.default.post(name: .safeCreationUpdate, object: nil)
-                }
-            }
-        }
+        onViewDidLoad()
     }
 
     @objc func accountCreated() {
         infoView1.set(status: .success)
+        // Fake delay until account is secured...
```
elgatovital: what about this delay? Is it necessary?

biafra23: I have no idea, what this message is supposed to indicate. So I do not know where to emit this notification. That is why there is this delay. We should probably get rid of this bullet point.

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
biafra23: It is needed because otherwise it'll be lost after `addOwnerAddress()` in line 160 in function `storeKeyAndCreateSafe()`

biafra23: Since we need it to be a property of the flow, I don't think it would be helpful to create it in a/the factory.

Multisig/UI/Safe Management/Create Safe/SafeCreatingViewController.swift
elgatovital: please describe why behind it. Current comment only creates confusion.

elgatovital (approved):

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -159,10 +202,25 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
     }
+    
+    
+    // CreateSafeFormUIModelDelegate protocol methods
+    func updateUI(model: CreateSafeFormUIModel) {
+        if model.state == .ready && !didSubmit {
+            model.relaySubmit()
+            didSubmit = true
+        } else if model.state == .error {
+            LogService.shared.error("---> updateUI() called: state: \(model.state)")
```
DmitryBespalov, line 213: Please take care of not logging extra information

Multisig/UI/Safe Management/Create Safe/SafeCreatingViewController.swift
```diff
@@ -66,13 +59,17 @@ class SafeCreatingViewController: UIViewController {
 
     @objc func accountCreated() {
         infoView1.set(status: .success)
+        // It is not defined, what this "owner secured" exactly means. Thats why we check this box after a short time automatically
+        Timer.scheduledTimer(withTimeInterval: 1, repeats: false) {_ in
+            NotificationCenter.default.post(name: .safeAccountOwnerSecured, object: nil)
+        }
```
DmitryBespalov, line 65: There was a todo comment above, so maybe this wasn't fully implemented and left for completion?

DmitryBespalov: Otherwise, since the whole block was deleted, maybe then this is a dead code now?

PR #3175, opened 2023-07-17
Add post clone script and compatible Firebase.dat

biafra23:
So xcode cloud can build our app


ci_scripts/ci-post-clone.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+#  ci-post-clone.sh
+#  Multisig
+#
+#  Created by Dirk Jäckel on 17.07.23.
+#  Copyright © 2023 Gnosis Ltd. All rights reserved.
+
+env
+echo `pwd`
```
elgatovital: where and how is ci-post-clone.sh is called? Could you please add some description comments. Also explanation why it is needed to move up a directory.

biafra23: See: https://developer.apple.com/documentation/xcode/writing-custom-build-scripts
I used `cd ..` because I do not trust our scripts to work from anywhere.

biafra23: Comments added

elgatovital (approved):

PR #3180, opened 2023-07-17
Use custom URI scheme with Google Login 

biafra23:
Needs new values for: WEB3AUTH_GOOGLE_CLIENT_ID, WEB3AUTH_GOOGLE_VERIFIER and WEB3AUTH_REDIRECT_URL

Handles #

- Use uri scheme instead of universal url
-
-


Multisig/Cross-layer/Configuration/Config.Example.xcconfig
```diff
@@ -183,14 +183,7 @@ OTHER_LDFLAGS = $(inherited) -fprofile-instr-generate -ObjC
...
 WEB3AUTH_GOOGLE_VERIFIER = YOUR_WEB3AUTH_GOOGLE_VERIFIER
-// Needs to be registered in entitlements of app
-WEB3AUTH_REDIRECT_URL = https:/$()/safe-wallet-web.staging.5afe.dev/web3auth/
-
-// Needs to be registered in entitlements of the app
-_WEB3AUTH_REDIRECT_URL_DEV = https:/$()/safe-wallet-web.dev.5afe.dev/web3auth/
-_WEB3AUTH_REDIRECT_URL_STAGING = https:/$()/safe-wallet-web.staging.5afe.dev/web3auth/
-_WEB3AUTH_REDIRECT_URL_PROD = https:/$()/app.safe.global/web3auth/
-WEB3AUTH_REDIRECT_URL = $(_WEB3AUTH_REDIRECT_URL_$(SERVICE_ENV):default=$(_WEB3AUTH_REDIRECT_URL_PROD))
+WEB3AUTH_REDIRECT_URL = YOUR_WEB3AUTH_REDIRECT_URL
```
elgatovital, line 188: where is this value coming from?

elgatovital:
@biafra23 could you add some description / reasons for this change and when available links to web3auth docs?

Multisig/Cross-layer/Configuration/Config.Example.xcconfig
elgatovital: is jenkins configured to inject it?

biafra23: From Jenkins environment. Or originally the Google Oauth config. It is the value for "iOS URL scheme"

Multisig/Cross-layer/Configuration/Config.Example.xcconfig
```diff
@@ -183,14 +183,9 @@ OTHER_LDFLAGS = $(inherited) -fprofile-instr-generate -ObjC
...
-WEB3AUTH_REDIRECT_URL = https:/$()/safe-wallet-web.staging.5afe.dev/web3auth/
-
-// Needs to be registered in entitlements of the app
-_WEB3AUTH_REDIRECT_URL_DEV = https:/$()/safe-wallet-web.dev.5afe.dev/web3auth/
-_WEB3AUTH_REDIRECT_URL_STAGING = https:/$()/safe-wallet-web.staging.5afe.dev/web3auth/
-_WEB3AUTH_REDIRECT_URL_PROD = https:/$()/app.safe.global/web3auth/
-WEB3AUTH_REDIRECT_URL = $(_WEB3AUTH_REDIRECT_URL_$(SERVICE_ENV):default=$(_WEB3AUTH_REDIRECT_URL_PROD))
+// Use value from ios OAuth client. Credentials -> Safe{Wallet} (type ios) ->  Additional Information -> iOS URL scheme
+// Don't forget to add :/$()/ at the end
+WEB3AUTH_REDIRECT_URL = YOUR_WEB3AUTH_REDIRECT_URL
```
elgatovital, line 188: is there a reason behind why apple related data is in Config.Example and google related data is injected?

elgatovital (approved):

elgatovital:
@biafra23 we can adjust WEB3AUTH_GOOGLE_VERIFIER key later after confirmation from web3auth

Multisig/Cross-layer/Configuration/Config.Example.xcconfig
biafra23: The apple verifiers have the bundle id confgured as the `aud` parameter. Apple sets this parameter and the SDK verifies it is correct. So no one else can use the respective verifier to create keys.

biafra23:
> could you add some description / reasons for this change and when available links to web3auth docs?

Unfortunately I am not aware of any docs outlining the difference between universal links and uri schemes on the web3auth side. Fact is, the universal links do not work reliable on all phones and do not work at all on the simulator. Instead of opening the app, the website is opened in a browser window.

PR #3202, opened 2023-07-23
Back to url schemes

biafra23:
Handles #

- According to Apples documentation url schemes are safe to use with [ASWebAuthenticationSession](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession)

``` 
ASWebAuthenticationSession ensures that only the calling app’s session receives 
the authentication callback, even when more than one app registers the same callback
URL scheme.
```


DmitryBespalov (approved):

Jenkinsfile
```diff
@@ -26,10 +26,9 @@ pipeline {
         CODECOV_TOKEN = credentials('CODECOV_TOKEN')
         WALLETCONNECT_PROJECT_ID_STAGING = credentials('WALLETCONNECT_PROJECT_ID_STAGING')
         WALLETCONNECT_PROJECT_ID_PROD = credentials('WALLETCONNECT_PROJECT_ID_PROD')
-	WEB3AUTH_GOOGLE_CLIENT_ID_STAGING = credentials('WEB3AUTH_GOOGLE_CLIENT_ID_STAGING')
-	WEB3AUTH_GOOGLE_CLIENT_ID_PROD = credentials('WEB3AUTH_GOOGLE_CLIENT_ID_PROD')
-	WEB3AUTH_GOOGLE_VERIFIER_STAGING = credentials('WEB3AUTH_GOOGLE_VERIFIER_STAGING')
-	WEB3AUTH_GOOGLE_VERIFIER_PROD = credentials('WEB3AUTH_GOOGLE_VERIFIER_PROD')
+	WEB3AUTH_GOOGLE_CLIENT_ID = credentials('WEB3AUTH_GOOGLE_CLIENT_ID')
```
MouazAlzahabi: I believe we still need to differentiate between Staging/Prod for verifier and Client ID

MouazAlzahabi: Even The Schema will be different in this case, no?

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -108,36 +108,36 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
-            view.onViewDidLoad = {
-                self.storeKeyAndCreateSafe(key: key, email: email, keyType: .web3AuthGoogle)
-            }
             self.show(view)
-        }
 
-        loginModel.loginWithCustomAuth(caller: navigationController)
-    }
+        } , keyGenerationComplete: { key, email in
+            self.storeKeyAndCreateSafe(key: key, email: email, keyType: .web3AuthGoogle)
```
MouazAlzahabi: We need a weak reference for self in this closure

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -108,36 +108,36 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
+        
         guard let privateKey = privateKey else {
             App.shared.snackbar.show(message: "Couldn't create private key from: [\(key)]")
             return
         }
         var keyInfo: KeyInfo? = try? KeyInfo.firstKey(address: privateKey.address)
         if keyInfo == nil {
             do {
-
+
```
MouazAlzahabi: Extra line ?

Multisig/UI/Safe Management/Create Safe/GoogleWeb3AuthLoginModel.swift
```diff
@@ -3,42 +3,46 @@ import CustomAuth
 import UIKit
 
 class GoogleWeb3AuthLoginModel {
-    var onClose: ((_ key: String?, _ email: String?) -> Void)
+    var authorizationComplete: () -> Void
+    var keyGenerationComplete: ((_ key: String, _ email: String?) -> Void)
```
MouazAlzahabi, line 8: no need to define parameter names ((String, String?) -> Void) should be enough

Multisig/UI/Safe Management/Create Safe/GoogleWeb3AuthLoginModel.swift
```diff
@@ -3,42 +3,46 @@ import CustomAuth
...
             return true
         }
         return false
     }
 
-    func loginWithCustomAuth(caller: UIViewController) {
+    func loginWithCustomAuth() {
         Task {
-            let sub = SubVerifierDetails(loginType: .web,
+            let sub = SubVerifierDetails(loginType: .installed,
```
MouazAlzahabi, line 25: Why installed? shouldn't it be safer if we used web?

Multisig/UI/Safe Management/Create Safe/GoogleWeb3AuthLoginModel.swift
```diff
@@ -3,42 +3,46 @@ import CustomAuth
...
                                          loginProvider: .google,
                                          clientId: App.configuration.web3auth.googleClientId,
                                          verifier: App.configuration.web3auth.googleVerifier,
-                                         redirectURL: App.configuration.web3auth.redirectUrl
+                                         redirectURL: App.configuration.web3auth.redirectScheme + "://"
```
MouazAlzahabi: this is the second place were we append "://", I would prefer if we predefine it

Jenkinsfile
biafra23: Yes, we need prod & staging variants here.

Schema is depending on the client id provided by google. that is correct.

Multisig/UI/Safe Management/Create Safe/GoogleWeb3AuthLoginModel.swift
biafra23: I find this more readable. And easier to understand. Is there a disadvantage to it?

Multisig/UI/Safe Management/Create Safe/GoogleWeb3AuthLoginModel.swift
biafra23: Won't work with .web.  
I guess .web needs to match the web oauth clients and .installed matched iOS type.

Jenkinsfile
biafra23: Why do we need _PROD & _STAGING in this file? Why not handle it like the INFURA_KEY?

biafra23: Jenkinsfile is updated

DmitryBespalov (approved):

PR #3215, opened 2023-07-25
Filter token list

MouazAlzahabi:
- Filter token list based on wallet address


biafra23 (approved):
Looks good!

Issue #3197, opened 2023-07-19
[iOS] [web3auth]Progress screen is running even when select google account is canceled


liliya-soroka:
**Bug description**
Progress screen is running even when select google account is canceled
**Steps To Reproduce**
1. start safe creation on goerli chain with web3auth
2. click on "create with Google" button
3. Cancel choose account
Current result: Progress screen is under with loading icons and the user has to go back-back to cancel the process
**Expected Result**
The user should be back to the "create a safe account with web3auth" screen instead of the "progress" screen if the google account were not selected 
**Screenshots**

**Device & App version**
 - Device/OS: iPhone 11/16.4
 - App version: 3.22.0 (1857)
 
**Environment**
prod, gnosis chain 
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.22.0(1876)

Checked:
- Cancel "select Google account"
- cancel "select Apple id"

PR #3218, opened 2023-07-26
GH-3197 - Replace callbacks to prevent multiple views to be opened on cancel

biafra23:
Handles #3197

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
+++ b/Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
@@ -12,8 +12,6 @@ import AuthenticationServices
 
 
 class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProviding, ASAuthorizationControllerDelegate, CreateSafeFormUIModelDelegate {
-    private var loginModel: GoogleWeb3AuthLoginModel?
-
     func presentationAnchor(for controller: ASAuthorizationController) -> ASPresentationAnchor {
         return ASPresentationAnchor()
     }
@@ -24,11 +22,12 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
     var owner: KeyInfo!
     var createPasscodeFlow: CreatePasscodeFlow!
     private var relayingTask: URLSessionTask?
-    let relayerService: SafeGelatoRelayService = App.shared.relayService
+    private let relayerService: SafeGelatoRelayService = App.shared.relayService
     private var appleWeb3AuthLogin: AppleWeb3AuthLogin!
     private let uiModel = CreateSafeFormUIModel()
     private var didSubmit = false
-    
+    private var loginModel: GoogleWeb3AuthLoginModel!
+
     init(_ factory: CreateSafeFlowFactory = CreateSafeFlowFactory(), completion: @escaping (_ success: Bool) -> Void) {
         self.factory = factory
         super.init(completion: completion)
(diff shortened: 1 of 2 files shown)


Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -108,18 +109,25 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
             }
             self.show(view)
-
-        } , keyGenerationComplete: { [weak self] key, email in
-            self?.storeKeyAndCreateSafe(key: key, email: email, keyType: .web3AuthGoogle)
-        })
-
-        loginModel.loginWithCustomAuth()
+        }
+        let keyGenerationComplete = { (key, email) in
```
DmitryBespalov: seems like `[weak self]` is missing here

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -108,18 +109,25 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
+            self.storeKeyAndCreateSafe(key: key, email: email, keyType: .web3AuthGoogle)
+        }
+        if let loginModel = loginModel {
+            loginModel.authorizationComplete = authorizationComplete
+            loginModel.keyGenerationComplete = keyGenerationComplete
+        } else {
+            loginModel = GoogleWeb3AuthLoginModel(authorizationComplete: authorizationComplete,
```
MouazAlzahabi: I would refactor GoogleWeb3AuthLoginModel and make the initializer takes no parameters in this case, this will improve the code her

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
biafra23: Same for **authorizationComplete**?

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
biafra23: Done

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -108,21 +108,27 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
-                self.stop(success: true)
+        let authorizationComplete = { [weak self] in
+            let view = self?.factory.safeCreatingViewController()
+            view?.onSuccess = {
+                self?.stop(success: true)
             }
-            self.show(view)
-
-        } , keyGenerationComplete: { [weak self] key, email in
+            self?.show(view!)
```
DmitryBespalov: the standard handling of the `[weak self]` when there's more than one call using it is to add guard in the closure, like this:

```
 let authorizationComplete = { [weak self] in
            guard let self = self else { return }
            let view = self.factory.safeCreatingViewController()
            view.onSuccess = { [weak self] in
                self.stop(success: true)
            }
            self.show(view)
```

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -108,21 +108,27 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
+        let keyGenerationComplete = { [weak self] (key, email) in
             self?.storeKeyAndCreateSafe(key: key, email: email, keyType: .web3AuthGoogle)
-        })
+        }
 
-        loginModel.loginWithCustomAuth()
+        if loginModel == nil {
+            loginModel = GoogleWeb3AuthLoginModel()
+        }
+        loginModel?.authorizationComplete = authorizationComplete
```
DmitryBespalov: at this point, the `loginModel` is never nil, why is the optional unwrapping here? maybe force unwrap or use a local variable and then assign that to the property

DmitryBespalov (approved):

Issue #3189, opened 2023-07-18
[iOS][Web3Auth] "Create a safe Account" screen is never closed for the first created safe when onboarding was skipped


liliya-soroka:
**Bug description**
"Create a safe Account" screen is never closed for the first created safe when onboarding was skipped 
**Steps To Reproduce**
1. clean app install
2. skip onboarding
3. skip the last step of onboarding
4. close "what is new" screen
5. click "Create a new safe" in the assets tab( or any other app tab)
6.  try to pass safe creation with social login ( I used Apple ID)
Current result: "Create a Safe Account" screen is never closed automatically even when the safe is created and loaded on UI. The user has to close it manually by going back->back-> back to the first page of the safe creation flow
**Expected Result**
Make sure that "Create a Safe account" screen is closed if the safe is created
**Screenshots**

https://github.com/safe-global/safe-ios/assets/338622/dac74849-04c7-49e6-88f5-0abca8f9d94a

**Device & App version**
 - Device/OS: iPhone 11/16.4
 - App version: 3.22.0(1848)
 
**Environment**
prod, Gnosis chain
**Additional info**
(Optional)

liliya-soroka:
Verified
app version 3.22.0(1883)

Checked:
- safe creation from onboarding screen
- safe creation when onboarding was skipped fully 
- safe creation from switch safe menu

PR #3220, opened 2023-07-27
GH-3189 dismiss screen explicitly

biafra23:
Handles #3189

- dismiss controller
-
-

### DIFF
--- a/Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
+++ b/Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
@@ -94,7 +94,7 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
                 let view = self.factory.safeCreatingViewController()
                 view.onSuccess = { [weak self, unowned view] in
                     view.dismiss(animated: true) {
-                        // The dismiss() should not be necessary here, biut the SafeCreatingViewCobntroller is not dismissed, if omitted
+                        // The dismiss() should not be necessary here, but the SafeCreatingViewCobntroller is not dismissed, if omitted
                         self?.stop(success: true)
                     }
                 }
@@ -116,7 +116,7 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
             guard let self = self else { return }
             let view = self.factory.safeCreatingViewController()
             view.onSuccess = { [weak self, unowned view] in
-                // The dismiss() should not be necessary here, biut the SafeCreatingViewCobntroller is not dismissed, if omitted
+                // The dismiss() should not be necessary here, but the SafeCreatingViewCobntroller is not dismissed, if omitted
                 view.dismiss(animated: true) {
                     self?.stop(success: true)
                 }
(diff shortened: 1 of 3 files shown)


Multisig/UI/Safe Management/Create Safe/SafeCreatingViewController.swift
```diff
@@ -75,6 +75,7 @@ class SafeCreatingViewController: UIViewController {
         Timer.scheduledTimer(withTimeInterval: 3, repeats: false) { [weak self] timer in
             DispatchQueue.main.async {
                 self?.onSuccess()
+                self?.dismiss(animated: true)
```
biafra23: This fixes the bug, but there might be a better way

DmitryBespalov: I think this is misplaced: we put the dismiss of a view controller near the place where it was opened (on the calling side)

MouazAlzahabi: same here, specially that we have onSuccess handler

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -113,6 +113,7 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
             let view = self.factory.safeCreatingViewController()
             view.onSuccess = { [weak self] in
```
MouazAlzahabi: I think this is wrong, 

As it's presented using "show" function then is should be dismissed when the flow stop function and dismiss should have no effect on it,

DmitryBespalov: second that

biafra23: The presenter (LoadSafeViewController) is dismissed in UIFlow.stop() and the Flow is set to nil in the completion closure, but (without this change) the view is not dismissed.

MouazAlzahabi: Then maybe we have a referencing problem and the correct fix to solve it

DmitryBespalov (approved):

Multisig/UI/Safe Management/Create Safe/CreateSafeFlow.swift
```diff
@@ -89,10 +89,14 @@ class CreateSafeFlow: UIFlow, ASAuthorizationControllerPresentationContextProvid
...
         appleWeb3AuthLogin = AppleWeb3AuthLogin(
-            authorizationComplete: {
+            authorizationComplete: { [weak self] in
+                guard let self = self else { return }
                 let view = self.factory.safeCreatingViewController()
-                view.onSuccess = {
-                    self.stop(success: true)
+                view.onSuccess = { [weak self, unowned view] in
+                    view.dismiss(animated: true) {
+                        // The dismiss() should not be necessary here, biut the SafeCreatingViewCobntroller is not dismissed, if omitted
```
DmitryBespalov: typo

Issue #3222, opened 2023-07-27
[On-ramping] Top up EOA of the selected Safe Account


the-real-chrizzo:
**What needs to be done?**
When pressing the "Buy" button on the home screen an overview of the currently selected Safe Account and its owners should appear. By pressing on either the owner key or the Safe Account the user will be able to top up since the MoonPay SDK will be triggered.

Implementation according to the designs.

**Acceptance criteria**
Solution is implemented according to the design.

**Reference:**
https://www.figma.com/file/sVPEmh6cvlcfQy5U2zl80H/On-ramping?type=design&node-id=1027%3A12826&mode=design&t=f00r7XTxZLM4EmMV-1

liliya-soroka:
Click on the owner address copies address instead of selecting the owner. To select owner you need to click in between the addresses
app version 3.22.0(1892)

liliya-soroka:
Returned back:
still safe address is used as recipient even when owner was clicked

app version 3.22.0(1894)

liliya-soroka:
Verified
app version 3.22.0(1896)

SelectedEOA is prefilled as recipient

Issue #3166, opened 2023-07-12
[On-ramping] MoonPay integration


the-real-chrizzo:
Integrate MoonPay

These are recommendations on how to start:
- Use the SDK to implement the widget. 
- The SDK provides a communication library so the widget can emit events to your site / app, specifically onLogin, onUnsupportedRegion, and onInitiateDeposit with additional events in our development pipeline. You can build a tighter integration with better UX when using the SDK. 
- iOS SDK: https://docs.moonpay.com/moonpay/developer-resources/sdks/ios-sdk
- Integration Best Practices: https://docs.moonpay.com/moonpay/implementation-guide/on-ramp/references/integration-best-practices
- MoonPay documentation https://docs.moonpay.com/moonpay/

- Design in figma https://www.figma.com/file/sVPEmh6cvlcfQy5U2zl80H/On-ramping?type=design&node-id=9-59&mode=design&t=elVYgLaODpOxxyHs-0

- From a recent message in slack (might be relevant): "Your developers should pass `?themeId=7e8968f6-99a4-43b5-8286-d3c290d3a0b2` when initializing the SDK"

- For the details on testing/QA (also e.g. for apple pay SDK testing details) please check https://www.notion.so/safe-global/On-ramp-MoonPay-Integration-2d82110dc7664d67a6d6236382d9c9c9

liliya-soroka:
Found issues:
1.  https://github.com/safe-global/safe-ios/issues/3204
2. https://github.com/safe-global/safe-ios/issues/3205
3. https://github.com/safe-global/safe-ios/issues/3208
4. https://github.com/safe-global/safe-ios/issues/3209
5. https://github.com/safe-global/safe-ios/issues/3210
6. https://github.com/safe-global/safe-ios/issues/3212

liliya-soroka:
assigned back to @MouazAlzahabi to finish filtering and prod mode set up

liliya-soroka:
moved back to check filtering in the prod app
app version 3.22.0(1892)

liliya-soroka:
Filtering works as expected for the netwroks with and without support

app version 3.22.0 (1908)

PR #3228, opened 2023-07-31
Create top up account screen and fix sdk initiation

MouazAlzahabi:
Handles #3222 #3166 

- Create select top up screen logic only, still need to do some UI adjustment later
- Create signature and fix token filtering issue


Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
```diff
@@ -96,7 +96,17 @@ class BalancesViewController: LoadableViewController, UITableViewDelegate, UITab
             guard let safe = try? Safe.getSelected() else {
                 return
             }
-            App.shared.ramper.startOnRamp(safe: safe)
+
+            let vc = SelectTopUpAddressViewController()
```
biafra23, line 100: Does it make sense to de-duplicate this? Looks like it is the same as in AssetsViewController.

Jenkinsfile
```diff
@@ -24,6 +24,8 @@ pipeline {
         INTERCOM_PROD_APP_ID = credentials('INTERCOM_PROD_APP_ID')
         MOONPAY_STAGING_API_KEY = credentials('MOONPAY_STAGING_API_KEY')
         MOONPAY_PROD_API_KEY = credentials('MOONPAY_PROD_API_KEY')
+	MOONPAY_STAGING_SECRET_KEY = credentials('MOONPAY_STAGING_SECRET_KEY')
```
elgatovital, line 27: what is a secret key? What's the difference btw api and secret key?

Multisig/UI/Assets/Ramper.swift
```diff
@@ -9,59 +9,69 @@
...
 
 class Ramper: MoonpayCallbackInterface {
 
     private let MOONPAY_THEME_ID_SAFE = "7e8968f6-99a4-43b5-8286-d3c290d3a0b2"
-    private let moonpay: MoonpaySDK
+    private let moonpay: MoonPayiOSSdk
 
     init() {
-        moonpay = MoonpaySDK()
+        moonpay = MoonPayiOSSdk()
```
elgatovital, line 20: MoonPayiOSSdk is hard to read. Maybe we should give moonpay team feedback about naming :)

Jenkinsfile
elgatovital: probably this https://docs.moonpay.com/moonpay/developer-resources/api/server-side-apis/authentication

elgatovital: I don't understand why we need both

elgatovital (approved):

Multisig/UI/Assets/Ramper.swift
MouazAlzahabi: This is not the only thing I would like to report actually, but I won't be in favor of waiting for another version of the sdk

Multisig/UI/Assets/BalancesViewController/BalancesViewController.swift
MouazAlzahabi: Will do in the next pr

Issue #3191, opened 2023-07-18
[iOS][Web3Auth]No errors handling, and the user is stuck on the "progress" screen when there is a problem with the tx submission.


liliya-soroka:
**Bug description**
No error handling for the cases when Tx submission is failed. The user stays on the "progress" screen and should return to the first screen to cancel the safe creation.

An easy way to check is to turn off the Wi-Fi on the "progress" screen during the safe creation with social login.
For the usual safe creation flow with the wallet, we display the errors under the "Payment" block for different problems with the tx submission.
<img width="380" alt="image" src="https://github.com/safe-global/safe-ios/assets/338622/4ce63faa-4d8d-4706-8cd1-1eafedfde27c">

**Steps To Reproduce**
1. Start the safe creation flow and turn the internet connection off
Current result: The user is stuck on the "progress" screen 

**Expected Result**
Do we have a timeout to show an error and a logic for the "progress" screen in case of an error with the tx submission/estimation/problems with our back-end?
**Screenshots**

**Device & App version**
 - Device/OS: [e.g. iPhone8/iOS13.4.1]
 - App version:
 
**Environment**

**Additional info**
(Optional)

liliya-soroka:
Reopened:
for the no internet connection

app version 3.22.0(1876)
staging app

check the steps in the ticket

liliya-soroka:
Verified
app version 3.22.0(1901)

PR #3235, opened 2023-07-31
GH-3191 step 1 - pass errors and show them

DmitryBespalov:
Handles #3191 

- Added error param in the completion handlers

Next step is to handle error case in the UI flow, so that the ui is back to the start from 'in progress'


biafra23 (approved):
LGTM 🚀

PR #3239, opened 2023-08-01
GH-3191 stopping the flow on error

DmitryBespalov:
Handles #3191 

- Stopping the flow on any error
- Restored enabling passocde


MouazAlzahabi (approved):

biafra23 (approved):
LGTM 🙏🏻 

When putting the phone in flight mode while safe creation screen is progressing, the whole flow is canceled. Excpected behaviour is that only the progress screen is dismissed.

DmitryBespalov:
That the flow is closed, is by intention. This makes handling of all of the error cases simple. I think it is good enough.

Will double-check with product

Issue #3201, opened 2023-07-20
[Easy onboarding] Settings updates 2


the-real-chrizzo:
We removed scope from the original ticket due to time constraints.

**What needs to be done**
- When clicking on ">" in the owners section, the "_Safe Settings_" appears
- Implement UI of "_Safe Settings_"

- For this "_Safe Settings_" screen we decided to:
1. Remove the blue box "This is one..."
2. Add "QR" code (as for the other key types)
3. Enable push notifications toggle in this first iteration
4. Add remove owner key

**Acceptance criteria**
- UI implemented as specified above

**Link to designs**
https://www.figma.com/file/WeIhkQJ9XrmT4VBNbRact1/Web2-style-onboarding?type=design&node-id=1479-23742&mode=design&t=DEa507eLk48m9Jom-4

biafra23:
Design | Current implementation
------- | ----------------
<img width="402" alt="image" src="https://github.com/safe-global/safe-ios/assets/246473/d9225de8-c258-4ccf-bc40-6d3046c53128"> |  ![image](https://github.com/safe-global/safe-ios/assets/246473/74ed9744-58bc-46ff-b64d-4b29708abd03)

biafra23:
~~This cell is used in a lot of places:~~
<img width="750" alt="image" src="https://github.com/safe-global/safe-ios/assets/246473/3a0af6cd-f742-4473-b417-ce0cbd9207b5">

~~But the Link to the explorer on the right should only be replaced with a chevron when used in the Safe Settings?~~

the-real-chrizzo:
@TanyaEfremova could you take a look here please

liliya-soroka:
Verified
app version 3.23.0(1948)

PR #3241, opened 2023-08-01
gh-3201 switch between block explorer and owner details

biafra23:
Handles #3201

social account | other accounts
--------------- | ---------------
image | image

- As per Tanyas request, the whole row is tappable and leads to the key details page


Multisig/UI/UI Library/Views/AddressInfoView.swift
```diff
@@ -15,6 +15,7 @@ class AddressInfoView: UINibView {
     @IBOutlet private weak var addressLabel: UILabel!
     @IBOutlet private weak var copyButton: UIButton!
     @IBOutlet private var detailButton: UIButton!
```
biafra23: I tried using setImage() before, but there was no image visible. See previous commits. So I made an extra whole button. :-/

DmitryBespalov:
>     The arrow is a bit tough to hit. I assumed we wanted to keep the copy key functionality. If that is not necessary the whole row can be tappable to go to the key details.

@TanyaEfremova does that sounds OK? I have a doubt that if there's an arrow on the row, user needs to click the arrow to open key details instead of just tapping anywhere in the row

Multisig/UI/UI Library/Views/AddressInfoView.swift
```diff
@@ -190,9 +208,14 @@ class AddressInfoView: UINibView {
...
-        if let url = sender.browseURL {
-            openInSafari(url)
+        if sender.showDetailNavigation,
+           let keyInfo = try? KeyInfo.firstKey(address: sender.address) {
+                let vc = OwnerKeyDetailsViewController(keyInfo: keyInfo)
+                show(vc, sender: self)
+        } else {
+            if let url = sender.browseURL {
+                openInSafari(url)
+            }
```
DmitryBespalov: this is not how this should be implemented. Instead, we need to use a simple table view row with accessory input, and put the address there. We should have examples of such thing in several places in the interface.

TanyaEfremova:
@DmitryBespalov @biafra23 I agree, that the whole row can be tapped, as in this case going to the owner details would be more important than copying the address and this way it's more accessible.

DmitryBespalov (approved):

Issue #3264, opened 2023-08-04
[iOS] [On-ramp] Buy button is available on UI on gnosis chain after a new safe creation after clean app installation


liliya-soroka:
**Bug description**
Buy button is available on UI on gnosis chain after a new safe creation after clean app installation
**Steps To Reproduce**
1. clean app install
2. start safe creation from onboarding 
3. select Gnosis chain
4. create a safe using social login
5. wait till safe is ready popup appears
6. click X to close ready pop-up
7. click Buy
8. select safe pr owner key
Current result: Buy flow is closed with error
**Expected Result**
Make sure that Buy is not visisable if there is no support for the chain on the moonpay side
**Screenshots**

https://github.com/safe-global/safe-ios/assets/338622/d8c54aab-cf54-4bcf-ab76-209f35c2a59f

**Device & App version**
 - Device/OS: iPhone 11/16.4
 - App version: 3.22.0(1913)
 
**Environment**
prod
**Additional info**
(Optional)

liliya-soroka:
Verified, app version 3.22.0(1915)
Checked
Assets page view for the next cases:
1. safe creation from onboarding for the networks without moonpay support
2. safe creation from onboarding for the networks with moonpay support
3. safe creation from switch safe menu for both cases
4. add existing safe

PR #3266, opened 2023-08-04
Refresh ui on view creation

MouazAlzahabi:
Handles #3264

### DIFF
--- a/Multisig/UI/Assets/AssetsViewController.swift
+++ b/Multisig/UI/Assets/AssetsViewController.swift
@@ -89,7 +89,6 @@ class AssetsViewController: ContainerViewController {
             self?.present(vc, animated: true)
         }
 
-        totalBalanceView.tokenBanner.isHidden = !shouldShowSafeTokenBanner
         totalBalanceView.tokenBanner.onClaim = { [unowned self] in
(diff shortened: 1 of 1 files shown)


Multisig/UI/Assets/AssetsViewController.swift
```diff
@@ -163,6 +165,10 @@ class AssetsViewController: ContainerViewController {
     
     @objc private func selectedSafeUpdatedReceived(notification: Notification) {
         self.safe = notification.object as? Safe
+        updateSafeOptions()
+    }
+
+    private func updateSafeOptions() {
```
DmitryBespalov, line 171: should this also be used on 'selected safe changed'?

DmitryBespalov (approved):

Multisig/UI/Assets/AssetsViewController.swift
```diff
@@ -163,6 +165,10 @@ class AssetsViewController: ContainerViewController {
     
     @objc private func selectedSafeUpdatedReceived(notification: Notification) {
         self.safe = notification.object as? Safe
+        updateSafeOptions()
```
DmitryBespalov, line 175: would be good to use it on the selected safe changed, too

Issue #3263, opened 2023-08-04
[iOS] [Web3Auth] Sometime the "progress" screen is not closed automatically


liliya-soroka:
**Bug description**
Sometimes the "progress" screen is not closed automatically. The issue happens sometime( not often). The user has to close the app and reopen 
I will research it more
**Steps To Reproduce**
TBD - 
1. create a safe using social login 
...
Current result: The "progress" screen is not closed

**Expected Result**
the "progress" screen should be closed after submitting tx
**Screenshots**
https://drive.google.com/file/d/1mWbtVugqJAM4B0i3SX1Nl75Hk6bjM5pJ/view?usp=sharing
**Device & App version**
 - Device/OS: iPhone 11/16.4
 - App version: 3.22.0(1913)
 
**Environment**
prod
**Additional info**
(Optional)

DmitryBespalov:
Steps to reproduce:
1. Fresh install and close the onboarding or delete all your safes
2. Tap "Create new Safe Account"
3. Create safe with Google
--- 
The UI is stuck in the progress screen

liliya-soroka:
I did check the issue according to the steps from the comment above. But I will check it better during MFA tasks testing.

app version 3.23.0 (1951)

liliya-soroka:
Verified
app version 3.23.0 (1958)

PR #3279, opened 2023-08-08
GH-3263 closing presented screen from the presenting screen

DmitryBespalov:
Handles #3263 

- Closing screen when presenter is not the presenting view controller

### DIFF
--- a/Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
+++ b/Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
@@ -71,7 +71,13 @@ class UIFlow: NSObject {
     }
 
     func stop(success: Bool) {
-        if let presenter = presenter {
+        if let presenter = presenter, presenter.presentedViewController != nil {
+            presenter.dismiss(animated: true) { [unowned self] in
+                completion(success)
+            }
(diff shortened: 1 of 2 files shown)


Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
```diff
@@ -71,7 +71,13 @@ class UIFlow: NSObject {
     }
 
     func stop(success: Bool) {
-        if let presenter = presenter {
+        if let presenter = presenter, presenter.presentedViewController != nil {
```
MouazAlzahabi, line 74: Is it possible that presenter != nil and  presenter.presentedViewController = nil ?

MouazAlzahabi (approved):

Multisig/UI/Settings/OwnerKeyManagement/Backup/UIFlow.swift
DmitryBespalov: yes, that's exactly the root cause of the view controller not dismissed. The view controller was presented from the child view controller, which forwards the presentation to the ancestor - main tab bar view controller. Calling `presenter.dismiss()` in that situation has no effect, because the presenter didn't have any controller presented.

This is not evident from the code, so I added another comment in the `else if` branch

Issue #3084, opened 2023-04-26
[iOS] Update Gnosis Chain Gas Estimation by 30%


the-real-chrizzo:
**Background**
Users have the problem that they cannot execute transactions on Gnosis Chain because the gas estimation is too low.

The web team already implemented a fix to mitigate this issue by increasing the estimation by 30% as specified here https://github.com/safe-global/web-core/blob/dev/src/hooks/useGasLimit.ts#L78-L86.

We need to the same for iOS.

**For additional reference** see
https://github.com/NethermindEth/nethermind/issues/5637

**Acceptance criteria**
Increased gas estimation on Gnosis Chain

DmitryBespalov:
preconditions:
* gnosis chain
* safeTxGas > 0
* contract version < 1.3.0

liliya-soroka:
Verified
app version 3.23.0(1948)

PR #3280, opened 2023-08-08
GH-3084 adjust gas for gnosis chain and safe transactions

DmitryBespalov:
Handles #3084

- Adjusting the gas estimation

### DIFF
--- a/Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
+++ b/Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
@@ -241,8 +241,7 @@ class TransactionExecutionController {
                        let safeTxGas = transaction.multisigInfo?.safeTxGas.value,
                        safeTxGas > 0 {
                         // `gas` is UInt64, so it can overflow if a large number returned from the estimator
-                        let (newGas, overflow) = gas!.multipliedReportingOverflow(by: 130)
-                        gas = (overflow ? .max : newGas) / 100
+                        gas = Sol.UInt64(clamping: Sol.UInt256( gas! ) * 130 / 100 )
                     }
 
                     let execTransactionSuccess = try partialResults.ethCall.get()
--- a/Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
+++ b/Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
@@ -242,7 +242,7 @@ class TransactionExecutionController {
                        safeTxGas > 0 {
                         // `gas` is UInt64, so it can overflow if a large number returned from the estimator
                         let (newGas, overflow) = gas!.multipliedReportingOverflow(by: 130)
-                        gas = overflow ? .max : newGas
+                        gas = (overflow ? .max : newGas) / 100
                     }
 
                     let execTransactionSuccess = try partialResults.ethCall.get()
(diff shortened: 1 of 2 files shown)


elgatovital (approved):

Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
```diff
@@ -234,6 +234,18 @@ class TransactionExecutionController {
...
                     // to avoid that, we handle both cases in one do-catch block.
                     gas = try partialResults.gas.get()
+                    
+                    // GH-3084: fix for a bug in Nethermind requiring 30% increase in the gas estimation
+                    // on gnosis chain for transactions with positive safeTxGas
+                    if chain.id == Chain.ChainID.gnosis,
+                       let safeTxGas = transaction.multisigInfo?.safeTxGas.value,
+                       safeTxGas > 0 {
+                        // `gas` is UInt64, so it can overflow if a large number returned from the estimator
+                        let (newGas, overflow) = gas!.multipliedReportingOverflow(by: 130)
```
MouazAlzahabi: Isn't it safer if we multiply by 1.3 and then cast down the result to nearest UInt64 value ?

Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
```diff
@@ -234,6 +234,18 @@ class TransactionExecutionController {
...
+                        if !overflow {
```
MouazAlzahabi: let's set the gas to be UInt64.max if overflow occurs?

Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
DmitryBespalov: we'd have to multiply by an integer value since it's uint64, however, can also use the highest number if it overflows

Multisig/UI/Transaction/ExecuteTransaction/TransactionExecutionController.swift
```diff
@@ -234,6 +234,16 @@ class TransactionExecutionController {
...
                     gas = try partialResults.gas.get()
+                    
+                    // GH-3084: fix for a bug in Nethermind requiring 30% increase in the gas estimation
+                    // on gnosis chain for transactions with positive safeTxGas
+                    if chain.id == Chain.ChainID.gnosis,
+                       let safeTxGas = transaction.multisigInfo?.safeTxGas.value,
+                       safeTxGas > 0 {
+                        // `gas` is UInt64, so it can overflow if a large number returned from the estimator
+                        let (newGas, overflow) = gas!.multipliedReportingOverflow(by: 130)
+                        gas = (overflow ? .max : newGas) / 100
```
MouazAlzahabi: I know it's an edge case but I think this is wrong, imagine if gas > .max/100, this way we have reduced the gas instead of increase

DmitryBespalov: the other way is be to temp convert to u256, make an operation, and then limit with higher bound at u64 max;

    gas = u64( u256(gas) * 130 / 100 )

    would handle all cases

yet another way to add a 1/3 of the value;

    let (newGas, overflow) = gas + gas / 3
    gas = overflow ? .max : newGas

    ^^ for gas < 3 will result in no change;

MouazAlzahabi (approved):

Issue #3282, opened 2023-08-09
[iOS] [On-ramp] Light mode background icon on home screen


the-real-chrizzo:
Update the icon in light mode since resolution is not high enough.

**Acceptance criteria**

Updated icon in light mode

liliya-soroka:
Verified
app version 3.23.0(1948)

PR #3287, opened 2023-08-10
Change icon

MouazAlzahabi:
Handles #3282


Multisig/Assets.xcassets/illustrations/ico-no-collectibles.imageset/Contents.json
```diff
@@ -1,7 +1,7 @@
 {
   "images" : [
     {
-      "filename" : "ico-no-collectibles.png",
+      "filename" : "↳ Icon.pdf",
```
DmitryBespalov: such a strange filename 🤷

DmitryBespalov (approved):

Multisig/Assets.xcassets/illustrations/ico-no-collectibles.imageset/Contents.json
MouazAlzahabi: Honestly I just extract from Figma and use it without notice the name, I will just rename to avoid any troubles

Issue #3277, opened 2023-08-07
[MFA] Setup Security Password


the-real-chrizzo:
**Background/What needs to be done**

We want to set up the accessibility of a password as 1 share of the key.
Entry point for the screens will be when user clicks on "Security password" on the Security Overview.
PW assessment capabilities are reflected by the implementation of the password library item (https://github.com/safe-global/safe-ios/issues/3274)

This story entails 5 screens:

Start screen: https://www.figma.com/file/Oo95065RJXyuQIFkdMQ9Cs/MFA-authentication?type=design&node-id=1237%3A3734&mode=design&t=3wfKYbJOeVl2CmAr-1
Set Password screen: https://www.figma.com/file/Oo95065RJXyuQIFkdMQ9Cs/MFA-authentication?type=design&node-id=1237%3A3792&mode=design&t=3wfKYbJOeVl2CmAr-1

Set Password error screen: https://www.figma.com/file/Oo95065RJXyuQIFkdMQ9Cs/MFA-authentication?type=design&node-id=1237%3A3838&mode=design&t=3wfKYbJOeVl2CmAr-1
Set Password Matched screen: https://www.figma.com/file/Oo95065RJXyuQIFkdMQ9Cs/MFA-authentication?type=design&node-id=1237%3A3864&mode=design&t=3wfKYbJOeVl2CmAr-1
Success screen: https://www.figma.com/file/Oo95065RJXyuQIFkdMQ9Cs/MFA-authentication?type=design&node-id=1237%3A3712&mode=design&t=3wfKYbJOeVl2CmAr-1

Analytics events:
https://docs.google.com/spreadsheets/d/1mj9iQIhpM-Pak7lQMb1dmxhQn1QdJoJyhbc730dZNzA/edit#gid=1949479240

<google-sheets-html-origin><style type="text/css"></style>

screen_set_pw_start | Set password start screen
-- | --
user_click_pw_continue | User clicks on continue on the "Set password start screen"
user_click_pw_create_pw | User clicks on create password
screen_set_pw_sucess | Screen password created successfully appears

**Acceptance criteria**

Screens implemented as specified above
Analytics events implemented

the-real-chrizzo:
@MouazAlzahabi would you like to implement the "pw logic" here as well (re this ticket on pw library research https://github.com/safe-global/safe-ios/issues/3274) or should we create a separate ticket for the integration.

PR #3298, opened 2023-08-16
Gh 3277/setup security password

MouazAlzahabi:
Handles #3277 

- Create password flow
- Handle password validation will be done later

https://github.com/safe-global/safe-ios/assets/58591867/7892f542-00f6-43e9-87bb-f44cf39deea4


Multisig/UI/UI Library/Views/PasswordInputField.swift
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        stateLabel.setStyle(stateStyle)
+        stateLabel.text = stateText
+        setMessageText(message)
+        stateImageView.image = UIImage(named: stateImage)
+    }
+
+    @IBAction private func textChanged(_ sender: Any) {
+        let state = onTextChanged(text!)
```
DmitryBespalov, line 110: For the future: I think there should be 0.2 debouncing after entering the text so that the password strength feedback wouldn't be annoying to use

DmitryBespalov (approved):
code-wise, looks good.

Multisig/UI/UI Library/Views/PasswordInputField.swift
MouazAlzahabi: I have this in my mind, but thought to do that together with the validation part

Issue #3297, opened 2023-08-14
[WebLinks] [iOS] Links to the web app are outdated


katspaugh:
**Original request**
In the web app, we switched from URLs like `/eth:0x1234/transactions` to `/transactions?safe=eth:0x1234`.
I.e. the chain:address part is now a query parameter instead of a part of the path.

The old URLs will still work in the foreseeable future, as they are redirected, but they pollute our page view analytics. So ideally, please switch to the new format.

### Sub-items
- #3305 
- #3306 
- #3307 
- #3308

DmitryBespalov:
https://www.notion.so/safe-global/Page-routing-1e5cb2c314b149aa828466144061dc9e

DmitryBespalov:
#  Routes

## Variables

    safe-address = chain-prefix : checksummed-address
    
    chain-prefix = short name of a chain, such as 'eth', 'matic' or 'gno',
        according to the https://safe-client.safe.global/v1/chains 'shortName' field
    
    checksummed-address = ethereum & evm-chain address in a hexadecimal, 
        checksummed format according to https://eips.ethereum.org/EIPS/eip-55
    
    transaction-id = transaction identifier string

## Legacy web app &rarr; mobile app routes

`/safe-address/balances` &rarr; `Assets Tab - Balances`

`/safe-address/transactions` &rarr; `Transactions Tab - History`
`/safe-address/transactions/history` &rarr; `Transactions Tab - History`
`/safe-address/transactions/queue` &rarr; `Transactions Tab - Queue`
`/safe-address/transactions/transaction-id` &rarr; `Modal Transaction Details`

## Current web app &rarr; mobile app routes

### Onboarding and Home

`/welcome` &rarr; `Assets Tab`
    start page

`/home?safe=safe-address` &rarr; `Assets Tab`
    home / dashboard

`/new-safe/load` &rarr; `Modal Load Existing Safe`
    load new safe flow

`/new-safe/create` &rarr; `Modal Create Safe`
    create new safe flow
    
### Assets
    
`/balances?safe=safe-address` &rarr; `Assets Tab - Coins`
    assets of a safe
    
`/balances/nfts?safe=safe-address` &rarr; `Assets Tab - Collectibles`
    nfts of a safe
    
### Transactions
   
`/transactions?safe=safe-address` &rarr; `Transactions Tab - Queue`
    redirects to transaction queue
    
`/transactions/history?safe=safe-address` &rarr; `Transactions Tab - History`
[...]

PR #3300, opened 2023-08-21
GH-3297 support current web app url format

DmitryBespalov:
Handles #3297 

- Keep supporting previous format
- Added tests

Overall:
- [x] asset - balances
- [x] assets - nfts
- [x] transactions history
- [x] transaction queue
- [x] transaction details
- [ ] safe creation
- [ ] safe loading
- [ ] safe settings
- [ ] app settings, legal info

### DIFF
--- a/Multisig/Features/Connect to Web/UI/List/WebConnectionsViewController.swift
+++ b/Multisig/Features/Connect to Web/UI/List/WebConnectionsViewController.swift
@@ -212,6 +212,10 @@ extension WebConnectionsViewController: QRCodeScannerViewControllerDelegate {
 }
 
 extension WebConnectionsViewController: NavigationRouter {
+    func routeFrom(from url: URL) -> NavigationRoute? {
+        nil
+    }
+    
     func canNavigate(to route: NavigationRoute) -> Bool {
         route.path == NavigationRoute.connectToWeb().path
     }
--- a/Multisig/UI/App/NavigationRouter.swift
+++ b/Multisig/UI/App/NavigationRouter.swift
@@ -130,6 +130,8 @@ class ExtendedNavigationRouter: NavigationRouter {
     func routeFrom(from url: URL) -> NavigationRoute? {
         
         switch url.path {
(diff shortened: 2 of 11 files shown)


Multisig/UI/App/NavigationRouter.swift
```diff
@@ -24,6 +25,43 @@ extension NavigationRouter {
     }
 }
 
+class CompositeNavigationRouter: NavigationRouter {
+    static let shared = CompositeNavigationRouter(routers: [ExtendedNavigationRouter(), DefaultNavigationRouter()])
```
MouazAlzahabi, line 29: Do we really need this differentiate between Routers? can we just merge them all ?

Multisig/Features/Connect to Web/UI/List/WebConnectionsViewController.swift
```diff
@@ -212,6 +212,10 @@ extension WebConnectionsViewController: QRCodeScannerViewControllerDelegate {
 }
 
 extension WebConnectionsViewController: NavigationRouter {
+    func routeFrom(from url: URL) -> NavigationRoute? {
```
MouazAlzahabi, line 215: Can you add a comment why this returns nil for now?

DmitryBespalov: it's nil because this router is not used to get the url -> navroute conversion.

The method is not used on this instance (and others).

Multisig/UI/App/NavigationRouter.swift
DmitryBespalov: it was easier to handle different web links versions - I just added the new routes and it works with the previous way, too, leading to less complicated classes

DmitryBespalov: in the future, when the old way is not supported, then we just remove that router

MouazAlzahabi (approved):

elgatovital (approved):

PR #3312, opened 2023-08-24
fix: dependency resolution

metalurgical:
Handles #
Dependency resolution of secp256k1


biafra23:
Thanks for the support!

I did the necessary changes in `GH-xxxx/MPC-with-password`

Issue #3305, opened 2023-08-23
[WebLinks] Create and load safe


DmitryBespalov:
## Variables

    safe-address = chain-prefix : checksummed-address
    
    chain-prefix = short name of a chain, such as 'eth', 'matic' or 'gno',
        according to the https://safe-client.safe.global/v1/chains 'shortName' field
    
    checksummed-address = ethereum & evm-chain address in a hexadecimal, 
        checksummed format according to https://eips.ethereum.org/EIPS/eip-55
   

## Routes

`/new-safe/load` &rarr; `Modal Load Existing Safe`
    load new safe flow

`/new-safe/load?chain=chain-prefix&address=checksummed-address`
    pre-select the chain and fill in the safe address

`/new-safe/load?chain=chain-prefix`
    pre-select the chain

`/new-safe/load?address=safe-address`
    pre-select the chain and safe address from the eip-3770 safe address

`/new-safe/create` &rarr; `Modal Create Safe`
    create new safe flow

`/new-safe/create?chain=chain-prefix` &rarr; `Modal Create Safe`
   pre-select the chain

_Originally posted by @DmitryBespalov in https://github.com/safe-global/safe-ios/issues/3297#issuecomment-1682421287_

liliya-soroka:
@DmitryBespalov , is it expected that navigation works only if the app is called from the background and doesn't work if it's cold app call

app version 3.23.0 (1951)

DmitryBespalov:
@liliya-soroka that's a good one, I overlooked it

PR #3314, opened 2023-08-28
GH-3305 handle chain and address parameters; createSafe route

DmitryBespalov:
Handles #3305 

- Added input parameters to the flow
- Handling `createSafe` route

### DIFF
--- a/Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
+++ b/Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
@@ -40,9 +40,7 @@ class EnterSafeAddressViewController: UIViewController {
         navigationItem.rightBarButtonItem = nextButton
 
         if let address = preselectedAddress {
-            DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500), execute: { [weak self] in
-                self?.didEnterText(address)
-            })
+            didEnterText(address)
         }
     }
 
--- a/Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
+++ b/Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
@@ -15,6 +15,8 @@ class EnterSafeAddressViewController: UIViewController {
     var completion: (_ address: Address, _ safeVersion: String) -> Void = { _, _ in }
     var chain: SCGModels.Chain!
     var safeVersion: String?
+    var preselectedAddress: String?
+    
     lazy var trackingParameters: [String: Any]  = { ["chain_id" : chain.chainId.description] }()
 
     @IBOutlet private weak var headerLabel: UILabel!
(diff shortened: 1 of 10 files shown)


elgatovital (approved):

MultisigTests/Logic/NavigationRouterTests.swift
```diff
@@ -86,6 +86,57 @@ final class ExtendedNavigationRouterTests: XCTestCase {
         XCTAssertEqual(route?.path, "/loadSafe")
     }
 
+    func testLoadSafeWithNetwork() {
+        let url = "https://some.host/new-safe/load?chain=eth"
```
MouazAlzahabi: Can we extract the base url to constant variable as it's used in all methods? maybe it will be helpful later

Multisig/UI/Safe Management/Add Safe/SelectNetworkViewController/SelectNetworkViewController.swift
```diff
@@ -89,10 +91,22 @@ class SelectNetworkViewController: LoadableViewController, UITableViewDelegate,
...
         }
     }
+    
+    // Select the chain automatically based on the given input chain id
+    private func findPreselectedChainId() {
+        guard let chainId = preselectedChainId else { return }
+        
+        if let chain = model.models.first(where: { $0.id == chainId }) {
+            completion(chain)
+        } else if model.next != nil {
```
MouazAlzahabi, line 106: I think this is ok for now, but if we had multiple pages to be loaded then it will produce negative UX

Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
```diff
@@ -36,6 +38,12 @@ class EnterSafeAddressViewController: UIViewController {
         nextButton.isEnabled = false
 
         navigationItem.rightBarButtonItem = nextButton
+
+        if let address = preselectedAddress {
```
MouazAlzahabi, line 42: Isn't it safer if we did this part in viewDidAppear then we assign preselectedAddress = nil?

MouazAlzahabi (approved):

Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
DmitryBespalov: I don't understand about safer - do you mean that the code will be reachable?

MouazAlzahabi: I prefer to use viewDidAppear over dispatchAfter because of viewController life cycle

Multisig/UI/Safe Management/Add Safe/SelectNetworkViewController/SelectNetworkViewController.swift
DmitryBespalov: in that case, it will show a loading indicator for a longer time

Multisig/UI/Safe Management/Add Safe/EnterSafeAddressViewController/EnterSafeAddressViewController.swift
DmitryBespalov: because of the life cycle, I put it to the viewDidLoad - so that this code would be executed just once

DmitryBespalov: actually, I'll just remove the dispatchAfter

Issue #3306, opened 2023-08-23
[WebLinks] Transactions


DmitryBespalov:
### Variables

    safe-address = chain-prefix : checksummed-address
    
    chain-prefix = short name of a chain, such as 'eth', 'matic' or 'gno',
        according to the https://safe-client.safe.global/v1/chains 'shortName' field
    
    checksummed-address = ethereum & evm-chain address in a hexadecimal, 
        checksummed format according to https://eips.ethereum.org/EIPS/eip-55
    
    transaction-id = transaction identifier string

### Transactions
   
`/transactions?safe=safe-address` &rarr; `Transactions Tab - Queue`
    redirects to transaction queue
    
`/transactions/history?safe=safe-address` &rarr; `Transactions Tab - History`
    history of past safe transactions

`/transactions/queue?safe=safe-address` &rarr; `Transactions Tab - Queue`
    queue of staged safe transactions

`/transactions/messages?safe=safe-address` &rarr; `Transactions Tab`
    signed messages

`/transactions/tx?safe=safe-address&id=transaction-id` &rarr; `Modal Transaction Details`
    transaction details

_Originally posted by @DmitryBespalov in https://github.com/safe-global/safe-ios/issues/3297#issuecomment-1682421287_

liliya-soroka:
https://safe-wallet-web.staging.5afe.dev/transactions/history?safe=gor:0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c - queue tab is opened instead of history

app version 3.23.0 (1951)

liliya-soroka:
closed, app version 3.23.0(1960)

PR #3315, opened 2023-08-28
GH-3306 added transactions routes

DmitryBespalov:
Handles #3306 

- Routing for remaining transaction routes

### DIFF
--- a/Multisig/UI/App/NavigationRouter.swift
+++ b/Multisig/UI/App/NavigationRouter.swift
@@ -173,7 +173,7 @@ class ExtendedNavigationRouter: NavigationRouter {
                 chainId: safeAddress.chainId
             )
             return route
-        case "/transactions/queue":
+        case "/transactions/queue", "/transactions", "/transactions/messages":
             guard let safeAddress = eip3770AddressQueryParameter(named: "safe", in: url) else {
                 return nil
             }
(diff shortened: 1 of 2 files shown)


Multisig/UI/App/NavigationRouter.swift
```diff
@@ -173,7 +173,7 @@ class ExtendedNavigationRouter: NavigationRouter {
                 chainId: safeAddress.chainId
             )
             return route
-        case "/transactions/queue":
+        case "/transactions/queue", "/transactions", "/transactions/messages":
```
elgatovital, line 178: we don't have messages in the app, why supporting this route and forwarding it to tx list?

DmitryBespalov: here forwarding is to the closest possible screen, thus handling that specific URL as well. This way we handle all of the URLs that are present in the safe web app

MouazAlzahabi: @DmitryBespalov Could you please just add a TODO to support route to messages if we did it in future ?

MouazAlzahabi (approved):

Multisig/UI/App/NavigationRouter.swift
elgatovital: I think there is no need to forward something that is not available in the app

DmitryBespalov: wouldn't this be found anyway?

DmitryBespalov: can add the explanatory comment, but todo is too much now, as implementation of this is not planned yet

